PR #33, opened 2022-01-20
delete old prometheus entries

KalmanMeth:


pkg/pipeline/encode/encode_prom.go
```diff
@@ -28,49 +29,56 @@ import (
...
-type gaugeInfo struct {
-	input     string
-	tags      []string
-	promGauge *prometheus.GaugeVec
-}
-
-type counterInfo struct {
-	input     string
-	tags      []string
+type genericMetricInfo struct {
```
eranra, line 38: you can just change the name the metricInfo -- the `generic` is not really needed

pkg/pipeline/encode/encode_prom.go
```diff
@@ -28,49 +29,56 @@ import (
...
+	eInfo entrySignature
+	value float64
 }
 
-type gaugeEntryInfo struct {
-	gaugeName  string
-	gaugeValue float64
-	labels     map[string]string
+type metricCacheEntry struct {
+	label       prometheus.Labels
```
eranra, line 58: a lot of overlapping fields with the  `genericMetricInfo struct`. Maybe you can embed this struct in the cache struct? 

Look here: https://eli.thegreenplace.net/2020/embedding-in-go-part-1-structs-in-structs/

Or maybe have a dedicated field pointing to a  `genericMetricInfo struct`

pkg/pipeline/encode/encode_prom.go
```diff
@@ -122,34 +133,116 @@ func (e *encodeProm) EncodeGauge(metric config.GenericMap) []interface{} {
...
+		cEntry.promGauge = gaugeInfo.promGauge
+	}
 	return out
 }
 
+func (e *encodeProm) saveEntryInCache(entry entryInfo, entryLabels map[string]string) *metricCacheEntry {
+	// save item in cache; use eInfo as key to the cache
+	var cEntry *metricCacheEntry
+	secs := time.Now().Unix()
+	eInfoBytes, _ := json.Marshal(&entry.eInfo)
```
eranra, line 160: A bit expensive operation ... maybe there is some simpler way to get a "signature" ... maybe just use print of the tags and name ... this should take much less resources.

Also, there are "labels" and "tags" all over .. can we agree on a name for those and make sure we are consistent

pkg/pipeline/encode/encode_prom_test.go
```diff
@@ -103,18 +109,45 @@ func Test_NewEncodeProm(t *testing.T) {
...
+	require.Equal(t, 2, len(entriesMap))
+
+	eInfoBytes, _ := json.Marshal(&gEntryInfo1.eInfo)
+	encodeProm.mu.Lock()
+	_, found := encodeProm.mCache[string(eInfoBytes)]
+	encodeProm.mu.Unlock()
+	require.Equal(t, true, found)
+
+	// wait a couple seconds so that the entry will expire
+	time.Sleep(2 * time.Second)
```
eranra, line 145: Great test :-)

pkg/pipeline/encode/encode_prom.go
```diff
@@ -122,34 +133,116 @@ func (e *encodeProm) EncodeGauge(metric config.GenericMap) []interface{} {
...
+}
+
+// cleanupExpiredEntries - any entry that has expired should be removed from the prometheus reporting and cache
+func (e *encodeProm) cleanupExpiredEntries() {
+	log.Debugf("entering cleanupExpiredEntries")
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	log.Debugf("cache = %v", e.mCache)
+	log.Debugf("list = %v", e.mList)
+	secs := time.Now().Unix()
```
eranra, line 215: s/secs/nowInSecs

pkg/pipeline/encode/encode_prom.go
```diff
@@ -122,34 +133,116 @@ func (e *encodeProm) EncodeGauge(metric config.GenericMap) []interface{} {
...
+		c := entry.Value.(*metricCacheEntry)
+		log.Debugf("timeStamp = %d, expireTime = %d", c.timeStamp, expireTime)
+		log.Debugf("c = %v", c)
+		if c.timeStamp > expireTime {
+			// no more expired items
+			return
+		}
+
+		// clean up the entry
+		log.Debugf("secs = %d, deleting %s", secs, c.label)
```
eranra, line 232: you probably want to add also expireTime to the debug print here

pkg/pipeline/encode/encode_prom.go
eranra, line 58: s/label/labels

pkg/pipeline/encode/encode_prom.go
```diff
@@ -211,19 +319,28 @@ func NewEncodeProm() (Encoder, error) {
...
 		prefix:     promPrefix,
 		counters:   counters,
 		gauges:     gauges,
 		histograms: histograms,
+		expiryTime: expiryTime,
+		mList:      list.New(),
+		mCache:     make(metricCache),
 	}
 	go startPrometheusInterface(w)
+	go w.cleanupExpiredEntriesLoop()
```
eranra, line 344: There are nicer ways to implement a background thread with a timer ... something like https://levelup.gitconnected.com/timer-and-ticker-using-golang-933a6d00a832

Using a channel and timer instead of forever loop with sleep

pkg/pipeline/encode/encode_prom.go
```diff
@@ -28,49 +29,56 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"sync"
+	"time"
 )
 
-// TODO when do we clean up old entries?
+const defaultExpiryTime = 120
```
eranra, line 36: Should this be moved to the API area somehow?

pkg/pipeline/encode/encode_prom.go
KalmanMeth: Changed `tags` to `labelNames`.  Changed Marshal to fmt.Sprintf, as suggested.

pkg/pipeline/encode/encode_prom.go
KalmanMeth: Done

pkg/pipeline/encode/encode_prom.go
KalmanMeth: Factored out the common fields into PromMetric struct.

pkg/pipeline/encode/encode_prom.go
KalmanMeth: OK

pkg/pipeline/encode/encode_prom.go
KalmanMeth: expireTime appears in a separate debug print just above.

pkg/pipeline/encode/encode_prom.go
KalmanMeth: Done

pkg/pipeline/encode/encode_prom.go
```diff
@@ -177,15 +270,20 @@ func NewEncodeProm() (Encoder, error) {
...
+	expiryTime := int64(jsonEncodeProm.ExpiryTime)
+	if expiryTime == 0 {
+		expiryTime = defaultExpiryTime
+	}
+	log.Debugf("expiryTime = %d", expiryTime)
 
-	counters := make(map[string]counterInfo)
-	gauges := make(map[string]gaugeInfo)
-	histograms := make(map[string]histInfo)
+	counters := make(map[string]genericMetricInfo)
```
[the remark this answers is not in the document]
KalmanMeth: Done

pkg/pipeline/encode/encode_prom.go
KalmanMeth: I changed it to use a ticker instead of sleep.

eranra: OK ... there are more "fancy" things to add here but this is ok for now :-) the other things are more like clearing the thread when the code exits ... but this is not critical for us.

pkg/pipeline/encode/encode_prom.go
```diff
@@ -176,53 +249,67 @@ func NewEncodeProm() (Encoder, error) {
...
+		labels := mInfo.Labels
 		log.Debugf("fullMetricName = %v", fullMetricName)
-		log.Debugf("labels = %v", labels)
-		switch metricInfo.Type {
+		log.Debugf("Labels = %v", labels)
+		pMetric.metricType = mInfo.Type
+		switch mInfo.Type {
 		case api.PromEncodeOperationName("Counter"):
 			counter := prometheus.NewCounterVec(prometheus.CounterOpts{Name: fullMetricName, Help: ""}, labels)
 			err := prometheus.Register(counter)
```
eranra, line 269: I am looking on https://github.com/netobserv/flowlogs2metrics/pull/33/files#diff-30d11303a4ad1ec82e4f39e3ae6dc7d044c8dd370c536a98ba90febbfe2d425aR277 and on https://github.com/netobserv/flowlogs2metrics/pull/33/files#diff-30d11303a4ad1ec82e4f39e3ae6dc7d044c8dd370c536a98ba90febbfe2d425aR269 and on https://github.com/netobserv/flowlogs2metrics/pull/33/files#diff-30d11303a4ad1ec82e4f39e3ae6dc7d044c8dd370c536a98ba90febbfe2d425aR286 can you collapse this code so it will not be repeating in each of the cases ???

eranra (approved):

PR #40, opened 2022-01-20
push_image to quay

eranra:


jotak:
Just for information, in our other repositories we have adopted the other way around: we configured a build trigger in quay whenever something is merged (quay can listen to branches on github repos).

But to be honest, maybe we'll adopt this pattern instead (with GH action) ; we had troubles recently with quay being overloaded and not able to build...
(cc @netobserv/maintainers  )

jotak (approved):

PR #52, opened 2022-02-01
add support for connection tracking

eranra:


pkg/pipeline/transform/connection_tracking/connection_tracking.go
```diff
@@ -0,0 +1,151 @@
...
+	"hash/fnv"
+	"sync"
+	"time"
+)
+
+type cacheKey uint32
+
+type cacheInfo struct {
+	key             cacheKey
+	flowlogsCount   int
```
KalmanMeth, line 32: What is the purpose of the `flowlogsCount` field? It doesn't seem to be used at present.

pkg/pipeline/transform/connection_tracking/connection_tracking_test.go
```diff
@@ -0,0 +1,71 @@
...
+	require.Equal(t, false, isNew)
+	require.Equal(t, CT.cacheMap[getHashKey("test")].flowlogsCount, 2)
+}
+
+func Test_IsFlowNew(t *testing.T) {
+	InitConnectionTracking()
+	_ = CT.AddFlow("test")
+	isNew := CT.IsFlowNew("test")
+	require.Equal(t, true, isNew)
+	isNew = CT.IsFlowNew("test_false")
```
KalmanMeth, line 47: The name `IsFlowNew` is confusing. The "test_false" flow will not be in the cache so it will return `false`. This is not because it is not new, but rather because it does not exist. An entry is in the cache not because it is new, but rather because it is currently active.

pkg/pipeline/transform/connection_tracking/connection_tracking.go
eranra: correct, the plan going forward is to use that field and others (like flow bytes) to add another extract phase that will be topx  ... so we will be able to report the flows with "top amount of packets" , "top amount of bytes" etc. ...

pkg/pipeline/transform/connection_tracking/connection_tracking_test.go
eranra: Renamed to `IsFlowKnown`

pkg/pipeline/transform/connection_tracking/connection_tracking.go
```diff
@@ -0,0 +1,151 @@
...
+	return isNew
+}
+
+func (ct ConnectionTracking) cleanupExpiredEntriesLoop() {
+	ticker := time.NewTicker(time.Duration(ct.expiryTime) * time.Second)
+	done := make(chan bool)
+	go func() {
+		for {
+			select {
+			case <-done:
```
KalmanMeth, line 113: Where is this channel ever set?

pkg/pipeline/transform/connection_tracking/connection_tracking.go
```diff
@@ -0,0 +1,151 @@
...
+		if pCacheInfo.lastUpdatedTime > expireTime {
+			return
+		}
+		delete(ct.cacheMap, pCacheInfo.key)
+		ct.cacheList.Remove(listEntry)
+	}
+}
+
+func InitConnectionTracking() {
+	CT = ConnectionTracking{
```
KalmanMeth, line 143: This structure (`CT`) seems to be here exclusively for testing.

KalmanMeth (approved):

pkg/pipeline/transform/connection_tracking/connection_tracking.go
eranra: @KalmanMeth is not ....  in the current implementation. When we move to multi-threaded or scale-down we might need that. This is the common pattern in golang for waiting on a timer.

PR #55, opened 2022-02-03
Kafka ingress

KalmanMeth:


eranra:
@jotak please review. Can you advise where is the code that sends the logs to Kafka? So that @KalmanMeth will be able to verify that we can consume the flow logs in the correct format.

jotak:
@eranra I'm adding other team members so that they can have a look too (it doesn't mean that everyone will review, just increase the chance to get a review faster :-) )

jotak:
in particular for Kafka, @jpinsonneau and @mariomac have been more involved than me

jpinsonneau:
Looks good ! 

We used the same approach however there is a lot of extra configuration that should interest you. 
Check our [yaml configuration](https://github.com/netobserv/goflow2-kube-enricher/blob/afcd39efced182449c48b6742f3c53a08eafcad9/pkg/config/kafka.go) for both reader & writer, [reader implementation](https://github.com/netobserv/goflow2-kube-enricher/blob/afcd39efced182449c48b6742f3c53a08eafcad9/pkg/format/kafka/kafka.go#L20) and [writer implementation](https://github.com/netobserv/goflow2-kube-enricher/blob/afcd39efced182449c48b6742f3c53a08eafcad9/pkg/export/kafka/kafka.go#L32) if you plans to do that way also.

pkg/pipeline/ingest/ingest_kafka.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// connect to the kafka server
+	brokers := jsonIngestKafka.Brokers
+	topic := jsonIngestKafka.Topic
+	groupId := jsonIngestKafka.GroupId
+	kafkaReader := kafka.NewReader(kafka.ReaderConfig{
+		Brokers: brokers,
+		Topic:   topic,
+		GroupID: groupId,
+	})
```
jpinsonneau, line 69: I think you should at least add `StartOffset` and `GroupBalancers` that are useful

jotak:
Note that the PR is not merged yet, on our side. It's here:  https://github.com/netobserv/goflow2-kube-enricher/pull/23

pkg/pipeline/ingest/ingest_kafka.go
mariomac: I think that `StartOffset` should always be the default `FirstOffset` value. Any use case to ignore uncommited messages?

jpinsonneau: Well if you stopped reading for some time and have a lot of old messages you don't care about you may want to skip them to directly start at new elements ? Or you need to flush kafka before restart ?

I was thinking at the situations where an admin wants to capture a specific period of time. He would not expect messages before he pressed "start".

But I agree by default if would be FirstOffset :+1:

mariomac: I understand. Thanks for the clarification!

pkg/pipeline/ingest/ingest_kafka.go
```diff
@@ -0,0 +1,80 @@
...
+	kafkaParams api.IngestKafka
+	kafkaReader *kafka.Reader
+}
+
+// Ingest ingests entries from kafka topic and sends them down the pipeline
+func (r *ingestKafka) Ingest(process ProcessFunction) {
+	lines := make([]interface{}, 1)
+
+	for {
+		m, err := r.kafkaReader.ReadMessage(context.Background())
```
eranra, line 40: please use significant name and not m

pkg/pipeline/ingest/ingest_kafka.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			log.Errorln(err)
+		}
+		log.Debugf("message at topic:%v partition:%v offset:%v	%s = %s\n", m.Topic, m.Partition, m.Offset, string(m.Key), string(m.Value))
+		lines[0] = string(m.Value)
```
eranra, line 45: So, are we getting here one line by the other, or bulk of messages ... why this is like lines[0] ... this will make the "process" function much less efficient

README.md
```diff
@@ -42,7 +42,9 @@ Flags:
       --pipeline.encode.type string            Encode type: prom, none (default "none")  
       --pipeline.extract.aggregates string     Aggregates (see docs)  
       --pipeline.extract.type string           Extract type: aggregates, none (default "none")  
+      --pipeline.ingest.collector string       Ingest collector API  
       --pipeline.ingest.file.filename string   Ingest filename (file)  
+      --pipeline.ingest.kafka string           Kafka ingest API
```
eranra, line 47: I think that there is a place to add an explanation on Kafka ingest as a small section in the ream-me  --- as alternative if we want to look forward ... you can extend https://github.com/netobserv/flowlogs2metrics/pull/55/files#diff-67a6e36014be865053c43d38b2091bd2d3f17a91b117d97c0f5244cd7d71127dR29 to include all the info ... going forward my thinking was to add less and less text to the README and move the text to be part of the automatic generated code

eranra (commented):
@KalmanMeth I assume you are working on some tests for this

pkg/pipeline/ingest/ingest_kafka.go
KalmanMeth: We read whatever the kafka topic has to offer. I did not do here any parsing of the input, since I don't yet know what to expect. The rest of the pipeline expects an array of lines. If we know the kinds of divisions expected in the kafka, then I should divide it up here into "lines".

KalmanMeth:
@eranra Let's discuss how to write tests for an ingest routine. There is the problem that we need to connect to an outside entity in order to have this component run, but during the test we do not assume we have any outside entities. At present, none of our ingest routines have tests.

pkg/pipeline/ingest/ingest_kafka.go
KalmanMeth: @jpinsonneau @mariomac What do we need in terms of authentication/authorization to use the kafka topic?

eranra:
> @eranra Let's discuss how to write tests for an ingest routine. There is the problem that we need to connect to an outside entity in order to have this component run, but during the test we do not assume we have any outside entities. At present, none of our ingest routines have tests.

@KalmanMeth maybe using something like https://github.com/orlangure/gnomock/tree/master/preset/kafka

pkg/pipeline/ingest/ingest_kafka.go
eranra: @KalmanMeth ok ... so this is maybe another good reason to have some simple test to show what kind of input we expect to get in kafaka. 

@mariomac @jpinsonneau any data to share on the expected format of Kafka messages?

jpinsonneau: Each message is the output of flow collector before or after enrichment and other format processing from the pipeline so I guess there is no expected format other than a valid `JSON` ?

Here is an example of a single message we get after enrichment:
`{"BiFlowDirection":0,"Bytes":20800,"DstAS":0,"DstAddr":"10.96.0.1","DstMac":"0a:58:0a:f4:00:01","DstNet":0,"DstPort":443,"DstVlan":0,"EgressVrfID":0,"Etype":2048,"EtypeName":"IPv4","ForwardingStatus":0,"FragmentId":0,"FragmentOffset":0,"IPTTL":0,"IPTos":0,"IPv6FlowLabel":0,"IcmpCode":0,"IcmpName":"","IcmpType":0,"InIf":12,"IngressVrfID":0,"NextHop":"","NextHopAS":0,"OutIf":0,"Packets":400,"Proto":6,"ProtoName":"TCP","SamplerAddress":"10.244.0.2","SamplingRate":0,"SequenceNum":577,"SrcAS":0,"SrcAddr":"10.244.0.5","SrcHostIP":"10.89.0.2","SrcMac":"0a:58:0a:f4:00:05","SrcNamespace":"local-path-storage","SrcNet":0,"SrcPod":"local-path-provisioner-78776bfc44-p2xkl","SrcPort":56144,"SrcVlan":0,"SrcWorkload":"local-path-provisioner","SrcWorkloadKind":"Deployment","TCPFlags":0,"TimeFlowEnd":0,"TimeFlowStart":0,"TimeReceived":1628419398,"Type":"IPFIX","VlanId":0}`

For now on our side, if some mandatory fields are missing, the messages will be filtered by the queries on loki side and not displayed in the console plugin. 
Do you think we should have an option here to check some user specified fields / values before writing to kafka and skip messages that doesn't match ?

eranra: @jpinsonneau thanks :-) .... what you shared is the raw json after enrichment ... do you have some specific Kafka info that @KalmanMeth will be able to use ... something that has to do with topics, single vs. bulk messages ... etc?

Actually, on that front, I am in favor of a much more relaxed approach ... if we have a sub-set of the information and it does provide value I would not advise dropping it. We can implement those checks if you think that this is needed. 

@KalmanMeth your thoughts ???

KalmanMeth: I think our kafka ingester should be as generic as possible, and we can add decoders to handle different types of input. If each json map is a single kafka record, I think we are fine. I am adding to the code the ability to process all the entries available in the topic, so we can process multiple lines in each loop of the process.

eranra: @KalmanMeth agree :-) ./... ok

pkg/pipeline/ingest/ingest_kafka.go
jpinsonneau: You first need to create a kafka user and [set credentials in the dialer](https://github.com/segmentio/kafka-go#sasl-support)

For now I only played with [strimzi](https://strimzi.io/) default config which is:

```yaml
kind: KafkaUser
apiVersion: kafka.strimzi.io/v1beta2
metadata:
  name: my-user
  labels:
    strimzi.io/cluster: my-cluster
  namespace: openshift-operators
spec:
  authentication:
    type: tls
  authorization:
    type: simple
    acls:
      - resource:
          type: topic
          name: my-topic
          patternType: literal
        operation: Read
        host: '*'
      - resource:
          type: topic
          name: my-topic
          patternType: literal
        operation: Describe
        host: '*'
      - resource:
          type: group
          name: my-group
          patternType: literal
        operation: Read
        host: '*'
      - resource:
          type: topic
          name: my-topic
          patternType: literal
        operation: Write
        host: '*'
      - resource:
          type: topic
          name: my-topic
          patternType: literal
        operation: Create
        host: '*'
      - resource:
          type: topic
          name: my-topic
          patternType: literal
        operation: Describe
        host: '*'
```

You should not need `Write` and `Create` authorization for reading.

How do you currently create your kafka instance ?

KalmanMeth: I currently run kafka directly on my development VM. No user management.

pkg/pipeline/ingest/ingest_kafka.go
KalmanMeth: done

pkg/pipeline/ingest/ingest_kafka.go
mariomac: I guess the format for the flows is still to be decided: JSON vs IPFIX vs any other binary encoding, so it's fine to keep it as generic as possible at the moment.

pkg/pipeline/ingest/ingest_kafka.go
eranra: 👍

README.md
KalmanMeth: Added section to api.md

KalmanMeth:
I converted the kafka ingest to use the new clean exit, and I added a unit test that covers most of the code.

README.md
```diff
@@ -35,14 +35,15 @@ Flags:
...
-      --pipeLine.ingest.collector string       Ingest collector API  
       --pipeline.decode.aws string             aws fields  
       --pipeline.decode.type string            Decode type: aws, json, none (default "none")  
       --pipeline.encode.prom string            Prometheus encode API  
       --pipeline.encode.type string            Encode type: prom, none (default "none")  
       --pipeline.extract.aggregates string     Aggregates (see docs)  
       --pipeline.extract.type string           Extract type: aggregates, none (default "none")  
+      --pipeline.ingest.collector string       Ingest collector API  
       --pipeline.ingest.file.filename string   Ingest filename (file)  
+      --pipeline.ingest.kafka string           kafka parameters: brokers, topic, etc
```
eranra, line 46: please change to `Ingest Kafka API`

The details are in [docs/api.md](https://github.com/netobserv/flowlogs2metrics/pull/55/files#diff-9eddf4dc51bf6a0125ca7fb094468ad284112270385c41cebce4f8f0a29620ab)

cmd/flowlogs2metrics/main.go
```diff
@@ -129,7 +129,8 @@ func initFlags() {
 	rootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "error", "Log level: debug, info, warning, error")
 	rootCmd.PersistentFlags().StringVar(&config.Opt.PipeLine.Ingest.Type, "pipeline.ingest.type", "", "Ingest type: file, collector,file_loop (required)")
 	rootCmd.PersistentFlags().StringVar(&config.Opt.PipeLine.Ingest.File.Filename, "pipeline.ingest.file.filename", "", "Ingest filename (file)")
-	rootCmd.PersistentFlags().StringVar(&config.Opt.PipeLine.Ingest.Collector, "pipeLine.ingest.collector", "", "Ingest collector API")
+	rootCmd.PersistentFlags().StringVar(&config.Opt.PipeLine.Ingest.Collector, "pipeline.ingest.collector", "", "Ingest collector API")
+	rootCmd.PersistentFlags().StringVar(&config.Opt.PipeLine.Ingest.Kafka, "pipeline.ingest.kafka", "", "kafka parameters: brokers, topic, etc")
```
eranra, line 133: same ... actually this is the place to change and the README will update ;-)

pkg/api/ingest_kafka.go
```diff
@@ -0,0 +1,27 @@
...
+ *
+ */
+
+package api
+
+type IngestKafka struct {
+	Brokers          []string `yaml:"brokers" doc:"list of kafka broker addresses"`
+	Topic            string   `yaml:"topic" doc:"kafka topic to listen on"`
+	GroupId          string   `yaml:"groupid" doc:"separate groupid for each consumer on specified topic"`
+	GroupBalancers   []string `yaml:"groupbalancers" doc:"list of balancing stategies (range, roundRobin, rackAffinity)"`
```
eranra, line 24: I am asking myself if we need to change the yaml format to `groupBalancers` (i.e. use capital letter for the second word .... this applies to all ... not sure what make more sense ?)

pkg/api/ingest_kafka.go
```diff
@@ -0,0 +1,27 @@
...
+	StartOffset      string   `yaml:"startoffset" doc:"FirstOffset (least recent - default) or LastOffset (most recent) offset available for a partition"`
```
eranra, line 25: same

pkg/api/ingest_kafka.go
```diff
@@ -0,0 +1,27 @@
...
+	BatchReadTimeout int64    `yaml:"batchreadtimeout" doc:"how often (in milliseconds) to process input"`
```
eranra, line 26: same

docs/api.md
```diff
@@ -25,6 +25,18 @@ Following is the supported API format for the netflow collector:
...
 </pre>
+## Ingest Kafka API
+Following is the supported API format for the kafka ingest:
+
+<pre>
+ kafka:
+         brokers: list of kafka broker addresses
+         topic: kafka topic to listen on
+         groupid: separate groupid for each consumer on specified topic
+         groupbalancers: list of balancing stategies (range, roundRobin, rackAffinity)
```
eranra, line 36: s/stategies/strategies

pkg/pipeline/ingest/ingest_kafka.go
```diff
@@ -0,0 +1,172 @@
...
+	for _, gb := range jsonIngestKafka.GroupBalancers {
+		switch gb {
+		case "range":
+			groupBalancers = append(groupBalancers, &kafkago.RangeGroupBalancer{})
+		case "roundRobin":
+			groupBalancers = append(groupBalancers, &kafkago.RoundRobinGroupBalancer{})
+		case "rackAffinity":
+			groupBalancers = append(groupBalancers, &kafkago.RackAffinityGroupBalancer{})
+		default:
+			return nil, fmt.Errorf("the provided kafka balancer is not valid: %s", gb)
```
eranra, line 139: @KalmanMeth Maybe we need to set default if there was no input from the user?

pkg/pipeline/ingest/ingest_kafka_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+type fakeKafkaReader struct {
+	mock.Mock
+}
+
+var fakeRecord = []byte(`{"Bytes":20801,"DstAddr":"10.130.2.1","DstPort":36936,"Packets":401,"SrcAddr":"10.130.2.13","SrcPort":3100}`)
+
+var performedRead = false
+
+// ReadMessage runs in the kafka client thread, which blocks until data is avaialble.
```
eranra, line 168: s/avaialble/available

pkg/pipeline/ingest/ingest_kafka_test.go
```diff
@@ -0,0 +1,212 @@
...
+	// make the ingest thread exit
+	ingestKafka.exitChan <- true
+	time.Sleep(time.Second)
+}
+
+type fakeKafkaReader struct {
+	mock.Mock
+}
+
+var fakeRecord = []byte(`{"Bytes":20801,"DstAddr":"10.130.2.1","DstPort":36936,"Packets":401,"SrcAddr":"10.130.2.13","SrcPort":3100}`)
```
eranra, line 164: @KalmanMeth I miss a test that is able to show how we are moving from this []byte message in kafka into a json message ... this is part of the decode phase ... so maybe you can add this test in "pipeline_test". 

I want to make sure that we can handle this string --> json all the way from kafaka so that later we can continue with the regular pipeline phases of transformation etc ....

README.md
KalmanMeth: done

cmd/flowlogs2metrics/main.go
KalmanMeth: done

docs/api.md
KalmanMeth: done

pkg/pipeline/ingest/ingest_kafka.go
KalmanMeth: OK

pkg/pipeline/ingest/ingest_kafka.go
```diff
@@ -0,0 +1,172 @@
...
+
+	kafkaReader := kafkago.NewReader(kafkago.ReaderConfig{
+		Brokers:        jsonIngestKafka.Brokers,
+		Topic:          jsonIngestKafka.Topic,
+		GroupID:        jsonIngestKafka.GroupId,
+		GroupBalancers: groupBalancers,
+		StartOffset:    startOffset,
+	})
+	if kafkaReader == nil {
+		errMsg := "NewIngestKafka: failed to create kafkago reader"
```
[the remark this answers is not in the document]
KalmanMeth: OK

pkg/pipeline/ingest/ingest_kafka_test.go
KalmanMeth: done

PR #59, opened 2022-02-09
catch sigterm and exit go routines cleanly

KalmanMeth:


pkg/pipeline/utils/exit.go
```diff
@@ -0,0 +1,42 @@
...
+
+var (
+	registeredChannels []chan bool
+	chanMutex          sync.Mutex
+)
+
+func RegisterExitChannel(ch chan bool) {
+	chanMutex.Lock()
+	registeredChannels = append(registeredChannels, ch)
+	chanMutex.Unlock()
```
eranra, line 19: can you change to `defer chanMutex.Unlock()` instead of writing in the end ?

pkg/pipeline/ingest/ingest_collector.go
```diff
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"github.com/netobserv/flowlogs2metrics/pkg/api"
 	"github.com/netobserv/flowlogs2metrics/pkg/config"
+	exitUtils "github.com/netobserv/flowlogs2metrics/pkg/pipeline/utils"
```
eranra, line 27: maybe pipeline_utils or pUtils - if we add more utils it doesn't make sense to call that `exitUtils`

pkg/pipeline/utils/exit.go
```diff
@@ -0,0 +1,42 @@
...
+	go func() {
+		// wait for exit signal; then stop all the other go functions
+		sig := <-exitSigChan
+		log.Debugf("received exit signal = %v", sig)
+		chanMutex.Lock()
+		// exit signal received; stop other go functions
+		for _, ch := range registeredChannels {
+			ch <- true
+		}
+		chanMutex.Unlock()
```
eranra, line 38: defer unlock? instead of fixed call

pkg/pipeline/utils/exit.go
```diff
@@ -0,0 +1,42 @@
...
+		log.Debugf("exiting SetupElegantExit go function")
```
eranra, line 39: How do we know the order of things here ... I understand that you notify all registered channels ... and that they will exit ... but when will the entire process exit??? maybe you need to wait a sec or two to make sure all registeredChannels have a chance to exit

eranra (approved):
Very nice :-) I really like this ... it makes good order in the exit process  :-)

pkg/pipeline/utils/exit.go
KalmanMeth: done

pkg/pipeline/utils/exit.go
KalmanMeth: added a second in main before exit.

PR #60, opened 2022-02-09
added encode_json support plus test

KalmanMeth:


pkg/pipeline/encode/encode_json.go
```diff
@@ -0,0 +1,51 @@
...
+		if err != nil {
+			log.Errorf("encodeJson Decode: error marshalling a line: %v", err)
+			continue
+		}
+		out = append(out, string(line))
+	}
+	return out
+}
+
+// NewEndcodeJson create a new decode
```
eranra, line 47: s/"new decode"/"new encode"

eranra (approved):
@KalmanMeth can you add documentation for this. 

Actually should `config.Opt.PipeLine.Encode.Type` be part of the API now ??? I think it should and than each of the encode options should be part of enum

pkg/pipeline/encode/encode_json_test.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"encoding/json"
+	"github.com/netobserv/flowlogs2metrics/pkg/config"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func initNewEncodeJson(t *testing.T) Encoder {
+	newEncode, err := NewEncodeJson()
+	require.Equal(t, err, nil)
```
ronensc, line 29: I think the following is more expressive:
```go
require.NoError(t, err)
```

pkg/pipeline/encode/encode_json_test.go
```diff
@@ -0,0 +1,62 @@
...
+	require.Equal(t, 0, len(out))
+	in = append(in, map1)
+	in = append(in, map2)
+	in = append(in, map3)
+	out = encodeJson.Encode(in)
+	require.Equal(t, len(in), len(out))
+	outputString1, _ := json.Marshal(map1)
+	outputString2, _ := json.Marshal(map2)
+	outputString3, _ := json.Marshal(map3)
+	require.Equal(t, string(outputString1), out[0].(string))
```
ronensc, line 59: I think the test should use a predefined string of the expected json and compare the output to it, rather than using `json.Marshal()` as in the tested function
```go
	expected1 := `{"varInt":12,"varBool":true,"varString":"testString1"}`
	actual1 := out[0].(string)
	require.JSONEq(t, expected1, actual1)
```

KalmanMeth: We then have to worry about the way the map is converted into a string and the order of the elements. If we use the Marshal function, then we should get the same order as is done by the program.

pkg/pipeline/encode/encode_json_test.go
KalmanMeth: done

pkg/pipeline/encode/encode_json.go
KalmanMeth: done

pkg/pipeline/encode/encode_json.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+type encodeJson struct {
+}
+
+// Encode encodes json to string
+// All entries should be saved as strings
+func (e *encodeJson) Encode(in []config.GenericMap) []interface{} {
+	out := make([]interface{}, 0)
+	for _, metric := range in {
```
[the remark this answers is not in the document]
KalmanMeth: done

KalmanMeth:
I changed the name of the files to encode_json_to_string. The input is json (configMap), but the difference is that it outputs a string for each configMap. There is no API detail needed since there are no parameters that the user needs to provide.

pkg/pipeline/encode/encode_json_test.go
KalmanMeth: done

KalmanMeth:
In light of our discussion, I changed the name back to encode_json.go.
I also added a few sentences in the README to document the json encoder.

ronensc (approved):

PR #64, opened 2022-02-10
initial loki integration

eranra:


pkg/pipeline/write/write_loki.go
```diff
@@ -0,0 +1,252 @@
...
+	for k, v := range l.apiConfig.StaticLabels {
+		labels[k] = v
+	}
+
+	l.addNonStaticLabels(record, labels)
+
+	// Remove labels and configured ignore list from record
+	ignoreList := append(l.apiConfig.IgnoreList, l.apiConfig.Labels...)
+	for _, label := range ignoreList {
+		delete(record, label)
```
jotak, line 123: Just a remark here: the input map is mutated. In goflow-kube that was ok because the Loki write was a terminal action. Maybe it's the same here. But I prefer to catch your attention here. Also, depending on how we implement the pipeline fork mechanism later, it may become an issue.

pkg/pipeline/pipeline_test.go
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package pipeline
```
KalmanMeth, line 18: Why is this test in `pipeline` module and not in `write` module?

eranra:
image

pkg/pipeline/write/write_loki.go
eranra: @jotak with the new version of the PR (that sends the records over a channel to Loki writer) I think that this is ok. This record is now a copy of the original one and there is no influence on the continuation of the pipeline so this should be ok.

Thanks for raising that :-)

pkg/pipeline/write/write_loki.go
```diff
@@ -0,0 +1,283 @@
...
+		return cfg, fmt.Errorf("failed to parse client URL: %w", err)
+	}
+	cfg.URL = clientURL
+	return cfg, nil
+}
+
+func (l *Loki) ProcessRecord(record config.GenericMap) error {
+	if !l.IsReady() {
+		return errors.New("WriteLoki is not ready")
+	}
```
jpinsonneau, line 111: It seems you can remove this; `IsReady` function and `ready` field are always `true` in your implementation

jpinsonneau (approved):

pkg/pipeline/write/write_loki.go
```diff
@@ -0,0 +1,283 @@
...
+
+const channelSize = 1000
+
+// Loki record writer
+type Loki struct {
+	lokiConfig loki.Config
+	apiConfig  api.WriteLoki
+	client     emitter
+	timeNow    func() time.Time
+	ready      bool
```
mariomac, line 57: In a later PR I suggested to remove the "ready" field: https://github.com/netobserv/goflow2-kube-enricher/pull/29/files#diff-f0ee701b73c37f99481d73281b339c6abe6d17c2d5a1f494f80edd16dacb7e79L59-L72

the reason is that this field does not provide any changing state and it is only used to override the behavior with the "empty loki" collector, which is only used for testing. I think we shouldn't permeate testing behavior into production code.

We can just remove this `ready` field and `IsReady` method.

pkg/pipeline/write/write_loki.go
mariomac: ➕1

eranra: removed, done

jotak (approved):
LGTM
thanks @eranra !

KalmanMeth (approved):

pkg/pipeline/pipeline.go
```diff
@@ -156,8 +158,9 @@ func (p Pipeline) Process(entries []interface{}) {
 		flowEntry = transform.ExecuteTransforms(p.Transformers, entry)
 		transformed = append(transformed, flowEntry)
 	}
-	log.Debugf("transformed = %v", transformed)
+
+	_ = p.Writer.Write(transformed)
+
```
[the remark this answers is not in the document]
eranra: @KalmanMeth as we agreed (in the conversation we had) we will merge this PR and work to rename some of the pipeline stages ... maybe duplicate the writing phase and maybe in the work you will do with @mariomac we will decide differently.

@mariomac FYI: the current change is actually a quick and dirty "fork" we need something better going forward. This solves and allows sending flow-logs to Loki and metrics to Prometheus but we will want something better ASAP.

PR #68, opened 2022-02-14
Add dedupe

### LINKED ISSUE #13: dedupe pipeline rules in confGenerator
When defining rules in various stages of the pipeline some of the rules might overlap, this is true for stages such as transformation and aggregation as well as others. It will be efficient if confGenerator has deduplication functionality to remove all of those redundant rules. This will improve efficiency and make the rules more readable and accurate

ronensc:
fixes #13

### DIFF
--- a/pkg/confgen/dedup.go
+++ b/pkg/confgen/dedup.go
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C) 2021 IBM, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package confgen
+
+import (
+	"github.com/netobserv/flowlogs2metrics/pkg/api"
+	"github.com/netobserv/flowlogs2metrics/pkg/pipeline/extract/aggregate"
+	log "github.com/sirupsen/logrus"
+	"reflect"
+)
+
+func (cg *ConfGen) dedupe() {
+	cg.transformRules = dedupeNetworkTransformRules(cg.transformRules)
+	cg.aggregateDefinitions = dedupeAggregateDefinitions(cg.aggregateDefinitions)
+}
+
+type void struct{}
+var voidMember void
+
+func dedupeNetworkTransformRules(rules api.NetworkTransformRules) api.NetworkTransformRules {
+	// There are no built-in sets in go
+	//https://stackoverflow.com/a/34020023/2749989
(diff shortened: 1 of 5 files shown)


pkg/confgen/dedup.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (cg *ConfGen) dedupe() {
+	cg.transformRules = dedupeNetworkTransformRules(cg.transformRules)
+	cg.aggregateDefinitions = dedupeAggregateDefinitions(cg.aggregateDefinitions)
+}
+
+type void struct{}
+var voidMember void
+
+func dedupeNetworkTransformRules(slice api.NetworkTransformRules) api.NetworkTransformRules {
```
eranra, line 35: s/slice/rules

pkg/confgen/dedup.go
```diff
@@ -0,0 +1,74 @@
...
+	// There are no built-in sets in go
+	//https://stackoverflow.com/a/34020023/2749989
+	uniqueSet := make(map[api.NetworkTransformRule]void)
+	var dedpueSlice []api.NetworkTransformRule
+	for i, item := range slice {
```
eranra, line 40: s/item/rule

pkg/confgen/dedup.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		uniqueSet[item] = voidMember
+		dedpueSlice = append(dedpueSlice, item)
+	}
+	return dedpueSlice
+}
+
+// dedupeAggregateDefinitions is inefficient because we can't use a map to look for duplicates.
+// The reason is that aggregate.Definition is not hashable due to its By field which is a slice.
+func dedupeAggregateDefinitions(slice aggregate.Definitions) aggregate.Definitions {
```
eranra, line 54: s/slice/aggregateDefinitions

pkg/confgen/dedup.go
```diff
@@ -0,0 +1,74 @@
...
+	var dedpueSlice []aggregate.Definition
+	for i, item := range slice {
+		if containsAggregateDefinitions(dedpueSlice, item) {
+			// duplicate item
+			log.Debugf("Remove duplicate AggregateDefinitions %v at index %v", item, i)
```
eranra, line 59: s/itme/aggregateDefinition

pkg/confgen/dedup.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func containsAggregateDefinitions(slice []aggregate.Definition, searchItem aggregate.Definition) bool {
+	for _, item := range slice {
+		if reflect.DeepEqual(item, searchItem) {
+			return true
+		}
+	}
+	return false
+}
```
eranra, line 74: @ronensc the second implementation (this one) feels like it can be a "generic" solution. Can you reuse the code also for dedupeNetworkTransformRules ... and so we will have one implementation of getting the unique items used twice. I know that the first implementation is more efficient (no reflection) but this is totally not relevant for this use-case (we have very small amount of items, and the code is not working in real-time)

pkg/confgen/dedup.go
ronensc: done

pkg/confgen/dedup.go
ronensc: done

pkg/confgen/dedup.go
ronensc: done

PR #85, opened 2022-02-17
JSON decode: keep value types, do not convert to string

eranra:


eranra:
> To test this, we have to change in the tests (utils, transform_*_test.go the appropriate fields to integers (instead of strings) and verify that everything still works.

@KalmanMeth I think that test like https://github.com/netobserv/flowlogs2metrics/pull/85/files#diff-06c34b5e9ba491d8ba67f577931652caacddf9fa65c3a177a2d5bafb0ecae68cR50 is showing exactly that

KalmanMeth:
There was some failure occurring when we previously used integers rather than strings, and it had to do with converting json to a map and back. Try changing the tests in transform to use integers for port numbers, bytes, etc, and see if they pass.

eranra:
@KalmanMeth I changed the tests to use numeric (int) values for ports and protocol (as a number) and the code is working as expected == parsing the ports correctly and the protocol. 

I looked at the code and this looks like this:: `portNumber, err := strconv.Atoi(fmt.Sprintf("%v", outputEntries[rule.Input]))`  --- this makes things very generic. We do not care if this is a string or a number the result will always be the port number as int.

pkg/pipeline/decode/decode_json_test.go
```diff
@@ -47,9 +47,24 @@ func TestDecodeJson(t *testing.T) {
...
+	decodeJson := newDecode.(*decodeJson)
+	inputString1 := "{\"unixTime\": 1645104030 }"
+	var in []interface{}
+	var out []config.GenericMap
+	out = decodeJson.Decode(in)
+	require.Equal(t, 0, len(out))
+	in = append(in, inputString1)
+	out = decodeJson.Decode(in)
+	require.Equal(t, len(out), 1)
+	// verify that all items come back as strings
```
KalmanMeth, line 68: Delete (or fix) this comment.

KalmanMeth (approved):

pkg/pipeline/decode/decode_json_test.go
eranra: fixed

PR #87, opened 2022-02-20
add health server (http)

eranra:


KalmanMeth (approved):

pkg/health/health.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func NewHealthServer() *Server {
+
+	handler := healthcheck.NewHandler()
+	address := net.JoinHostPort(defaultServerHost, config.Opt.Health.Port)
+
+	handler.AddLivenessCheck("goroutine-threshold", healthcheck.GoroutineCountCheck(100))
+	handler.AddLivenessCheck("GCMaxPauseCheck", healthcheck.GCMaxPauseCheck(60*time.Second))
+	handler.AddReadinessCheck("HTTPGetCheck", healthcheck.HTTPGetCheck("https://quay.io", 60*time.Second))
```
jotak, line 51: About the HTTP Get check, what if the pod is started in a disconnected environment?
Overall I think these health checks are interesting but I wonder if they could create more harm in some situations, even considering the two liveness checks above. We must keep in mind that when liveness probe is defined in kube, unhealthy pods are restarted, which is not always the best remedy to the symptoms.
So there are several options, we could make these checks more configurable, or simply being able to turn them on/off (which we can do just on the operator side, as it's where the kube probes are defined), or last option just remove them and keep health check more minimalist (ie. just return 200 when server is ready, no more)
Thoughts?

jotak: I think in any case we should be able to turn off kube probes from the operator, through CR config

pkg/health/health.go
```diff
@@ -0,0 +1,61 @@
...
+type Server struct {
+	handler healthcheck.Handler
+	address string
+}
+
+func (hs *Server) Serve() {
+	for {
+		err := http.ListenAndServe(hs.address, hs.handler)
+		log.Errorf("http.ListenAndServe error %v", err)
+		time.Sleep(60 * time.Second)
```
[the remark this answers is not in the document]
eranra: @KalmanMeth no ... same as in `startPrometheusInterface` 

@KalmanMeth --> Adding exit to those HTTP servers is a complicated task in golang prior to 1.18+ and I do not see a reason (at this stage) to add this complexity to the code. If you really want to do that, we can but I would like to avoid if possilbe.

pkg/health/health.go
eranra: @jotak I changed the code just to check that the "pipeline" is ready and alive based on a flag ... going forward we will add more and more validations that things are really alive and ready

jotak (approved):

PR #91, opened 2022-02-20
add k8s owner

eranra:


eranra:
image

Deamonset owner

eranra:
image

stateful owner

eranra:
More types ::

image

eranra:
@jotak if you can review ... I think that this is what we need for owner

jotak (approved):
thanks, looks good to me!

pkg/pipeline/transform/kubernetes/kubernetes.go
```diff
@@ -189,7 +244,7 @@ func (k KubeData) InitFromConfig(kubeConfigPath string) error {
 	return nil
 }
 
-func (k KubeData) initInformers(client kubernetes.Interface) error {
+func (k *KubeData) initInformers(client kubernetes.Interface) error {
 	//defer close(stopChan)
```
KalmanMeth, line 248: Is this line no longer relevant? Delete.

jotak:
actually just a small remark, it seems like `Type` is all lower case whereas OwnerType is CamelCase, would it be easy to also have CamelCase for Pod / Service / Node (so that it matches the k8s kinds) ?

pkg/pipeline/transform/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+type InformerInterface interface {
+	cache.SharedInformer
+	AddIndexers(indexers cache.Indexers) error
+	GetIndexer() cache.Indexer
+}
+
+func (indexMock *IndexerMock) ByIndex(indexName, indexedValue string) ([]interface{}, error) {
+	//indexMock.Mock.Called(indexName, indexedValue)
```
KalmanMeth, line 46: Delete unused function call?

KalmanMeth (approved):

pkg/pipeline/transform/kubernetes/kubernetes.go
eranra: removed, done

pkg/pipeline/transform/kubernetes/kubernetes_test.go
eranra: removed, done

pkg/pipeline/transform/kubernetes/kubernetes.go
eranra: done

eranra:
> actually just a small remark, it seems like `Type` is all lower case whereas OwnerType is CamelCase, would it be easy to also have CamelCase for Pod / Service / Node (so that it matches the k8s kinds) ?

I changed to CamelCase to match :-)

PR #95, opened 2022-02-22
Change gauges to counters

### LINKED ISSUE #61: Prometheus: change existing metrics from gauges to counters

ronensc:
Fixes #61

I implemented a variation of option (1). Rather than adding a delta field as suggested, I added a slice with the raw values since the last report. This approach allowed my to add a missing part in the Histogram support.

### DIFF
--- a/pkg/pipeline/encode/encode_prom.go
+++ b/pkg/pipeline/encode/encode_prom.go
@@ -135,11 +135,14 @@ func (e *encodeProm) EncodeMetric(metric config.GenericMap) []interface{} {
 			mInfo.promGauge.With(entryLabels).Set(valueFloat)
 			cEntry.PromMetric.promGauge = mInfo.promGauge
 		case api.PromEncodeOperationName("Counter"):
-			mInfo.promCounter.With(entryLabels).Add(valueFloat)
+			for _, v := range metric["recentRawValues"].([]float64) {
+				mInfo.promCounter.With(entryLabels).Add(v)
+			}
 			cEntry.PromMetric.promCounter = mInfo.promCounter
 		case api.PromEncodeOperationName("Histogram"):
-			// TODO: Need to perform proper function for histogram
-			//mInfo.promHist.With(entryLabels).Set(valueFloat)
+			for _, v := range metric["recentRawValues"].([]float64) {
+				mInfo.promHist.With(entryLabels).Observe(v)
+			}
 			cEntry.PromMetric.promHist = mInfo.promHist
 		}
 	}
--- a/pkg/pipeline/extract/aggregate/aggregate.go
+++ b/pkg/pipeline/extract/aggregate/aggregate.go
@@ -125,6 +125,7 @@ func (aggregate Aggregate) UpdateByEntry(entry config.GenericMap, normalizedValu
 
 	if operation == OperationCount {
 		groupState.value = float64(groupState.count + 1)
+		groupState.RecentRawValues = append(groupState.RecentRawValues, 1)
 	} else {
 		if recordKey != "" {
 			value, ok := entry[recordKey]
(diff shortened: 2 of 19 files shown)


ronensc:
@eranra I run `make local-deploy` with an updated image and both prometheus and Grafna show changes in the metrics over time

pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -55,6 +55,7 @@ type Definition struct {
 
 type GroupState struct {
 	normalizedValues NormalizedValues
+	RecentRawValues  []float64
```
eranra, line 58: Why do we need a list (sorry to be cheap here) .... this will require an unneeded (and uncontrolled) amount of extra memory. Can we have just one value float64 value and add `value` all the time. In the end of the day, this is what you are doing https://github.com/netobserv/flowlogs2metrics/pull/95/files#diff-30d11303a4ad1ec82e4f39e3ae6dc7d044c8dd370c536a98ba90febbfe2d425aR139 ... it is not that we are adding those value with the correct timestamp or something like that ... so why not just add them here?

pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -179,10 +182,13 @@ func (aggregate Aggregate) GetMetrics() []config.GenericMap {
 			"by":                                 strings.Join(aggregate.Definition.By, ","),
 			"aggregate":                          string(group.normalizedValues),
 			"value":                              fmt.Sprintf("%f", group.value),
+			"recentRawValues":                    group.RecentRawValues,
 			"count":                              fmt.Sprintf("%d", group.count),
 			aggregate.Definition.Name + "_value": fmt.Sprintf("%f", group.value),
 			strings.Join(aggregate.Definition.By, "_"): string(group.normalizedValues),
 		})
+		// Once reported, we reset the raw values accumulation
+		group.RecentRawValues = make([]float64, 0)
```
eranra, line 191: @ronensc Reminder- reminder ::: I think that we do not have a problem with synchronization ... but you had thoughts that we might ... do you think we should add a semaphore to protect `RecentRawValues` ... actually thinking on this ... when @KalmanMeth will change the pipeline to all channels this will be totally not needed so maybe the only thing I can ask is that you will try to **create a test** to expose a potential race problem ... so we will know that it happens.

eranra (approved):
@ronensc thanks, this looks very nice and clean to me ... I think we can merge as is, I added some comments bellow :-)

pkg/pipeline/extract/aggregate/aggregate.go
ronensc: For the case of `Counters`, you are right. We can keep only the sum of the slice, rather than the raw values. But, it won't work for case of `Histograms`. In Histograms, we must use the raw values so they could be put in the right buckets. That's why I chose to use a slice.

pkg/pipeline/extract/aggregate/aggregate.go
ronensc: @eranra Looking at how the code is implemented currently, I don't see a synchronization problem because all the interaction with `RecentRawValues` is done in the same goroutine/thread.
Anyway, I'll try to add a test that checks that it is populated correctly.

pkg/pipeline/encode/encode_prom.go
```diff
@@ -135,11 +135,14 @@ func (e *encodeProm) EncodeMetric(metric config.GenericMap) []interface{} {
...
+			for _, v := range metric["recentRawValues"].([]float64) {
+				mInfo.promCounter.With(entryLabels).Add(v)
+			}
 			cEntry.PromMetric.promCounter = mInfo.promCounter
 		case api.PromEncodeOperationName("Histogram"):
-			// TODO: Need to perform proper function for histogram
-			//mInfo.promHist.With(entryLabels).Set(valueFloat)
+			for _, v := range metric["recentRawValues"].([]float64) {
+				mInfo.promHist.With(entryLabels).Observe(v)
+			}
```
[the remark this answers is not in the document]
eranra: :-) looking forward to more nice code here :-)

PR #104, opened 2022-02-24
Fix basic pipeline test

mariomac:
There was a couple of minor, unnoticed mistakes, that prevented the pipeline transformers from being loaded.

This test also checks that the transformation stage has been properly applied.


pkg/pipeline/pipeline_test.go
```diff
@@ -104,4 +91,49 @@ func Test_SimplePipeline(t *testing.T) {
...
+		"fl2m_dstPort": float64(36936),
+		"fl2m_packets": float64(400),
+		"fl2m_srcAddr": "10.130.2.13",
+		"fl2m_srcPort": float64(3100),
+	}, writer.PrevRecords[0])
+}
+
+func BenchmarkPipeline(b *testing.B) {
+	t := &testing.T{}
+	loadGlobalConfig(t)
```
eranra, line 110: @mariomac can you look on https://github.com/netobserv/flowlogs-pipeline/blob/main/Makefile#L88 --- maybe we can somehow improve this >>> ???? I started to create some benchmark area for the project and I totally agree we need to improve that

eranra: @mariomac In any event, can we agree to create dedicated go files just for benchmark and split from the rest of the tests so we can run those stand-alone??

mariomac: With respect to your second question, we could do that if you prefer it. Anyway benchmarks are not run by default even if they are in the same file as the tests. If you mean skipping tests when you run benchmarks, you can add the `-test.run=^$` argument to the `go test` command to skip any test.

But I'm fine if you feel it's better organizing everyting in the same benchmarks file.

mariomac: @eranra I see what's happening with the benchmark in `benchmark_test.go`. It basically runs many times the whole command, so it would measure basically the performance of starting the FL2M process and finishing it, not the actual FL2M processing performance.

Go benchmarks are more a sort of "micro-benchmarks" aimed to test some parts of the code, and that's why they are usually located in the test files of the components that they are benchmarking. For example, in the benchmark of this PR, it just tests the time of sending and processing a file of ~5000 flows with a very simple dummy pipeline (no real ingest, no real writing...), but it allows us measuring the impact of the sequential vs parallel pipeline mechanism.

I'd suggest to (in another PR to not loose the focus of our current task) remove the current `benchmark_test.go` and prepare some benchmark that:

1. starts the FL2M service with a real ingester
2. spins up a client that is able to send real IPFIX flows ([we did one using a VM library in our goflow-kube](https://github.com/netobserv/goflow2-kube-enricher/blob/24fc6050d63bfb913269af4e8675cdf66ceef310/test/integration/service.go#L124-L160))
3. Measures how many flows it's able to process in a given amount of time. E.g. the last stage of the pipeline could be just a counter.

In the future, this could be improved e.g. spinning parallel clients

pkg/pipeline/pipeline_test.go
```diff
@@ -18,15 +18,18 @@
 package pipeline
 
 import (
-	"github.com/json-iterator/go"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+
+	jsoniter "github.com/json-iterator/go"
 	"github.com/netobserv/flowlogs2metrics/pkg/config"
```
KalmanMeth, line 26: goland editor will re-order these imports to be in alphabetical order.

pkg/pipeline/pipeline_test.go
```diff
@@ -102,6 +91,54 @@ func Test_SimplePipeline(t *testing.T) {
...
+	if t.Failed() {
+		b.Fatalf("unexpected error loading config")
+	}
+	for n := 0; n < b.N; n++ {
+		b.StopTimer()
+		p, err := NewPipeline()
+		if err != nil {
+			t.Fatalf("unexpected error %s", err)
+		}
+		b.StartTimer()
```
KalmanMeth, line 125: Where is the timing information used?

KalmanMeth (approved):

pkg/pipeline/pipeline_test.go
mariomac: In `netobserv` we usually sort the imports with `goimports`, which is independent of the IDE (many team members use VScode). It can be configured to be used from Goland/IDEA too: 

image

But if this is an inconvenience for you I can adapt to use the default Goland. WDYT @jotak @jpinsonneau @OlivierCazade ?

pkg/pipeline/pipeline_test.go
mariomac: `StopTimer` and `StartTimer` are used to exclude from the benchmark some parts of the code that you don't want to measure. In this case we don't want to measure the Pipeline creation time but just the amount of metrics that you can forward.

mariomac:
superseeded by https://github.com/netobserv/flowlogs-pipeline/pull/105

pkg/pipeline/pipeline_test.go
jotak: I'm also using `goimports` (configured in vscode)

PR #105, opened 2022-02-25
Asynchronous pipeline, initial integration

mariomac:
Wraps the current interfaces into channel loops that are triggered asynchronously in goroutines and communicate each step of the pipeline.

This PR is a starting point to other 2 branches:

- Incorporating the components and functionalities from goflow-kube into flowlogs-pipeline.
- Rework some interfaces to work by themselves as stages of the pipeline, without requiring to wrap them into loop functions (ongoing in branch: https://github.com/mariomac/flowlogs-pipeline/tree/async-pipeline).


jpinsonneau (approved):
LGTM

I have put some questions but not necessary things to update so feel free to merge

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/heptiolabs/healthcheck v0.0.0-20211123025425-613501dd5deb
 	github.com/ip2location/ip2location-go/v9 v9.2.0
 	github.com/json-iterator/go v1.1.12
+	github.com/mariomac/go-pipes v0.1.0
```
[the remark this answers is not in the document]
mariomac: Yes. Moving it. I wasn't unsure if it was going to netobserv or to any other organization, as it is not directly related with network observability.

pkg/pipeline/pipeline.go
```diff
@@ -149,31 +156,81 @@ func NewPipeline() (*Pipeline, error) {
...
+
+	// starting the graph
+	for _, s := range p.start {
+		s.Start()
+	}
+
+	// blocking the execution until the graph terminal stages to end before
+	for _, t := range p.terminal {
+		<-t.Done()
+	}
```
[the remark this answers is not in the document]
mariomac: No, this jus stops the execution of this goroutine until the pipeline has ended (which at the moment would be the same as "wait forever", since we haven't yet implemented any mechanism co cancel the execution).

pkg/pipeline/pipeline.go
```diff
@@ -43,6 +45,9 @@ type Pipeline struct {
 	Extractor    extract.Extractor
 	Encoder      encode.Encoder
 	IsRunning    bool
+
+	start    []*node.Init
+	terminal []*node.Terminal
```
KalmanMeth, line 50: @eranra @mariomac We should discuss the `Terminal` node (e.g. an export function) and whether to do away with the notions of `Transform`, `Extract`, etc that we currently have.

pkg/pipeline/pipeline.go
```diff
@@ -149,31 +156,81 @@ func NewPipeline() (*Pipeline, error) {
 		Writer:       writer,
 	}
 
+	// TODO: all this apply* functions can be removed if we change the Ingester, Transformer, etc...
+	// interfaces and use them directly here
```
KalmanMeth, line 160: @eranra @mariomac Let's discuss this in our next meeting. We need to decide on the order of merging PRs to make this transition.

pkg/pipeline/pipeline.go
```diff
@@ -149,31 +156,81 @@ func NewPipeline() (*Pipeline, error) {
...
+	ingests := node.AsInit(p.Ingester.Ingest)
+	decodes := node.AsMiddle(p.applyDecode)
+	transforms := node.AsMiddle(p.applyTransforms)
+	writes := node.AsTerminal(p.applyWrite)
+	extracts := node.AsMiddle(p.applyExtract)
+	encodes := node.AsTerminal(p.applyEncode)
+
```
KalmanMeth, line 167: At present, @eranra uses the same transformed data to be fed into `write` and `extract` (which then feeds into `encode`). We probably want to combine `write` and `encode` to something like `export`. We may need to maintain the notion of `decode` since it takes a different format for its input than the other types of stages.

mariomac: I see the point. I think it's an interesting discussion.

eranra: @mariomac we talked about a wrapper function that will allow using stages that do have output as `AsTerminal` phases. So that function will just drop the output of the phase so it will fit when needed. 

An example is the Prometheus encode phase ... it does have output and in some cases, for example, debug you want to push the output into another stdout write phase but for other cases, this should be a termination phase ... so with such wrapping, we can support both.

pkg/pipeline/pipeline_test.go
```diff
@@ -75,33 +79,67 @@ pipeline:
...
+		"fl2m_bytes":   float64(20800),
+		"fl2m_dstAddr": "10.130.2.2",
+		"fl2m_dstPort": float64(36936),
+		"fl2m_packets": float64(400),
+		"fl2m_srcAddr": "10.130.2.13",
+		"fl2m_srcPort": float64(3100),
+	}, writer.PrevRecords[0])
+}
+
+func BenchmarkPipeline(b *testing.B) {
```
eranra, line 112: @mariomac can you move the `Benchmark` into a separate file so we will be able to run just benchmarks and it will be easy to see that something is benchmarking (in compared to a test). THX

pkg/pipeline/pipeline.go
```diff
@@ -149,31 +156,81 @@ func NewPipeline() (*Pipeline, error) {
...
+	ingests.SendsTo(decodes)
+	decodes.SendsTo(transforms)
+	transforms.SendsTo(writes, extracts)
+	extracts.SendsTo(encodes)
```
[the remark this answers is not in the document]
mariomac: Yes, this will come in a later PR

pkg/pipeline/pipeline.go
```diff
@@ -149,31 +156,81 @@ func NewPipeline() (*Pipeline, error) {
 		Writer:       writer,
 	}
 
+	// TODO: all this apply* functions can be removed if we change the Ingester, Transformer, etc...
```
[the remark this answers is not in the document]
mariomac: Yes. I preferred to do multiple, small, incremental PRs rather than a big PR

pkg/pipeline/pipeline_test.go
eranra: ok ... NP

pkg/pipeline/ingest/ingest_file_chunks.go
```diff
@@ -0,0 +1,85 @@
+/*
```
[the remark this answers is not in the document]
mariomac: I observed that RH Openshift organization is usually not adding copyright to their files... so I just kept the original copyright from the `ingest_file.go` that I used as a starting point.

Anyway, RH is property of IBM :D

pkg/pipeline/ingest/ingest_kafka.go
```diff
@@ -91,7 +90,8 @@ func (ingestK *ingestKafka) processLogLines(process ProcessFunction) {
 		case <-time.After(duration): // Maximum batch time for each batch
 			// Process batch of records (if not empty)
 			if len(records) > 0 {
-				process(records)
+				log.Debugf("ingestKafka sending %d records", len(records))
+				out <- records
```
eranra, line 94: @KalmanMeth @mariomac do we even need this timer thing .... now that we have channels in between maybe we can just send the records every time we get one ... why do we need batch all together ?>

pkg/pipeline/pipeline.go
```diff
@@ -43,13 +45,26 @@ const (
...
+	startNodes    []*node.Init
+	terminalNodes []*node.Terminal
+	IsRunning     bool
+	// TODO: this field is only used for test verification. We should rewrite the build process
+	// to be able to remove it from here
+	pipelineStages []*pipelineEntry
+}
+
+// builder stores the information that is only required during the build of the pipeline
+type builder struct {
```
eranra, line 57: @mariomac do you want to move the` builder` and all the functions into a dedicated file ... this file is packed and it make sense to split

pkg/pipeline/ingest/ingest_kafka.go
mariomac: You are right. Sending batches will anyway improve a bit the performance because you minimize the communication via channels but I think is worth giving a try and see if the simpler code and testing pays the overhead.

In a later PR I have fixed the timer and also made that, when the batch length reaches a given size, it is forwarded without waiting for the timer. We can try removing batches in yet another PR and evaluate the impact in the pipeline benchmark. That would also allow removing the "file_chunk" functionality.

pkg/pipeline/ingest/ingest_file_chunks.go
```diff
@@ -0,0 +1,85 @@
...
+	"os"
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/utils"
+	log "github.com/sirupsen/logrus"
+)
+
+const chunkLines = 100
+
+// FileChunks ingest entries from a file and resends them in chunks of fixed number of lines.
```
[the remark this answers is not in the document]
mariomac: What do you mean by "readme test"? An entry in the README documentation? I see that the ingesters are not documented there

pkg/pipeline/ingest/ingest_file.go
```diff
@@ -68,7 +76,21 @@ func (ingestF *IngestFile) Ingest(process ProcessFunction) {
 			case <-ticker.C:
 				log.Debugf("ingestFile; for loop; before process")
```
KalmanMeth, line 77: Technically we no longer call `process()`, so perhaps the comment should be updated to remove reference to `process`.

pkg/pipeline/pipeline_builder.go
```diff
@@ -0,0 +1,335 @@
...
+	var stage interface{}
+	// TODO: modify all the types' interfaces to not need to write loops here, the same
+	// as we do with Ingest
+	switch pe.stageType {
+	case StageIngest:
+		init := node.AsInit(pe.Ingester.Ingest)
+		b.startNodes = append(b.startNodes, init)
+		stage = init
+	case StageWrite:
+		term := node.AsTerminal(func(in <-chan []config.GenericMap) {
```
KalmanMeth, line 174: It might be convenient to allow write_stdout at any step of the pipeline, allow it multiple times, and allow it to have followers. This way we can debug and have write_stdout after each other stage (decode, transform1, transform2, etc) in a pipeline. This is related to what @eranra mentioned with the possibility of (write, export) nodes being sometimes terminal nodes and sometimes middle nodes.

pkg/pipeline/pipeline_builder.go
```diff
@@ -0,0 +1,335 @@
...
+	}, nil
+}
+
+func (b *builder) getStageNode(pe *pipelineEntry) (interface{}, error) {
+	if stg, ok := b.createdStages[pe.configStage.Name]; ok {
+		return stg, nil
+	}
+	var stage interface{}
+	// TODO: modify all the types' interfaces to not need to write loops here, the same
+	// as we do with Ingest
```
KalmanMeth, line 167: I assume this is to be done in the next PR?

pkg/pipeline/pipeline_builder.go
```diff
@@ -0,0 +1,335 @@
...
+		stage = node.AsMiddle(func(in <-chan []config.GenericMap, out chan<- []config.GenericMap) {
+			for i := range in {
+				out <- pe.Encoder.Encode(i)
+			}
+		})
+	case StageTransform:
+		stage = node.AsMiddle(func(in <-chan []config.GenericMap, out chan<- []config.GenericMap) {
+			for i := range in {
+				out <- transform.ExecuteTransform(pe.Transformer, i)
+			}
```
KalmanMeth, line 197: `transform` could now be the same structure as `encode` or `extract`, so we can do away with the `ExecuteTransform()` intermediary and call `pe.Transformer.Transform()` directly.

pkg/pipeline/pipeline_builder.go
mariomac: We could modify the interface so it can fit into a middle node instead of a terminal mode. I'd do it in another PR, to not loose the focus on the goal of this PR.

pkg/pipeline/pipeline_builder.go
mariomac: I was considering to do it in yet another PR, after Goflow Kube is integrated with Flowlows Pipeline, we verify that works and then can cleanup some TODOs.

pkg/pipeline/pipeline_builder.go
mariomac: Currently, this would involve adding the loop inside Execute transform in any transform implementation (as the `Transform` interface accepts and returns a single `GenericMap` and here we are playing with arrays).

I was thinking on address your proposal in a later PR, where we put in common all the functional interfaces and change them by input/output channels.

pkg/pipeline/pipeline_builder.go
```diff
@@ -0,0 +1,335 @@
...
+	"fmt"
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/decode"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/encode"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/extract"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/ingest"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/transform"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/write"
+	"github.com/netobserv/gopipes/pkg/node"
```
KalmanMeth, line 14: Do we have to include code review of `github.com/netobserv/gopipes/pkg/node` in this PR?

pkg/pipeline/pipeline.go
mariomac: Now I understand better this proposal after another comment from @KalmanMeth .

We can do two things:
- Change the "write_stdout" interface to work as middle node: receives an entry, submits it, forwards it.

image

- Don't do nothing by now. Just consider that you can currently attach terminal nodes to any stage of the pipeline:
 
image

pkg/pipeline/pipeline_builder.go
mariomac: Actually it is merged in its original repo. I'd maybe do it apart. Feel free to open a PR in gopipes with some suggestions, questions, or comments, and I can address them.

KalmanMeth (approved):

PR #106, opened 2022-02-27
Add a test for extract_aggregate

ronensc:
This is a follow up to #95 as part of #61

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ clean: ## Clean
 # note: to review coverage execute: go tool cover -html=/tmp/coverage.out
 .PHONY: test
 test: validate_go ## Test
-	go test -p 1 -race -covermode=atomic -coverprofile=/tmp/coverage.out ./...
+	go test -p 1 -race -coverpkg=./... -covermode=atomic -coverprofile=/tmp/coverage.out ./...
 
 # note: to review profile execute: go tool pprof -web /tmp/flowlogs-pipeline-cpu-profile.out (make sure graphviz is installed)
 .PHONY: benchmarks
--- a/pkg/pipeline/extract/extract_aggregate_test.go
+++ b/pkg/pipeline/extract/extract_aggregate_test.go
@@ -55,10 +55,17 @@ aggregates:
   - service
   operation: count
   recordkey: ""
+
 - name: bandwidth_sum
   by:
   - service
   operation: sum
+  recordkey: bytes
+
+- name: bandwidth_max
+  by:
+  - service
+  operation: max
   recordkey: bytes
 `
 	var err error
@@ -88,10 +95,12 @@ aggregates:
 				{"service": "tcp", "bytes": 2.0},
 			},
 			expectedAggs: []config.GenericMap{
-				createAgg("bandwidth_sum", "bytes", "service", "http", aggregate.OperationSum, 30, 2, []float64{10.0, 20.0}),
-				createAgg("bandwidth_sum", "bytes", "service", "tcp", aggregate.OperationSum, 3, 2, []float64{1.0, 2.0}),
 				createAgg("bandwidth_count", "", "service", "http", aggregate.OperationCount, 2, 2, []float64{1.0, 1.0}),
 				createAgg("bandwidth_count", "", "service", "tcp", aggregate.OperationCount, 2, 2, []float64{1.0, 1.0}),
+				createAgg("bandwidth_sum", "bytes", "service", "http", aggregate.OperationSum, 30, 2, []float64{10.0, 20.0}),
+				createAgg("bandwidth_sum", "bytes", "service", "tcp", aggregate.OperationSum, 3, 2, []float64{1.0, 2.0}),
+				createAgg("bandwidth_max", "bytes", "service", "http", aggregate.OperationMax, 20, 2, []float64{10.0, 20.0}),
+				createAgg("bandwidth_max", "bytes", "service", "tcp", aggregate.OperationMax, 2, 2, []float64{1.0, 2.0}),
 			},
 		},
 		{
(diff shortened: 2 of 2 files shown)


pkg/pipeline/extract/extract_aggregate_test.go
```diff
@@ -0,0 +1,172 @@
...
+			"aggregate":       "http",
+			"service":         "http",
+			"operation":       aggregate.Operation(aggregate.OperationSum),
+			"value":           "30.000000",
+			"bandwidth_value": "30.000000",
+			"recentRawValues": []float64{10.0, 20.0},
+			"count":           "2",
+		},
+		{
+			"name":            "bandwidth",
```
eranra, line 75: @ronensc what is the difference from the previous test? (do we need this test ?)

pkg/pipeline/extract/extract_aggregate_test.go
```diff
@@ -0,0 +1,172 @@
...
+			"aggregate":             "http",
+			"service":               "http",
+			"operation":             aggregate.Operation(aggregate.OperationCount),
+			"value":                 "2.000000",
+			"bandwidth_count_value": "2.000000",
+			"recentRawValues":       []float64{1.0, 1.0},
+			"count":                 "2",
+		},
+		{
+			"name":                  "bandwidth_count",
```
eranra, line 99: The same question seems that these tests again count. ... 

BTW; maybe add also avg test?

pkg/pipeline/extract/extract_aggregate_test.go
```diff
@@ -0,0 +1,172 @@
...
+			"value":                 "4.000000",
+			"bandwidth_count_value": "4.000000",
+			"recentRawValues":       []float64{1.0, 1.0},
+			"count":                 "4",
+		},
+	}
+	actualAggs2 := extractAggregate.Extract(input2)
+	require.ElementsMatch(t, expectedAggs2, actualAggs2)
+
+}
```
eranra, line 172: @ronensc general remark, maybe put the use cases in some for loop and just run them all of them against `extractAggregate` --- this is very common for tests --- look here for an example https://github.com/netobserv/flowlogs-pipeline/commit/b24be79c22c1b982f4dbd25ea3e2e777f22ae0fc#diff-168486be9e126a8de1f0588534a9bf2fd49261555cbdc3759a5be2b1d2825cd3R26

pkg/pipeline/extract/extract_aggregate_test.go
```diff
@@ -0,0 +1,172 @@
+/*
+ * Copyright (C) 2021 IBM, Inc.
```
eranra, line 2: we are in 2022 ... 2021 is history :-)

eranra (approved):
Looks good to me, some comments but this is good ... waiting for @KalmanMeth to review :-)

Makefile
```diff
@@ -81,7 +81,7 @@ clean: ## Clean
 # note: to review coverage execute: go tool cover -html=/tmp/coverage.out
 .PHONY: test
 test: validate_go ## Test
-	go test -p 1 -race -covermode=atomic -coverprofile=/tmp/coverage.out ./...
+	go test -p 1 -race -coverpkg=./... -covermode=atomic -coverprofile=/tmp/coverage.out ./...
```
ronensc, line 84: @eranra please note that I added the `-coverpkg` flag.
I added it because I noticed that the code coverage didn't increase for `aggregate.go` although this PR tests execution paths that were not tested by the other tests. The reason is that the test in this PR is in a different go package from `aggregate.go`. Adding the flag solves this. The following repo helped me understand this:
https://github.com/yudai/go_cover

pkg/pipeline/extract/extract_aggregate_test.go
ronensc: @eranra The test tests `extract_aggregate` as a whole. It can be thought of as an integration test between `extract_aggregate.go` and `aggregate.go` and `aggregates.go`.
The test sends flows in 2 batches and verifies the extractor's output after each batch. The output of the 2nd batch depends on the 1st batch. I'll check if the referred pattern is suitable here as well.

pkg/pipeline/extract/extract_aggregate_test.go
```diff
@@ -0,0 +1,172 @@
...
+	require.NoError(t, err)
+	config.Opt.PipeLine.Extract.Aggregates = string(jsonBytes)
+
+	extractAggregate, err := NewExtractAggregate()
+	require.NoError(t, err)
+
+	input1 := []config.GenericMap{
+		{"service": "http", "bytes": 10.0},
+		{"service": "http", "bytes": 20.0},
+		{"service": "tcp", "bytes": 1.0},
```
KalmanMeth, line 58: Why are "bytes" floats and not integers?

pkg/pipeline/extract/extract_aggregate_test.go
```diff
@@ -0,0 +1,172 @@
...
+			"name":            "bandwidth",
+			"record_key":      "bytes",
+			"by":              "service",
+			"aggregate":       "http",
+			"service":         "http",
+			"operation":       aggregate.Operation(aggregate.OperationSum),
+			"value":           "30.000000",
+			"bandwidth_value": "30.000000",
+			"recentRawValues": []float64{10.0, 20.0},
+			"count":           "2",
```
KalmanMeth, line 72: We now handle integers and not just strings, so "count" should be an integer. Same for "value" and "bandwidth_value".

KalmanMeth (approved):

pkg/pipeline/extract/extract_aggregate_test.go
```diff
@@ -0,0 +1,152 @@
...
+	jsoniter "github.com/json-iterator/go"
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/extract/aggregate"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
+	"testing"
+)
+
+func createAgg(name, recordKey, by, agg, op string, value float64, count int, rrv []float64) config.GenericMap {
+	valueString := fmt.Sprintf("%f", value)
```
KalmanMeth, line 31: Why do we need these to be strings?

pkg/pipeline/extract/extract_aggregate_test.go
ronensc: I used strings because that's the format returned by: `GetMetrics()`
I admit that I don't understand why it's string...

https://github.com/netobserv/flowlogs-pipeline/blob/59852cfbbf42d7b532b39267d326e5e00ccb243b/pkg/pipeline/extract/aggregate/aggregate.go#L178-L189

pkg/pipeline/extract/extract_aggregate_test.go
ronensc: Same reason as https://github.com/netobserv/flowlogs-pipeline/pull/106#discussion_r815715204

pkg/pipeline/extract/extract_aggregate_test.go
ronensc: I'll fix this

pkg/pipeline/extract/extract_aggregate_test.go
ronensc: Added

eranra:
@ronensc it looks very-very nice now :-) thanks for putting effort to create those tests :-)

PR #116, opened 2022-03-02
Enable goimports

ronensc:
This PR enables `goimports` in `make lint` and runs it on all the go files in the repository.


KalmanMeth:
@ronensc How will this affect what goland does to the order of the import files? Is there some goland configuration that needs to be adjusted to work with this?

eranra (approved):

ronensc:
@KalmanMeth Yes, Eran found the configuration to change in Goland:
image

PR #118, opened 2022-03-03
kafka end-to-end test

KalmanMeth:
In order to run the kafka end-to-end test, you need to have a kafka server running on localhost.
This can be done by entering the directory `pkg/test` and running `docker-compose up -d` to bring up docker containers to run zookeeper and kafka. It takes time for these to be fully functional, so the test cannot be run immediately. If kafka is not available, then the kafka end-to-end test exits gracefully and does not fail the `make test`.


eranra (approved):
Can you move to a new end2end folder and put in sub-folder

pkg/pipeline/kafka_end_to_end_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// run the pipeline in a separate go-routine
+	go func() {
+		mainPipeline.Run()
+	}()
+
+	// wait for the pipeline to come up
+	time.Sleep(time.Second)
+
+	inputData := [][]byte{
```
eranra, line 94: Any chance that you change this to feed the log lines from one of the files ??? 

Or at least generalize this so that we will be able to use this example set of inputs also for other end2end tests ?

pkg/pipeline/kafka_end_to_end_test.go
```diff
@@ -0,0 +1,141 @@
...
+		[]byte("{\"Bytes\":20803,\"DstAddr\":\"10.130.2.3\",\"DstPort\":36936,\"Packets\":403,\"SrcAddr\":\"10.130.2.13\",\"SrcPort\":3100}"),
+	}
+	expectedOutputData := [][]byte{
+		[]byte("{\"k_bytes\":20801,\"k_dstAddr\":\"10.130.2.1\",\"k_dstPort\":36936,\"k_packets\":401,\"k_srcAddr\":\"10.130.2.13\",\"k_srcPort\":3100}"),
+		[]byte("{\"k_bytes\":20802,\"k_dstAddr\":\"10.130.2.2\",\"k_dstPort\":36936,\"k_packets\":402,\"k_srcAddr\":\"10.130.2.13\",\"k_srcPort\":3100}"),
+		[]byte("{\"k_bytes\":20803,\"k_dstAddr\":\"10.130.2.3\",\"k_dstPort\":36936,\"k_packets\":403,\"k_srcAddr\":\"10.130.2.13\",\"k_srcPort\":3100}"),
+	}
+
+	// prepare a kafka producer on topic_in
+	kafkaProducer := kafkago.Writer{
```
eranra, line 106: Why do we need golang code for this ... can we produce using some docker container ??? (adding this to the docker-compose area instead of writing golang code ??)

pkg/pipeline/kafka_end_to_end_test.go
KalmanMeth: done

PR #120, opened 2022-03-03
Integrate components from Goflow-Kube

mariomac:
It integrates the following components from the old [Goflow-Kube repository](https://github.com/netobserv/goflow2-kube-enricher):

* Kube metadata enricher and Informers. There is already a Kube Enricher with its own informers but the output transformation was not the same as the Goflow-Kube payload. In later PRs we can work on merging both and/or adapting the dependant services to the new format.
* End-to-End pipeline test directly taken from Goflow-Kube to verify the compatiblity of the output messages.
* Some changes done to the API to improve testability of the code.


pkg/api/ingest_collector.go
```diff
@@ -18,6 +18,7 @@
 package api
 
 type IngestCollector struct {
-	HostName string `yaml:"hostName" doc:"the hostname to listen on"`
-	Port     int    `yaml:"port" doc:"the port number to listen on"`
+	HostName    string `yaml:"hostName" doc:"the hostname to listen on"`
+	Port        int    `yaml:"port" doc:"the port number to listen on"`
+	BatchMaxLen int    `yaml:"batchMaxLen" doc:"the number of accumulated flows before being forwarded for processing"`
```
KalmanMeth, line 23: @eranra @mariomac Do we also want to be able to specify a timeout value after which to send whatever we have?

pkg/config/config.go
```diff
@@ -80,9 +80,15 @@ type Decode struct {
...
-	Network api.TransformNetwork
+	Type       string
+	Generic    api.TransformGeneric
+	Network    api.TransformNetwork
+	KubeEnrich KubeEnrich `json:"k8s_enrich"`
+}
+
+type KubeEnrich struct {
+	KubeConfigPath string            `json:"kubeConfigPath"`
+	IPFields       map[string]string `json:"ipFields"`
```
KalmanMeth, line 91: Should this perhaps go into the api directory and get included in the api.md?

pkg/pipeline/ingest/ingest_collector.go
```diff
@@ -141,45 +141,60 @@ func (ingestC *ingestCollector) initCollectorListener(ctx context.Context) {
...
+	flushRecords := time.NewTicker(ingestC.batchFlushTime)
+	defer flushRecords.Stop()
 	for {
 		select {
 		case <-ingestC.exitChan:
 			log.Debugf("exiting ingestCollector because of signal")
 			return
 		case record := <-ingestC.in:
+			// TODO: for efficiency, consider forwarding directly as map,
+			// as this is reverted back from string to map in later pipeline stages
```
KalmanMeth, line 154: and skip the decode stage .....

pkg/api/api.go
```diff
@@ -33,4 +33,5 @@ type API struct {
 	TransformNetwork TransformNetwork    `yaml:"network" doc:"## Transform Network API\nFollowing is the supported API format for network transformations:\n"`
 	WriteLoki        WriteLoki           `yaml:"loki" doc:"## Write Loki API\nFollowing is the supported API format for writing to loki:\n"`
 	ExtractAggregate AggregateDefinition `yaml:"aggregates" doc:"## Aggregate metrics API\nFollowing is the supported API format for specifying metrics aggregations:\n"`
+	KubeEnrich       KubeEnrich          `yaml:"k8s_enrich" doc:"## Kubernetes metadata enrichment rules\nFollowing is the supported API format for specifying some network flow enrichment rules from Kubernetes metadata:\n"`
```
jotak, line 36: Hey, we may have not talked enough, but looks like we're not on the same page here. The purpose of the previous work done by Eran during the past weeks was already to close the gap with goflow-kube's kube enrichment (keeping the current enrichment work done in FL2M) and, to my knowledge, we already reached a point where we are satisfied. The purpose of PRs [here](https://github.com/netobserv/network-observability-operator/pull/60) and [there](https://github.com/netobserv/network-observability-console-plugin/pull/78) are adapting with the slight model differences that we can live with (there's no functional regression as far as I know). So I don't think we need an alternative way to do enrichment. When these two PRs are merged, we should be able to complete the merge with features parity.
Of course if we've missed something / if the gap isn't closed as I thought, it's something to bring up

jotak: PS: or maybe I misunderstand this PR? but it seems to me that you're reintroducing the goflow-kube's enrichment in place of the one already present here?

mariomac: Ok. I initially wanted to provide a 100% compatible alternative with the previous goflow-kube-enricher, but if you think it's not needed I can just remove this Enricher and use the previous network enricher.

jotak: yes, that would be great. I guess there's other changes that you're doing here that are relevant regardless the kube enrichment, like `BatchMaxLen`

eranra: @mariomac in FL2P the kube enrichment is done in network transform/ AddKubernetes operation --- as @jotak stated it should provide now very similar functionally to the original kube enrichment --- but there are two points to consider:

1. if you think that we need to move this out of `network transform` and be a stand-alone stage that handles just kubernethes ... and if you think on renaming from transforming into `enrich`.

2. Also, if there are any gaps that are missed ... we should add them in

pkg/api/ingest_collector.go
eranra: @mariomac I like the one-by-one + channels approach ... getting more performance by "multi-thread/core" is what we need.

pkg/config/config.go
```diff
@@ -80,9 +80,10 @@ type Decode struct {
 }
 
 type Transform struct {
-	Type    string
-	Generic api.TransformGeneric
-	Network api.TransformNetwork
+	Type       string
+	Generic    api.TransformGeneric
+	Network    api.TransformNetwork
+	KubeEnrich api.KubeEnrich `json:"k8s_enrich"`
```
eranra, line 86: blend the functionality into `TransformNetwork`

pkg/pipeline/pipeline.go
```diff
@@ -55,7 +55,11 @@ func NewPipeline() (*Pipeline, error) {
 	configParams := config.Parameters
 	log.Debugf("configParams = %v ", configParams)
 
-	return newBuilder(configParams, stages).build()
+	build := newBuilder(configParams, stages)
+	if err := build.readStages(); err != nil {
+		return nil, err
```
eranra, line 60: 👍

pkg/pipeline/pipeline_netobs_test.go
```diff
@@ -0,0 +1,201 @@
+package pipeline
```
eranra, line 1: @mariomac if this is "specific" for some "use-case" or "scenario" ... maybe the right thing is to open a folder called scenarios and put the test file inside this folder. We might have more "use-cases" tests in the future. Also, can you think of some other name for "netobs" it is not informative? Maybe something like a redhat-ocp scenario or something like that. We need to think on this repo as being an upstream repo.

pkg/pipeline/pipeline_netobs_test.go
```diff
@@ -0,0 +1,201 @@
...
+		Port:        ipfixPort,
+		BatchMaxLen: 1,
+	})
+	require.NoError(t, err)
+
+	// STAGE 2: decoding from flow's JSON data to GenericMAP
+	decoder, err := decode.NewDecodeJson()
+	require.NoError(t, err)
+
+	// STAGE 3: Kube Enricher getting its data from a mocked informer
```
eranra, line 63: We do not have the term "enrich" as part of FLP ... I think that we might replace transform with `enrich` (if you think that this is correct for kube) but we need to do that across all the project ... 

There are multiple placed in this test file that need to be updated

pkg/pipeline/pipeline_netobs_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"fmt"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/api"
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/decode"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/ingest"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/transform/netobserv"
```
eranra, line 13: need to remove this and move to network transform

pkg/pipeline/transform/kubernetes/kubernetes.go
```diff
@@ -202,36 +202,13 @@ func (k *KubeData) NewReplicaSetInformer(informerFactory informers.SharedInforme
...
-			config, err = clientcmd.BuildConfigFromFlags("", homeDir+"/.kube/config")
-			if err != nil {
-				// creates the in-cluster config
-				config, err = rest.InClusterConfig()
-				if err != nil {
-					return fmt.Errorf("can't access kubenetes. Tried using config from: config parameter, %s env, homedir and InClusterConfig", kubeConfigEnvVariable)
-				}
-			}
-		}
+	config, err := LoadConfig(kubeConfigPath)
```
eranra, line 209: @mariomac double-check ... same functionality or different ???

pkg/pipeline/transform/kubernetes/kubernetes.go
```diff
@@ -247,6 +224,33 @@ func (k *KubeData) InitFromConfig(kubeConfigPath string) error {
 	return nil
 }
 
+func LoadConfig(kubeConfigPath string) (*rest.Config, error) {
```
eranra, line 227: 👍

pkg/test/utils.go
```diff
@@ -113,3 +119,51 @@ func GetExtractMockEntry() config.GenericMap {
 	}
 	return entry
 }
+
+// FakeLokiHandler is a fake loki HTTP service that decodes the snappy/protobuf messages
+// and forwards them for later assertions
+func FakeLokiHandler(flowsData chan<- map[string]interface{}) http.HandlerFunc {
```
eranra, line 125: Can we move this to a specific file "loki mock or something like that"

pkg/test/ipfix.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+const (
+	templateID          = 256
+	observationDomainID = 1
+)
+
+// Values taken from https://www.iana.org/assignments/ipfix/ipfix.xhtml
+var (
+	sourceIPv4Address = entities.NewInfoElement("sourceIPv4Address", 8, entities.Ipv4Address, 0, 4)
```
eranra, line 18: This "feels" like end-to-end test ... and very specific to IPFIX ... do we need this ... what component is being tested here ???

pkg/pipeline/ingest/ingest_collector_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// GIVEN an IPFIX collector Ingester
+	go ic.Ingest(forwarded)
+
+	client, err := test.NewIPFIXClient(collectorPort)
+	require.NoError(t, err)
+
+	// The IPFIX client might send information before the Ingester is actually listening,
+	// so we might need to repeat the submission
+	test.Eventually(t, timeout, func(t require.TestingT) {
```
eranra, line 35: I really do not like "inconsistent" tests ... can we make this test more robust and drop the Eventually concept ... the entire Eventually concept feels like hiding from creating a consistent and deterministic test.

pkg/pipeline/transform/kubernetes/kubernetes.go
```diff
@@ -247,6 +224,33 @@ func (k *KubeData) InitFromConfig(kubeConfigPath string) error {
...
+	if kubeConfigPath == "" {
+		kubeConfigPath = os.Getenv(kubeConfigEnvVariable)
+	}
+	// otherwise, load it from the $HOME/.kube/config file
+	if kubeConfigPath == "" {
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			return nil, fmt.Errorf("can't get user home dir: %w", err)
+		}
+		kubeConfigPath = homeDir + "/.kube/config"
```
ronensc, line 238: Is `path.Join()` more suitable here?
https://pkg.go.dev/path#Join

pkg/pipeline/pipeline_builder.go
```diff
@@ -216,7 +215,7 @@ func getIngester(params config.StageParam) (ingest.Ingester, error) {
 	case "file", "file_loop", "file_chunks":
 		ingester, err = ingest.NewIngestFile(params)
 	case "collector":
-		ingester, err = ingest.NewIngestCollector(params)
+		ingester, err = ingest.NewIngestCollector(params.Ingest.Collector)
```
KalmanMeth, line 218: If we make this `New` ingest procedure specific, we should consider making all the other `New` procedures specific. If not,then this one should also be generic (`params` instead of `params.Ingest.Collector` for consistency.

pkg/pipeline/ingest/ingest_collector_test.go
mariomac: In this case, the "Eventually" is not hiding an inconsistent test but addressing a synchronization issue. Goflow2 is initializing in an internal, background goroutine, and there is no way that we can know when it has started to accept flows (as UDP connections are uni-directional).

There are two ways to address this:
1. Adding a Timeout of e.g. 1-2 seconds, but that would cause that this test is slow.
2. Keep sending flows until we observe that they are being forwarded.

In the next commit, I have rewritten the test to be more explicit that we follow case 2, without giving the impression of flakyness.

pkg/pipeline/pipeline_netobs_test.go
mariomac: This was located here because tests we often rely on private APIs. But you are right when you say that this is only a scenario-addressed test.

I think that, for the moment, I will remove this test, as it relies on other components that are already addressed.

In a future, I think it would be nice to enable integration tests that do not rely on any private API:

1. Enabling tests that are run via docker-compose, as it is actually being done.
2. Changing the components' constructors to not be constructed directly from a configuration file but allowing dependency injection.

pkg/test/ipfix.go
mariomac: This is used to test the Ingest Collector (`ingest_collector_test.go`), as it remained untested.

eranra: @mariomac ok ... this looks good to me  ... the only comment small comment I have is that maybe it makes sense to put this file, and maybe also the Loki mock file under some "internal" folder underneath "test" and not directly in test.

eranra (approved):
@mariomac thanks ... this looks very good to me

Issue #113, opened 2022-03-02
Update README to align with the new pipeline structure 


eranra:
@KalmanMeth Various sections of the README will not be accurate anymore. In particular the examples in  https://github.com/netobserv/flowlogs-pipeline#syntax-of-portions-of-the-configuration-file but also https://github.com/netobserv/flowlogs-pipeline#configuration and various other places. This is mainly a gap post-merge of  https://github.com/netobserv/flowlogs-pipeline/pull/83 but the changes that @mariomac introduces in https://github.com/netobserv/flowlogs-pipeline/pull/105 might also affect. We need to revisit the text and technical content to make sure it is accurate.

PR #121, opened 2022-03-03
updated README to include new config structure

KalmanMeth:
See issue #113

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ It is expected that the **ingest** module will receive flows every so often, and
 
 # Configuration
 
-It is possible to configure flowlogs-pipeline using command-line-parameters, configuration file, environment variables, or any combination of those options.
+It is possible to configure flowlogs-pipeline using command-line-parameters, configuration file, or any combination of those options.
 
 
 For example:
@@ -212,14 +212,6 @@ OR place the configuration information in any other file `<configFile>` and exec
    ```./flowlogs-pipeline --log-level info --config <configFile>```
 Options included in the command line override the options specified in the config file.
 
-3. using environment variables:
-- set environment variables
-```bash
-export FLOWLOGS-PIPELINE_LOG_LEVEL=info
-```
-- execute `./flowlogs-pipeline`
-
-Options specified in environment variables override the options specified in the config file.
 
 Supported options are provided by running:
 
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ It is possible to configure flowlogs-pipeline using command-line-parameters, con
 
 For example:
 1. Using configuration file:
-- create under $HOME/.flowlogs-pipeline.yaml the file:
+
 ```yaml
 log-level: info
 pipeline:
@@ -208,9 +208,8 @@ parameters
     write:
       type: stdout
 ```
-- execute `./flowlogs-pipeline`
+- execute
 
-OR place the configuration information in any other file `<configFile>` and execute
 
 `./flowlogs-pipeline --config <configFile>`
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ FLP allows to define mathematical transformations to generate condense metrics t
 
 In addition, FLP decorates the metrics with **context**, allowing visualization layers and analytics frameworks to present **network insights** to SRE’s, cloud operators and network experts.
 
-Along with Prometheus and its ecosystem tools such as Thanos, Cortex etc., FLP provides efficient scalable multi-cloud solution for comprehensive network analytics that rely **solely based on metrics data-source**.
+Along with Prometheus and its ecosystem tools such as Thanos, Cortex etc., FLP provides an efficient scalable multi-cloud solution for comprehensive network analytics that rely **solely on metrics data-source**.
 
 Default metrics are documented here [docs/metrics.md](docs/metrics.md).
 <br>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -103,57 +103,126 @@ FLP is a framework. The main FLP object is the **pipeline**. FLP **pipeline** ca
 
 # Architecture
 
-The pipeline is constructed of a sequence of stages:
+The pipeline is constructed of a sequence of stages. Each stage is classified into one of the following classes:
```
eranra, line 106: Do we use the term class or do we use `type` ... I do not want to confuse the readers

README.md
```diff
@@ -103,57 +103,126 @@ FLP is a framework. The main FLP object is the **pipeline**. FLP **pipeline** ca
...
 - **ingest** - obtain flows from some source, one entry per line
 - **decode** - parse input lines into a known format, e.g., dictionary (map) of AWS or goflow data
 - **transform** - convert entries into a standard format; can include multiple transform stages
 - **write** - provide the means to write the data to some target, e.g. loki, standard output, etc
 - **extract** - derive a set of metrics from the imported flows
 - **encode** - make the data available in appropriate format (e.g. prometheus)
 
-The **encode** and **write** stages may be combined in some cases (as in prometheus), in which case **write** is set to **none**
+The first stage in a pipeline must be an **ingest** stage, followed by a **decode** stage.
```
eranra, line 114: Can we have more than one decode ... it is not clear from the text?

README.md
```diff
@@ -103,57 +103,126 @@ FLP is a framework. The main FLP object is the **pipeline**. FLP **pipeline** ca
...
+Each stage (other than an **ingest** stage) specifies the stage it follows.
+Multiple stages may follow from a particular stage, thus allowing the same data to be fed into multiple parallel pipelines.
```
eranra, line 116: consider rephrase of the term "to be fed" - (should we use produce and consume terms to describe the process ... ???_)

README.md
```diff
@@ -103,57 +103,126 @@ FLP is a framework. The main FLP object is the **pipeline**. FLP **pipeline** ca
...
+Multiple **transform** stages may be performed and the results may be output to different targets.
```
eranra, line 117: Maybe also explain that there are two configuration sections, the first describes the high-level flow of information between the stages and builds the graph of consumption and production of information between stages 

(I am using the term information and not data ... I am not sure that even this is the correct term ...)

README.md
```diff
@@ -103,57 +103,126 @@ FLP is a framework. The main FLP object is the **pipeline**. FLP **pipeline** ca
...
 
 # Configuration
 
 It is possible to configure flowlogs-pipeline using command-line-parameters, configuration file, environment variables, or any combination of those options.
 
+
 For example:
 1. Using command line parameters:
-```./flowlogs-pipeline --pipeline.ingest.type file --pipeline.ingest.file.filename hack/examples/ocp-ipfix-flowlogs.json```
+```./flowlogs-pipeline --log-level info ```
```
eranra, line 185: Will that work ??? don't we need to define the input somehow ... this might be a longs JSON now, but still we can do that from the command line

README.md
```diff
@@ -103,57 +103,126 @@ FLP is a framework. The main FLP object is the **pipeline**. FLP **pipeline** ca
...
+    decode:
+      type: json
 ```
 - execute `./flowlogs-pipeline`
 3. using environment variables:
 - set environment variables
 ```bash
-export FLOWLOGS-PIPELINE_PIPELINE_INGEST_TYPE=file
-export FLOWLOGS-PIPELINE_PIPELINE_INGEST_FILE_FILENAME=hack/examples/ocp-ipfix-flowlogs.json
+export FLOWLOGS-PIPELINE_LOG_LEVEL=info
```
eranra, line 208: same as (1) this will not sore without some environment variables specifying the file

eranra (approved):

README.md
KalmanMeth: Added --config <configFile>. The command-line parameters over-ride the config file parameters.

README.md
KalmanMeth: Is the updated text now better?

eranra: I think so :-)

README.md
```diff
@@ -103,57 +103,138 @@ FLP is a framework. The main FLP object is the **pipeline**. FLP **pipeline** ca
...
+      type: json
 ```
 - execute `./flowlogs-pipeline`
+
+OR place the configuration information in any other file `<configFile>` and execute
+`./flowlogs-pipeline --config <configFile>`
+
+2. Using command line parameters:
+   ```./flowlogs-pipeline --log-level info --config <configFile>```
+Options included in the command line override the options specified in the config file.
```
eranra, line 213: @KalmanMeth I am still not confident with that area of the README ... let's talk about this tomorrow. If the command line and environment variable options are broken we can add something temporal and add an issue to fix both ... but I do not want to have something that does not represent reality./

README.md
KalmanMeth: The environment variables do not work at present. We are not allowed a `-` (dash) in an environment variable. I will remove this from the current documentation.

README.md
KalmanMeth: The environment variables do not work currently. The command line does work as shown in the example.

PR #128, opened 2022-03-07
Fix a bug with add_if and scientific notation

ronensc:
I noticed that when the expression evaluated in "add_if" contains a scientific notation value like `1.2345e-67`, there's an error like:
```
Cannot transition token types from NUMERIC [1.2345] to VARIABLE [e]
```
The solution I found is to pass the value as a parameter to `Evaluate()`


eranra (approved):
:-)

PR #129, opened 2022-03-07
Create branch-based and commit-based images

jotak:
- an image tagged "main" (tracking branch "main")
- an image, short-lived, tagged after commit SHA

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,56 @@
+name: release to quay.io
+on:
+  push:
+    tags: [v*]
+
+env:
+  REGISTRY_USER: netobserv+github_ci
+  REGISTRY_PASSWORD: ${{ secrets.QUAY_SECRET }}
+  REGISTRY: quay.io/netobserv
+  IMAGE: flowlogs-pipeline
+
+jobs:
+  push-image:
+    name: push image
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        go: ['1.17']
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: validate tag
+        id: validate_tag
+        run: |
+          tag=`git describe --exact-match --tags 2> /dev/null`
+          if [[ $tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-rc[0-9]+)?$ ]]; then
+              echo "$tag is a valid release tag"
+              set -e
+              echo "::set-output name=tag::$tag"
+          else
+              echo "$tag is NOT a valid release tag"
+              exit 1
+          fi
+      - name: install make
+        run: sudo apt-get install make
+      - name: set up go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+      - name: build images
+        run: DOCKER_TAG=${{ steps.validate_tag.outputs.tag }} make build-images
+      - name: podman login to quay.io
+        uses: redhat-actions/podman-login@v1
+        with:
+          username: ${{ env.REGISTRY_USER }}
+          password: ${{ env.REGISTRY_PASSWORD }}
+          registry: quay.io
+      - name: push to quay.io
+        id: push-to-quay
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ${{ env.IMAGE }}
+          tags: ${{ steps.validate_tag.outputs.tag }}
+          registry: ${{ env.REGISTRY }}
+      - name: print image url
+        run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
--- a/Makefile
+++ b/Makefile
@@ -116,11 +116,15 @@ run: build ## Run
 build-image:
 	DOCKER_BUILDKIT=1 $(OCI_RUNTIME) build -t $(DOCKER_IMG):$(DOCKER_TAG) -f contrib/docker/Dockerfile .
 
-.PHONY: build-main-and-sha
-build-main-and-sha: build-image
-	echo "FROM $(DOCKER_IMG):$(DOCKER_TAG)" > tmp.Dockerfile && \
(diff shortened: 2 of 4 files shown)


eranra:
@jotak I think that having a "latest" tag  is useful and common so I suggest adding that back
@jotak can you explain more the purpose of this code ... the trigger to this action is ```
on:
  push:
    branches: [ main ]``` --- so we get only main branch push

Makefile
```diff
@@ -116,6 +116,12 @@ run: build ## Run
 build-image:
 	DOCKER_BUILDKIT=1 $(OCI_RUNTIME) build -t $(DOCKER_IMG):$(DOCKER_TAG) -f contrib/docker/Dockerfile .
 
+.PHONY: build-main-and-sha
+build-main-and-sha: build-image
+	echo "FROM $(DOCKER_IMG):$(DOCKER_TAG)" > tmp.Dockerfile && \
```
eranra, line 121: @jotak I am not sure that I get the process here .... don't we need to execute `build-image` so we will have the `FROM $(DOCKER_IMG):$(DOCKER_TAG)` available as a prereq for `build-main-and-sha` ??? without running that why should we have that image and we can use that as "FROM" ?

eranra, line 121: Any option to change `tmp.Dockerfile` to be located in the `tmp` directory?

Makefile
```diff
@@ -116,6 +116,12 @@ run: build ## Run
 build-image:
 	DOCKER_BUILDKIT=1 $(OCI_RUNTIME) build -t $(DOCKER_IMG):$(DOCKER_TAG) -f contrib/docker/Dockerfile .
 
+.PHONY: build-main-and-sha
+build-main-and-sha: build-image
```
eranra, line 120: Should this be really `build-main-and-sha` or we can rename it to `build-image-with-commit-sha`

BTW: in general I do not like the complexity so much ... from my POV we need two images (1) a `latest` tag and (2) an image with a `version` tag . This is what I am used to from other projects ... I agree and it make sense to me that those  version tags can be temporal unless they are releases ... but the code here is a bit too complicated for me.

Makefile
jotak: but that's what it does: `build-main-and-sha: build-image`
Am I missing something?

jotak: sure I can do that

Makefile
jotak: This has proven helpful sometimes to the QE to be able to fallback on a recent image when the main/latest was broken

jotak:
> @jotak I think that having a "latest" tag is useful and common so I suggest adding that back

ok

> @jotak can you explain more the purpose of this code ... the trigger to this action is ` on: push: branches: [ main ]` --- so we get only main branch push

At the moment there's only a `main` branch, but at some point there will be more branches and we should then update the script to include them. This can be needed to backport fixes that target a version developed for a specific OCP version. We must consider the full picture operator+FLP+console plugin for that. There's sometimes breaking changes (e.g. in the pipeline config, like we've seen in the last few days) that makes it necessary to keep track of the version released for a given OCP version as a branch. If we adopt semver, it could be done with branches like `v0.1.x`, `v0.2.x`, and for a given OCP release, the netobserv components are bound to such a version branch.
Since we haven't released FLP yet, we still only have `main`, but we should have more soon.
To get back to image tagging, this build script will have to evolve, but the idea is to have one tag per branch in the future.

Makefile
jotak: the idea is really not complicated, it's only to rebuild the same image with a label `--label quay.expires-after=2w` and a different tag, nothing more. I wish there is a simpler way to implement it, but the simplest I found was to have this temporary dockerfile (if you can think about a simpler way, I'm all ears :) )

jotak:
@eranra I did a few changes, adding `latest` and removing the temporary Dockerfile (replaced with `shortlived.Dockerfile` in contrib dir), hope it makes the Makefile more easy to reason about

jotak:
I've also added a release action, same as in the console plugin repo
I haven't fully tested it, will do when we can try a release candidate

Makefile
eranra: @jotak No, it was me that did not understand ... now I get it + the new version is much much more simple and easy to understand ... :-) ...I like think that this is very simple and good now :-)

eranra: :-)

Makefile
eranra: Got it + now when you added the `[shortlived.Dockerfile](https://github.com/netobserv/flowlogs-pipeline/pull/129/files#diff-2ed92d81a3aa4dd87afe89557f95ec38735e3c175dcad5aabbe79595bbf9ff01)` thing it is much more understood (at least for me) ... so I am good with that

eranra:
> @eranra I did a few changes, adding `latest` and removing the temporary Dockerfile (replaced with `shortlived.Dockerfile` in contrib dir), hope it makes the Makefile more easy to reason about

Yes .. very good change :-)

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+    strategy:
+      matrix:
+        go: ['1.17']
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: validate tag
+        id: validate_tag
+        run: |
+          tag=`git describe --exact-match --tags 2> /dev/null`
```
eranra, line 25: A question and a suggestion ... we have some of that logic in the version building in the Makfile ... would it make sense to expose that with one a Makefile target instead of writing all this logic here ... maybe move to a script under hack ... I just think that having so much logic inside github action might not be good practice?


This is just a suggestion ... not blocking merge or something like that ,.,., this is good now :-0)

eranra (approved):

eranra:
> I've also added a release action, same as in the console plugin repo I haven't fully tested it, will do when we can try a release candidate

👍

.github/workflows/release.yml
jotak: Maybe one downside if we have it in a separate place, it's that the `echo "::set-output name=tag::$tag"` command, which is tied to the rest of the github action, would be less visible from the action, so it could be more confusing or error prone if the script is modified. What do you think?

PR #134, opened 2022-03-08
Descriptive error messages on bad pipeline formation

mariomac:
If a stage is not fully connected, it returns more descriptive messages, indicating which stage is wrongly connected. E.g.:

```
pipeline stage "ingest1": pipeline stage from type "ingest" should send data to at least another stage
pipeline stage "decode1": pipeline stage from type "decode" should receive data from at least another stage
pipeline stage "decode1": pipeline stage from type "decode" should send data to at least another stage
pipeline stage "write1": pipeline stage from type "write" should receive data from at least another stage
```

### DIFF
--- a/pkg/pipeline/pipeline_builder.go
+++ b/pkg/pipeline/pipeline_builder.go
@@ -136,7 +136,7 @@ func (b *builder) build() (*Pipeline, error) {
 			return nil, fmt.Errorf("stage %q of type %q can't send data",
 				connection.Follows, srcEntry.stageType)
 		}
-		log.Debugf("connecting stages: %s --> %s", connection.Follows, connection.Name)
+		log.Infof("connecting stages: %s --> %s", connection.Follows, connection.Name)
 
 		sendingNodes[connection.Follows] = struct{}{}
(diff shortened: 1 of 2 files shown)


pkg/pipeline/pipeline_builder.go
```diff
@@ -126,13 +138,18 @@ func (b *builder) build() (*Pipeline, error) {
 		}
 		log.Debugf("connecting stages: %s --> %s", connection.Follows, connection.Name)
```
eranra, line 139: I think that you can move this to be info

eranra (approved):

pkg/pipeline/pipeline_builder_test.go
```diff
@@ -0,0 +1,97 @@
...
+  name: decode1
+- write:
+    type: none
+  name: write1
+`
+
+func TestConnectionVerification_Pass(t *testing.T) {
+	test.InitConfig(t, baseConfig+`pipeline:
+- { follows: ingest1, name: decode1 }
+- { follows: decode1, name: write1 }
```
ronensc, line 28: COOL! I didn't know that curly braces are valid in YAML

PR #136, opened 2022-03-09
update readme and graphics

eranra:


jotak (approved):
LGTM

README.md
```diff
@@ -5,13 +5,22 @@
 
 # Overview
 
-**Flow-Logs Pipeline** (a.k.a. FLP) is an **observability tool** that consumes raw **network flow-logs** and
-transforms them from their original format (NetFlow or IPFIX) into **prometheus numeric metrics** format.
-FLP allows to define mathematical transformations to generate condense metrics that encapsulate network domain knowledge.
+**Flow-Logs Pipeline** (a.k.a. FLP) is an **observability tool** that consumes raw **network flow-logs** in
+their original format 
+([NetFlow](https://en.wikipedia.org/wiki/NetFlow) or [IPFIX](https://en.wikipedia.org/wiki/IP_Flow_Information_Export))
```
stleerh, line 10: Specify NetFlow version like NetFlow v9.

stleerh (commented):
Comments:
1. Kafka isn't mentioned anywhere in the doc but is in the diagram.
2. There needs to be some installation instructions for Loki, Grafana and possibly Kafka.  It should answer questions like: Can these components be in a different namespace?  Can they be external?

Future:
- It should indicate how much storage is needed by Prometheus and Loki.
- It should specify how much NetFlow it can handle.

eranra:
> Comments:
> 
> 1. Kafka isn't mentioned anywhere in the doc but is in the diagram.

Good catch - added issue https://github.com/netobserv/flowlogs-pipeline/issues/139 

> 2. There needs to be some installation instructions for Loki, Grafana and possibly Kafka.  It should answer questions like: Can these components be in a different namespace?  Can they be external?

I updated the deployment text a little. @stleerh in general one-click `make ocp-deploy` will do all of that for you :-) 

> 
> Future:
> 
> * It should indicate how much storage is needed by Prometheus and Loki.

Good question ... added issue https://github.com/netobserv/flowlogs-pipeline/issues/140  ---> maybe getting an intern on that area and/or get info from perf. test team in RH on that area.

> * It should specify how much NetFlow it can handle.

I agree ... from my POV this is the same question i.e. for how much input we provide how much output taking how many resources ... but Yes, we need some performance and benchmarking information here.

README.md
eranra: done

PR #137, opened 2022-03-09
fix local deploy

eranra:


Makefile
```diff
@@ -203,7 +203,7 @@ undeploy-netflow-simulator: ## Undeploy netflow simulator
 
 .PHONY: create-kind-cluster
 create-kind-cluster: $(KIND) ## Create cluster
-	$(KIND) create cluster
+	$(KIND) create cluster --config contrib/kubernetes/kind/kind.config.yaml
```
KalmanMeth, line 206: Where is kubernetes/kind/kind.config.yaml created? Should this perhaps be  kubernetes/flowlogs-pipeline.conf.yaml?

KalmanMeth (approved):

PR #146, opened 2022-03-14
Refactor extractor fields

ronensc:
This PR:
- Adds fields `recent_op_value` and `recent_count` to the extractor output to be used in prometheus counters
- Adds an integration test between the extractor and prom encoder

It will follow up with PRs to add RawValuesOperation and a filter field to prom encoder config

### DIFF
--- a/pkg/pipeline/aggregate_prom_test.go
+++ b/pkg/pipeline/aggregate_prom_test.go
@@ -17,10 +17,8 @@
 package pipeline
 
 import (
-	"fmt"
 	"testing"
 
-	"github.com/netobserv/flowlogs-pipeline/pkg/api"
 	"github.com/netobserv/flowlogs-pipeline/pkg/config"
 	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/encode"
 	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/extract"
@@ -28,24 +26,6 @@ import (
 	"github.com/netobserv/flowlogs-pipeline/pkg/test"
 	"github.com/stretchr/testify/require"
 )
-
-func createAgg(name, recordKey, by, agg, op string, value float64, count int, rrv []float64, recentOpValue float64, recentCount int) config.GenericMap {
-	valueString := fmt.Sprintf("%f", value)
-	return config.GenericMap{
-		"name":                                  name,
-		"record_key":                            recordKey,
-		"by":                                    by,
-		"aggregate":                             agg,
-		by:                                      agg,
-		"operation":                             api.AggregateOperation(op),
-		"total_value":                           valueString,
-		fmt.Sprintf("%v_total_value", name):     valueString,
-		"recentRawValues":                       rrv,
-		"total_count":                           fmt.Sprintf("%v", count),
-		fmt.Sprintf("%v_recent_op_value", name): recentOpValue,
-		fmt.Sprintf("%v_recent_count", name):    recentCount,
-	}
-}
 
 func createEncodeOutput(name string, labels map[string]string, value float64) config.GenericMap {
 	gm := config.GenericMap{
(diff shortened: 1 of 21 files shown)


pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -48,6 +48,8 @@ type Aggregate struct {
 type GroupState struct {
 	normalizedValues NormalizedValues
 	RecentRawValues  []float64
+	recentOpValue    float64
+	recentCount      int
```
eranra, line 52: s/recentCount/rescentOpCount

Logic: To align with `recentOpValue`

pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -167,20 +185,29 @@ func (aggregate Aggregate) Evaluate(entries []config.GenericMap) error {
...
+			aggregate.Definition.Name + "_total_value":     fmt.Sprintf("%f", group.value),
+			strings.Join(aggregate.Definition.By, "_"):     string(group.normalizedValues),
 		})
-		// Once reported, we reset the raw values accumulation
+		// Once reported, we reset the recentXXX fields
 		group.RecentRawValues = make([]float64, 0)
+		group.recentCount = 0
+		initVal, err := initValue(string(aggregate.Definition.Operation))
+		if err != nil {
+			log.Errorf("Error: %v", err)
```
eranra, line 208: Maybe change to Panic ... this should never happen ... on the other hand, no one will see this Error ... so I suggest to panic here and also in on the other place that `initValue` is used

pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -167,20 +185,29 @@ func (aggregate Aggregate) Evaluate(entries []config.GenericMap) error {
...
-			"recentRawValues":                    group.RecentRawValues,
-			"count":                              fmt.Sprintf("%d", group.count),
-			aggregate.Definition.Name + "_value": fmt.Sprintf("%f", group.value),
-			strings.Join(aggregate.Definition.By, "_"): string(group.normalizedValues),
+			"name":            aggregate.Definition.Name,
+			"operation":       aggregate.Definition.Operation,
+			"record_key":      aggregate.Definition.RecordKey,
+			"by":              strings.Join(aggregate.Definition.By, ","),
+			"aggregate":       string(group.normalizedValues),
+			"total_value":     fmt.Sprintf("%f", group.value),
```
eranra, line 195: s/group.value/group.totalValue ... this will help to make sure that when reading the code you understand what value we are talking about.

pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -167,20 +185,29 @@ func (aggregate Aggregate) Evaluate(entries []config.GenericMap) error {
...
+			"recentRawValues": group.RecentRawValues,
+			"total_count":     fmt.Sprintf("%d", group.count),
```
eranra, line 197: s/group.count/group.totalCount ...  same reason

pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -117,6 +129,7 @@ func (aggregate Aggregate) UpdateByEntry(entry config.GenericMap, normalizedValu
 
 	if operation == OperationCount {
 		groupState.value = float64(groupState.count + 1)
+		groupState.recentOpValue = float64(groupState.recentCount + 1)
 		groupState.RecentRawValues = append(groupState.RecentRawValues, 1)
```
eranra, line 133: s/RecentRawValues/recentRawValues .... no reason to expose + make sure we are consistnet

eranra (commented):
Need to update the documentation (at least the README.md) to reflect the changes.

pkg/pipeline/aggregate_prom_test.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/netobserv/flowlogs-pipeline/pkg/api"
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/encode"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/extract"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/extract/aggregate"
+	"github.com/netobserv/flowlogs-pipeline/pkg/test"
+	"github.com/stretchr/testify/require"
+)
+
+func createAgg(name, recordKey, by, agg, op string, value float64, count int, rrv []float64, recentOpValue float64, recentCount int) config.GenericMap {
```
eranra, line 32: Consider moving to test/utils ... we have other mock values there ... so it make sense to move that there so this can be used also by other test files

eranra: s/createAgg/createMockAgg

pkg/pipeline/aggregate_prom_test.go
```diff
@@ -0,0 +1,188 @@
...
+		"value":  value,
+	}
+	return gm
+}
+
+// Test_Extract_Encode tests the integration between extract_aggregate and encode_prom.
+// The test sends flows in 2 batches. Each batch is passed through the extractor and the encoder.
+// The output of each stage is verified.
+// The output of the 2nd batch depends on the 1st batch.
+func Test_Extract_Encode(t *testing.T) {
```
eranra, line 63: Nice test :-)

eranra (commented):
Need to update the https://github.com/netobserv/flowlogs-pipeline/blob/main/docs/confGenerator.md do to explain when to use `recent_op_value` and when to use `recent_count`

eranra (approved):
:-)

pkg/pipeline/extract/aggregate/aggregate.go
ronensc: I don't agree. In a later PR I'll add the `raw_values` operation to support histograms. 
That PR will redefine when the field `RecentRawValues` is set. Currently, it is always set.
That PR will cause it to be set only when the new `raw_values` operation is used.
Moreover, when `raw_values` is used, `recentOpValue` has no meaning.

To conclude, we will have 3 related fields:
1. recentRawValues
2. recentOpValue
3. recentCount

The 1st will be set only for `raw_values` operation,
The 2nd will be set for the other operations,
And the 3rd will be set for all operations.

I think that the common prefix should be `recent` rather than `recentOp`

pkg/pipeline/extract/aggregate/aggregate.go
ronensc: OK. So in that case I can change the signature of `initValue()` to return only float64 (no error) and panic inside the function when an known operation is encountered.

ronensc:
> Need to update the https://github.com/netobserv/flowlogs-pipeline/blob/main/docs/confGenerator.md do to explain when to use `recent_op_value` and when to use `recent_count`

I'll do that as part of #147 since it already edits the file

pkg/pipeline/extract/aggregate/aggregate.go
eranra: OK :-) I am convinced

PR #147, opened 2022-03-15
Add filter field to prometheus encoder

ronensc:
Follow-up for #146. Please review that one before this.

### DIFF
--- a/pkg/api/encode_prom.go
+++ b/pkg/api/encode_prom.go
@@ -35,11 +35,17 @@ func PromEncodeOperationName(operation string) string {
 }
 
 type PromMetricsItem struct {
-	Name     string    `yaml:"name" doc:"the metric name"`
-	Type     string    `yaml:"type" enum:"PromEncodeOperationEnum" doc:"one of the following:"`
-	ValueKey string    `yaml:"valuekey" doc:"entry key from which to resolve metric value"`
-	Labels   []string  `yaml:"labels" doc:"labels to be associated with the metric"`
-	Buckets  []float64 `yaml:"buckets" doc:"histogram buckets"`
+	Name     string            `yaml:"name" doc:"the metric name"`
+	Type     string            `yaml:"type" enum:"PromEncodeOperationEnum" doc:"one of the following:"`
+	Filter   PromMetricsFilter `yaml:"filter" doc:"the criterion to filter entries by"`
+	ValueKey string            `yaml:"valuekey" doc:"entry key from which to resolve metric value"`
+	Labels   []string          `yaml:"labels" doc:"labels to be associated with the metric"`
+	Buckets  []float64         `yaml:"buckets" doc:"histogram buckets"`
 }
 
 type PromMetricsItems []PromMetricsItem
+
+type PromMetricsFilter struct {
+	Key   string `yaml:"key" doc:"the key to match and filter by"`
+	Value string `yaml:"value" doc:"the value to match and filter by"`
+}
--- a/pkg/confgen/encode.go
+++ b/pkg/confgen/encode.go
@@ -40,17 +40,7 @@ func (cg *ConfGen) parseEncode(encode *map[string]interface{}) (*api.PromEncode,
 		log.Debugf("Unmarshal aggregate.Definitions err: %v ", err)
 		return nil, err
 	}
-
-	// validate that ValueKey is unique
-	for _, existingPromMetric := range cg.promMetrics {
-		for _, newPromMetric := range prom.Metrics {
-			if existingPromMetric.ValueKey == newPromMetric.ValueKey {
-				log.Panicf("error in parseEncode: ValueKey %s overlaps, metric encoding ignored !!!! ", newPromMetric.ValueKey)
-				return nil, err
-			}
-		}
-	}
-
+	
 	cg.promMetrics = append(cg.promMetrics, prom.Metrics...)
 	return &prom, nil
 }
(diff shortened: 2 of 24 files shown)


pkg/api/encode_prom.go
```diff
@@ -35,11 +35,17 @@ func PromEncodeOperationName(operation string) string {
...
+	Type     string            `yaml:"type" enum:"PromEncodeOperationEnum" doc:"one of the following:"`
+	Filter   PromMetricsFilter `yaml:"filter" doc:"the criterion to filter entries by"`
+	ValueKey string            `yaml:"valuekey" doc:"entry key from which to resolve metric value"`
+	Labels   []string          `yaml:"labels" doc:"labels to be associated with the metric"`
+	Buckets  []float64         `yaml:"buckets" doc:"histogram buckets"`
 }
 
 type PromMetricsItems []PromMetricsItem
+
+type PromMetricsFilter struct {
```
eranra, line 48: Do we need the prefix `Prom`  ... consider::  s/PromMetricsFilter /MetricsFilter

pkg/pipeline/encode/encode_prom.go
```diff
@@ -44,8 +44,10 @@ type PromMetric struct {
 }
 
 type metricInfo struct {
-	input      string
-	labelNames []string
+	input       string
+	filterKey   string
```
eranra, line 48: consider putting key and value in sub-struct  (I just think that will happen if we decide to change the parameters of the filter)

pkg/pipeline/encode/encode_prom.go
```diff
@@ -102,9 +104,15 @@ func (e *encodeProm) EncodeMetric(metric config.GenericMap) []config.GenericMap
 	// TODO: We may need different handling for histograms
 	out := make([]config.GenericMap, 0)
 	for metricName, mInfo := range e.metrics {
+		val, keyFound := metric[mInfo.filterKey]
+		shouldKeepRecord := keyFound && val == mInfo.filterValue
```
eranra, line 108: A very fast "thing" that you can improve here is to say ... if there is a Key and the value equals to some N/A value we filter just on the basis of the existence of the Key and we do not care about the value of the value ... trivial change and adds a lot of flexibility

pkg/confgen/encode.go
```diff
@@ -40,17 +40,7 @@ func (cg *ConfGen) parseEncode(encode *map[string]interface{}) (*api.PromEncode,
...
-	for _, existingPromMetric := range cg.promMetrics {
-		for _, newPromMetric := range prom.Metrics {
-			if existingPromMetric.ValueKey == newPromMetric.ValueKey {
-				log.Panicf("error in parseEncode: ValueKey %s overlaps, metric encoding ignored !!!! ", newPromMetric.ValueKey)
-				return nil, err
-			}
-		}
-	}
-
+
```
eranra: Any other cross phases check that you think we can and should add here???

eranra (approved):

pkg/api/encode_prom.go
ronensc: I added the the `Prom` prefix to follow the convention in the file.

pkg/pipeline/encode/encode_prom.go
ronensc: I see what you are suggesting and I agree that it will add flexibility. But, I think we should put more thought into that.
I don't like magic values that might collide with the user's values.

pkg/pipeline/encode/encode_prom.go
```diff
@@ -102,9 +104,15 @@ func (e *encodeProm) EncodeMetric(metric config.GenericMap) []config.GenericMap
...
+		if !shouldKeepRecord {
+			continue
+		}
+
 		metricValue, ok := metric[mInfo.input]
 		if !ok {
-			log.Debugf("field %v is missing", metricName)
+			log.Errorf("field %v is missing", mInfo.input)
```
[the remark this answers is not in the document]
ronensc: I believe the reason for using Debug originally is that before introducing dedicated `filter` field, the filtering was done based on the input field. So it wasn't an error if the expected input field didn't exist. It probably indicated that the record didn't belong to the metric in context.
The input field basically had two roles:
1. hold the value we wanted to measure
2. filter records - associate a record with a metric.

because of this, prefixes were added to the input key.
This PR allows removing these prefixes and remove the 2nd role of the input field.

pkg/confgen/encode.go
ronensc: Consider the following configuration of a prom metric
```yaml
      - name: bandwidth_per_network_service
        type: counter
        filter:
          key: name
          value: bandwidth_network_service
        valuekey: recent_op_value
        labels:
        - by
        - aggregate
        buckets: []
```


If `prom_encode` follows `extract_aggregate` I can think of the following validations:
1. `valuekey` is one of the keys of the following GenericMap https://github.com/netobserv/flowlogs-pipeline/blob/c0f17db3e2619c2cb95b6c945f2e7fa2982fed68/pkg/pipeline/extract/aggregate/aggregate.go#L189-L199
2. Same is true about `labels` and `filter.key`
3. if `filter.key == name` then we can validate that `filter.value` is defined.
4. regarding `type`
4.1. if `type == gauge` then `valuekey` is in `[total_value, total_count]`
4.2. if `type == counter` then `valuekey` is in `[recent_op_value, recent_count]`
4.3. if `type == histogram` then `valuekey` is `recentRawValues`
5. histograms have at least one bucket

Issue #151, opened 2022-03-18
Allow transform_generic to keep existing fields


amorenoz:
Currently transform_generic removes all fields from the flow if you don't explicitly set them in the configuration.
This makes it's usage quite limited because you have to know exactly all the fields that the log will contain or you cannot modify a single field. It would be great to have that behavior configurable through a flag.

Another option would be to have `transform_generic` as a simple field rewrite (without removing fields that are not explicitly set) and have another transformation stage where you can select the fields you want for your log. It could have a more semantically intuitive name such as `filter`or `select`

jotak:
sounds like a good enhancement to do, cc @eranra @KalmanMeth

KalmanMeth:
When the transform_generic was first written I had in mind that it would completely transform the input to some standard internal format. However, we can do whatever we now think is most appropriate, and update the unit tests accordingly.

amorenoz:
The problem with the current approach is that it limits the extensibility of the flow pipeline. If the flow exporter supports an additional field (through a newly defined template), we have to add it to the transform config list or else the new field will not be present in the log.

eranra:
#156 provides the needed functionality.

@joel **if needed** you can modify NOO configuration to use this updated behavior

PR #156, opened 2022-03-27
Allow transform_generic to keep existing fields

KalmanMeth:
Addresses issue #151

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,7 @@ parameters:
     transform:
       type: generic
       generic:
+        policy: replace_keys
         rules:
           - input: Bytes
             output: v1_bytes
@@ -168,6 +169,7 @@ parameters:
     transform:
       type: generic
       generic:
+        policy: replace_keys
         rules:
           - input: Bytes
             output: v2_bytes
@@ -249,7 +251,9 @@ The generic transform module maps the input json keys into another set of keys.
 This allows to perform subsequent operations using a uniform set of keys.
 In some use cases, only a subset of the provided fields are required.
 Using the generic transform, we may specify those particular fields that interest us.
-To include keys that are not specified in the `rules`, add the line `maintain: true`.
+Specify `policy: replace_keys` to use only the newly specified keys
+To include the original keys in addition to those specified in the `rules`,
+specify `policy: preserve_original_keys`.
 
 For example, suppose we have a flow log with the following syntax:
 ```
@@ -265,6 +269,7 @@ parameters:
     transform:
       type: generic
       generic:
+        policy: replace_keys
         rules:
           - input: Bytes
             output: bytes
@@ -299,6 +304,7 @@ parameters:
     transform:
       type: generic
       generic:
+        policy: replace_keys
         rules:
           - input: DstAddr
             output: dstAddr
@@ -308,6 +314,7 @@ parameters:
     transform:
       type: generic
       generic:
+        policy: replace_keys
         rules:
           - input: dstAddr
             output: dstIP
@@ -322,14 +329,14 @@ Before the first transform suppose we have the keys `DstAddr` and `SrcAddr`.
 After the first transform, we have the keys `dstAddr` and `srcAddr`.
 After the second transform, we have the keys `dstAddr`, `dstIP`, `srcAddr`, and `srcIP`.
 
-To maintain all the old key/values and change only the key `DstAddr` to `dstAddr`, use the following:
+To maintain all the old key/values and simply add the key `dstAddr` (derived from `DstAddr`), use the following:
 ```
 parameters:
   - name: transform1
     transform:
       type: generic
       generic:
-        maintain: true
+        policy: preserve_original_keys
         rules:
           - input: DstAddr
             output: dstAddr
--- a/README.md
+++ b/README.md
@@ -251,8 +251,8 @@ The generic transform module maps the input json keys into another set of keys.
 This allows to perform subsequent operations using a uniform set of keys.
 In some use cases, only a subset of the provided fields are required.
 Using the generic transform, we may specify those particular fields that interest us.
-Specify `policy: replace_keys` to use only the newly specified keys
-To include the original keys in addition to those specified in the `rules`,
+Specify `policy: replace_keys` to use only the newly specified keys.
+To include the original keys and values in addition to those specified in the `rules`,
 specify `policy: preserve_original_keys`.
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -249,6 +249,7 @@ The generic transform module maps the input json keys into another set of keys.
 This allows to perform subsequent operations using a uniform set of keys.
 In some use cases, only a subset of the provided fields are required.
 Using the generic transform, we may specify those particular fields that interest us.
+To include keys that are not specified in the `rules`, add the line `maintain: true`.
```
eranra, line 252: the term `maintain` is not very explanatory ... maybe use `preserveOriginalKeys` 

Also the explanation ... it is not that the fields are not specified in the `rules`, but that those are all fields in the original record ...

README.md
```diff
@@ -321,6 +322,19 @@ Before the first transform suppose we have the keys `DstAddr` and `SrcAddr`.
 After the first transform, we have the keys `dstAddr` and `srcAddr`.
 After the second transform, we have the keys `dstAddr`, `dstIP`, `srcAddr`, and `srcIP`.
 
+To maintain all the old key/values and change only the key `DstAddr` to `dstAddr`, use the following:
```
eranra, line 325: s/old/original
s/key/values/keys and values

pkg/pipeline/transform/transform_generic.go
```diff
@@ -33,21 +35,44 @@ func (g *Generic) Transform(input []config.GenericMap) []config.GenericMap {
 	output := make([]config.GenericMap, 0)
 	for _, entry := range input {
 		outputEntry := make(config.GenericMap)
-		for _, transformRule := range g.Rules {
+		for _, transformRule := range g.rules {
 			log.Debugf("transformRule = %v", transformRule)
 			outputEntry[transformRule.Output] = entry[transformRule.Input]
 		}
+		if g.maintain {
```
eranra, line 42: can you just change https://github.com/netobserv/flowlogs-pipeline/pull/156/files#diff-28f2a1789a319594e5a43929c866c072ac0e053716b9ba62a7158816d41621c8R35 to be a copy of input and not from an empty GenericMap

eranra, line 42: I expect this to be done before applying the rest of the rules (i.e. in line 35)

pkg/pipeline/transform/transform_generic.go
```diff
@@ -33,21 +35,44 @@ func (g *Generic) Transform(input []config.GenericMap) []config.GenericMap {
...
+		}
+		outputEntry[key] = value
+	}
+}
+
 // NewTransformGeneric create a new transform
 func NewTransformGeneric(params config.StageParam) (Transformer, error) {
 	log.Debugf("entering NewTransformGeneric")
+	log.Debugf("params.Transform.Generic = %v", params.Transform.Generic)
+	mapOfKeys := make(map[string]bool)
```
eranra, line 66: @jotak Is this "over-engineering" --- don't we just need to apply the rules on top of the original record ... adding extra fields if needed ????

KalmanMeth:
The default is now to include all previous key/values and to enrich them. To get only the newly specified keys, we now have `policy: replace_keys` as a parameter in the yaml.

pkg/pipeline/transform/transform_generic.go
jotak: I actually don't see `mapOfKeys` being used here

KalmanMeth: You are correct. `mapOfKeys`  is no longer used in the latest code and should be removed.

KalmanMeth:
addresses issue #151

eranra (commented):
@KalmanMeth looks good to me. You need to also update https://github.com/netobserv/flowlogs-pipeline/blob/main/pkg/confgen/flowlogs2metrics_config.go#L70 to reflect the new `policy` field

pkg/pipeline/transform/transform_generic.go
```diff
@@ -24,16 +24,22 @@ import (
...
 // Transform transforms a flow to a new set of keys
 func (g *Generic) Transform(input []config.GenericMap) []config.GenericMap {
-	log.Debugf("f = %v", g)
+	log.Debugf("entering Generic Transform g = %v", g)
 	output := make([]config.GenericMap, 0)
 	for _, entry := range input {
-		outputEntry := make(config.GenericMap)
-		for _, transformRule := range g.Rules {
+		outputEntry := entry
+		if g.policy == "replace_keys" {
```
ronensc, line 37: Do we want to panic on an invalid policy value? I mean, if the user writes `replace-keys` with a dash instead of an underline, we will treat it as `preserve_original_keys`

README.md
```diff
@@ -249,6 +251,9 @@ The generic transform module maps the input json keys into another set of keys.
 This allows to perform subsequent operations using a uniform set of keys.
 In some use cases, only a subset of the provided fields are required.
 Using the generic transform, we may specify those particular fields that interest us.
+Specify `policy: replace_keys` to use only the newly specified keys.
+To include the original keys and values in addition to those specified in the `rules`,
+specify `policy: preserve_original_keys`.
```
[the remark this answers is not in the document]
KalmanMeth: The new key overwrites the old value.

pkg/pipeline/transform/transform_generic.go
```diff
@@ -24,16 +24,22 @@ import (
...
 
 // Transform transforms a flow to a new set of keys
 func (g *Generic) Transform(input []config.GenericMap) []config.GenericMap {
-	log.Debugf("f = %v", g)
+	log.Debugf("entering Generic Transform g = %v", g)
 	output := make([]config.GenericMap, 0)
 	for _, entry := range input {
-		outputEntry := make(config.GenericMap)
-		for _, transformRule := range g.Rules {
+		outputEntry := entry
```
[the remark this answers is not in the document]
KalmanMeth: OK. Copied map entries individually.

pkg/pipeline/transform/transform_generic.go
KalmanMeth: added panic for illegal policy

eranra (approved):

README.md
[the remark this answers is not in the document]
ronensc: Do you think its worth noting in the docs?

PR #160, opened 2022-03-29
[NETOBSERV-237] gRPC+PB flow ingest/decoder for NetObserv eBPF Agent

mariomac:


pkg/pipeline/decode/decode_protobuf.go
```diff
@@ -0,0 +1,85 @@
...
+		"Proto":         flow.Transport.GetProtocol(),
+		"TimeFlowStart": flow.TimeFlowStart.GetSeconds(),
+		"TimeFlowEnd":   flow.TimeFlowEnd.GetSeconds(),
+		"TimeReceived":  time.Now().Unix(),
+		"Interface":     flow.Interface,
+	}
+	return out
+}
+
+func ipToStr(ip *pbflow.IP) string {
```
eranra, line 67: If `ipToStr` and `macToStr` are generic, maybe move to some utility or common file

pkg/pipeline/decode/decode_protobuf.go
```diff
@@ -0,0 +1,85 @@
...
+// as the IPFIX flows from ingest.IngestCollector
+type Protobuf struct {
+}
+
+func NewProtobuf() (Decoder, error) {
+	return &Protobuf{}, nil
+}
+
+// Decode decodes input strings to a list of flow entries
+func (c *Protobuf) Decode(in []interface{}) []config.GenericMap {
```
eranra, line 26: Are you using "c *Protobuf" in the function ???

pkg/pipeline/decode/decode_protobuf.go
```diff
@@ -0,0 +1,85 @@
...
+		out = append(out, pbFlowToMap(entry))
+	}
+	return out
+}
+
+func pbFlowToMap(flow *pbflow.Record) config.GenericMap {
+	if flow == nil {
+		return config.GenericMap{}
+	}
+	out := config.GenericMap{
```
eranra, line 47: Should this be built somehow dynamic from https://github.com/netobserv/netobserv-agent/blob/main/proto/flow.proto ????

pkg/pipeline/decode/decode_protobuf.go
mariomac: No, it is not used, but the API conventions of the flowlog pipeline require that `Decode` is run as a method of a type implementing the `Decoder` interface.

pkg/pipeline/decode/decode_protobuf.go
mariomac: That would involve flattening/changing the `flow.proto` definition to match the map structure, and then create some sort of code generation or use reflection to create the map. Since the number of fields is still relatively small, I'd keep using this explicit conversion, and consider another approach in the future if we feel that the fields grow until making this conversion unmaintainable.

eranra: @mariomac this is exactly my fear :-( .... that this will grow and no one will be able to find where this came from and why this is not working. Any chance to someone connect that to the source (i.e. to the https://github.com/netobserv/netobserv-agent) project ... if not in code than at least with some remarks maybe ?

pkg/pipeline/decode/decode_protobuf.go
mariomac: copy/paste issues. Changing to `p` or even `_`, as it is unused.

mariomac:
@eranra good point. There is already a task about that: https://issues.redhat.com/browse/NETOBSERV-199 but we didn't know exactly how to do it.

We can implement the tests in the E2E framework. Let's see if Kind+Github hosts allow installing the agent as a privileged Pod that can send eBPF bytecode to the kernel.

pkg/pipeline/decode/decode_protobuf.go
mariomac: Yeah, actually the `*pbflow.Records` type is imported from the `github.com/netobserv/netobserv-agent` directly, so IDE navigation leads you to the original protobuf definition (at least in Goland IDE).

But I agree that we should investigate a more automatic way to map protobuf --> generic map fields that does not penalize readability nor performance.

PR #173, opened 2022-04-06
NETOBSERV-229 create pre-merge image on PR via action

jotak:
Uses the 'ok-to-test' label to trigger the build/push

### DIFF
--- a/.github/workflows/push_image_pr.yml
+++ b/.github/workflows/push_image_pr.yml
@@ -1,18 +1,17 @@
-name: Build and push to quay.io
+name: Build and push PR image to quay.io
 on:
-  push:
-    branches: [ main ]
+  pull_request_target:
+    types: [labeled]
 
 env:
   REGISTRY_USER: netobserv+github_ci
-  REGISTRY_PASSWORD: ${{ secrets.QUAY_SECRET }}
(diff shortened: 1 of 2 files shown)


.github/workflows/push_image_pr.yml
```diff
@@ -0,0 +1,57 @@
...
+      - uses: actions/github-script@v5
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'New image: ${{ steps.push-to-quay.outputs.registry-paths }}. It will expire after two weeks.'
+            })
```
ronensc, line 57: I'm a bit confused. If this workflow is triggered by a PR, then what is the issue in context where the comment is created?

jotak: in github API, a PR is apparently some special kind of issue. But the comment is actually created on the PR

eranra (commented):
@jotak looks good to me

eranra (approved):

jotak:
thanks

PR #174, opened 2022-04-06
add custom network service names

eranra:


eranra:
@jpinsonneau I know that this is not used in the UI (the resolution is done in the UI) ... but still this adds the ability to custom the network services names using FLP - if we decide to use that <<--- maybe adding this feature also to the frontend makes sense.

jpinsonneau:
> @jpinsonneau I know that this is not used in the UI (the resolution is done in the UI) ... but still this adds the ability to custom the network services names using FLP - if we decide to use that <<--- maybe adding this feature also to the frontend makes sense.

Cool @eranra thanks for the update. @OlivierCazade is the one who implemented the config on front end side btw. 

I feel frontend config is enough for now but I agree we can add an option to be able to switch or even manage both with a merge mecanism :+1:

pkg/pipeline/transform/network_services/netdb.go
```diff
@@ -0,0 +1,207 @@
+// Package netdb provides a Go interface for the protoent and servent
```
KalmanMeth, line 1: Do we want to add the standard header we have on the other files with the Apache 2.0 license?

eranra: fixed

pkg/pipeline/transform/network_services/netdb.go
```diff
@@ -0,0 +1,217 @@
...
+
+		name := fields[0]
+		portproto := strings.SplitN(fields[1], "/", 2)
+		port, err := strconv.ParseInt(portproto[0], 10, 32)
+		if err != nil {
+			return err
+		}
+
+		proto := portproto[1]
+		aliases := fields[2:]
```
KalmanMeth, line 111: Is this OK if `fields` only has 2 entries i.e. if there is no alias specified for the service?

pkg/pipeline/transform/transform_network_test.go
```diff
@@ -313,3 +321,37 @@ func Test_Transform_AddIfScientificNotation(t *testing.T) {
...
+	err := netdb.Init("", "")
+	require.NoError(t, err)
+	output := networkTransform.Transform([]config.GenericMap{entry})
+	require.Equal(t, "ssh", output[0]["service"])
+
+	sshService := "custom_ssh             22/tcp                          # The Secure Shell (SSH) Protocol\n"
+	err = os.WriteFile("/tmp/services", []byte(sshService), 0644)
+	require.NoError(t, err)
+
+	err = netdb.Init("", "/tmp/services")
```
KalmanMeth, line 352: This test is for non-empty services file. A similar test should be done for non-empty protocol file.

KalmanMeth (approved):

pkg/pipeline/transform/network_services/netdb.go
eranra: Sure ... look on the following examples::

This is a line with no aliases::
`med-ci          24005/tcp               # med-ci`

This is a line **with** aliases::
`iceedcp_rx      31949/udp  iceedcp-rx   # Embedded Device Configuration Protocol RX` 

They are all parsed correctly as part of the `/etc/services` file

pkg/pipeline/transform/transform_network_test.go
eranra: Added another test for network protocols

PR #176, opened 2022-04-07
add operational_metrics support

eranra:


eranra:
@memodi please have a look at https://github.com/netobserv/flowlogs-pipeline/blob/e4c38d75c32ecc49b639fce89516a83e0bd272fc/docs/operational-metrics.md those are the new operational metrics that should help us to understand if and how FLP is working. If needed we can add more, please advise :-) 

@jotak FYI ^^^

pkg/pipeline/write/write_loki.go
```diff
@@ -58,6 +59,11 @@ type Loki struct {
 	exitChan   chan bool
 }
 
+var recordsWritten = operationalMetrics.NewCounter(prometheus.CounterOpts{
+	Name: "write_loki_records_written",
```
memodi, line 63: nit: on naming, simply "loki_records_written" would be more suitable?

pkg/pipeline/ingest/ingest_collector.go
```diff
@@ -57,6 +59,16 @@ type TransportWrapper struct {
 	c chan map[string]interface{}
 }
 
+var queueLength = operationalMetrics.NewGauge(prometheus.GaugeOpts{
+	Name: "ingest_collector_queue_length",
+	Help: "Queue length",
+})
+
+var linesProcessed = operationalMetrics.NewCounter(prometheus.CounterOpts{
+	Name: "ingest_collector_log_lines_processed",
```
memodi, line 68: sorry, again to pick on naming, would name such as "ingest_collector_flow_logs_processed" be more appropriate since its processing flow logs.

memodi (commented):
thanks @eranra  these will be certainly helpful. Added couple of comments on naming but probably have more feedback once I actually test it. 

Couple of questions:
1. are these enabled by default?
2. will these be pushed to prometheus?
3. I believe once this task https://issues.redhat.com/browse/NETOBSERV-247 is finished, we'd be also able to grab these metrics from the FLP endpoint, correct?

pkg/operational/metrics/metrics.go
```diff
@@ -0,0 +1,73 @@
...
+| **Name** | %s | 
+|:---|:---|
+| **Description** | %s | 
+| **Type** | %s | 
+
+`,
+			opts.Name,
+			opts.Name,
+			opts.Help,
+			opts.Type,
```
KalmanMeth, line 68: Whys does `opts.Name` appear twice?

KalmanMeth (approved):

eranra:
> thanks @eranra these will be certainly helpful. Added couple of comments on naming but probably have more feedback once I actually test it.
> 
> Couple of questions:
> 
> 1. are these enabled by default?

Yes, as long as you enable the Prometheus encode stage. Currently, this is NOT the configuration in NOO so we will need to change that for those metrics to be available for you :-) 

> 2. will these be pushed to Prometheus? 

YES

> 3. I believe once this task https://issues.redhat.com/browse/NETOBSERV-247 is finished, we'd be also able to grab these metrics from the FLP endpoint, correct?

@jotak please help with (3) --- I think that the answer is yes, but just want to be on the safe side here.

pkg/pipeline/write/write_loki.go
eranra: fixed

pkg/pipeline/ingest/ingest_collector.go
eranra: fixed

pkg/operational/metrics/metrics.go
eranra: Once in the title as the header and once in the name of the metric. I am keeping the format similar to the network metrics for now ... going forward we might change :-)

ronensc (commented):
I like the idea of auto-generated documentation of the operational metrics.

cmd/operationalmetricstodoc/main.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+
+	operationalMetrics "github.com/netobserv/flowlogs-pipeline/pkg/operational/metrics"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline"
+)
+
+func main() {
+	var _ *pipeline.Pipeline
```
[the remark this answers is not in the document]
eranra: This causes golang to execute all internal Init operations and instantiate all variables from the pipeline packages  ... and this fills up the `metricsOpts` variable with all the information to generate the docs. Without this unnamed variable, the golang optimizer ( I guess ) skips linking / using those packages.

docs/operational-metrics.md
```diff
@@ -0,0 +1,39 @@
+
+> Note: this file was automatically generated, to update execute "make docs"  
+  
+# flowlogs-pipeline Operational Metrics  
+  
+Each table below provides documentation for an exported flowlogs-pipeline operational metric.
```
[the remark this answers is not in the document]
eranra: @ronensc It might ... if this is ok with you, I would like to keep this as-is for now. This is aligned with the other metrics (network metrics) documentation format ... I agree that we can change this / improve etc ... I was thinking to add a description field with details etc. I hope that the RH team will be able to put that in front of tech-writer so get professional review :-)

pkg/pipeline/write/write_loki.go
```diff
@@ -123,6 +129,8 @@ func (l *Loki) ProcessRecord(record config.GenericMap) error {
 	if err != nil {
 		return err
 	}
+
+	recordsWritten.Inc()
 	return l.client.Handle(labels, timestamp, string(js))
```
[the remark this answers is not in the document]
eranra: fixed :-)

cmd/operationalmetricstodoc/main.go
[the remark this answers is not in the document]
eranra: Added :-)

PR #178, opened 2022-04-10
add support for deleting expired aggregates

eranra:


pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -213,3 +229,21 @@ func (aggregate Aggregate) GetMetrics() []config.GenericMap {
 
 	return metrics
 }
+
+func (aggregate Aggregate) cleanupExpiredEntries() {
+	nowInSecs := time.Now().Unix()
+	expireTime := nowInSecs - expiryTime
```
KalmanMeth, line 235: Where is `expiryTime` defined?
The same `expiryTime` (default value) for all aggregates? Not configurable? OK

pkg/pipeline/extract/aggregate/aggregate_test.go
```diff
@@ -116,8 +118,8 @@ func Test_Evaluate(t *testing.T) {
 	err := aggregate.Evaluate(entries)
 
 	require.Equal(t, err, nil)
-	require.Equal(t, aggregate.Groups[normalizedValues].totalCount, 2)
-	require.Equal(t, aggregate.Groups[normalizedValues].totalValue, float64(7))
+	require.Equal(t, aggregate.GroupsMap[normalizedValues].totalCount, 2)
+	require.Equal(t, aggregate.GroupsMap[normalizedValues].totalValue, float64(7))
```
KalmanMeth, line 122: You can also add here a test on the length of `GroupsList`.

pkg/pipeline/extract/aggregate/aggregates.go
```diff
@@ -18,17 +18,26 @@
...
 )
 
+const defaultExpiryTime = 60 * 10 // 10 minutes
+
 type Aggregates []Aggregate
 type Definitions []api.AggregateDefinition
 
+var (
+	mutex      = &sync.Mutex{}
+	expiryTime = int64(defaultExpiryTime)
```
KalmanMeth, line 38: This a global mutex and expirtyTIme for all Aggregates. Is this the desired functionality? Is everything anyway done serially? OK

pkg/pipeline/extract/aggregate/aggregates.go
```diff
@@ -54,20 +63,39 @@ func (aggregates Aggregates) GetMetrics() []config.GenericMap {
...
-		Groups:     map[NormalizedValues]*GroupState{},
+		GroupsMap:  map[NormalizedValues]*GroupState{},
+		GroupsList: list.New(),
 	}
 
 	appendedAggregates := append(aggregates, aggregate)
 	return appendedAggregates
 }
 
-func (aggregates Aggregates) RemoveAggregate(by api.AggregateBy) (Aggregates, error) {
```
KalmanMeth, line 64: When was `RemoveAggregate` ever called? No longer relevant.

pkg/pipeline/extract/aggregate/aggregates.go
```diff
@@ -54,20 +63,39 @@ func (aggregates Aggregates) GetMetrics() []config.GenericMap {
...
 		}
+	}()
+}
+
+func (aggregates Aggregates) cleanupExpiredEntries() {
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	for _, aggregate := range aggregates {
+		aggregate.cleanupExpiredEntries()
```
KalmanMeth, line 96: Each aggregate has its own cleanup loop. If so, then can conflict on the common mutext lock. Why don't you put a separate mutex on each aggregate so there won't be any conflicts between them?

KalmanMeth (approved):

pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -213,3 +229,21 @@ func (aggregate Aggregate) GetMetrics() []config.GenericMap {
 
 	return metrics
 }
+
+func (aggregate Aggregate) cleanupExpiredEntries() {
+	nowInSecs := time.Now().Unix()
```
ronensc, line 234: IMO comparing `Time` objects rather than ints of seconds from Epoch time, is more idiomatic to Go.

pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -43,7 +45,8 @@ type NormalizedValues string
 
 type Aggregate struct {
 	Definition api.AggregateDefinition
-	Groups     map[NormalizedValues]*GroupState
+	GroupsMap  map[NormalizedValues]*GroupState
+	GroupsList *list.List
```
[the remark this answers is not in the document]
eranra: @ronensc you are correct that they are similar. I started to generalize but saw that it adds a lot of code and complexity. I am not adding this code to this PR and we need to consider if this makes sense or not. I added an issue https://github.com/netobserv/flowlogs-pipeline/issues/181 to track.

pkg/pipeline/extract/aggregate/aggregate.go
eranra: I added this as a point to issue https://github.com/netobserv/flowlogs-pipeline/issues/181 ... this is currently the same behavior we have in all other places. I keep this as-is for this PR and we will fix it when moving to generic LRU with expiration cache.

pkg/pipeline/extract/aggregate/aggregate.go
eranra: Currently --- yes one default - I added code to have different expirations for each aggregate but for now this is common to all. Also, currently, my thinking is that this is a big hardcoded value (10 minutes). When we will start to benchmark, if needed we will add that to the API and split to fit the needs of each aggregate. At this point, I do not see a good reason to differentiate between the current aggregates that we have in the default FLP confiuration

pkg/pipeline/extract/aggregate/aggregate_test.go
eranra: done

pkg/pipeline/extract/aggregate/aggregates.go
eranra: Agree - I moved the mutext to the aggregate level - no need for global mutext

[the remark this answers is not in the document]
eranra: Agree - not changing for now just because we have that also in all other places and this will be handled when we generalize the LRU with expiration cache

pkg/pipeline/extract/aggregate/aggregates.go
eranra: right, it is not needed this is why I removed

pkg/pipeline/extract/aggregate/aggregates.go
eranra: done

pkg/pipeline/extract/aggregate/aggregates_test.go
```diff
@@ -44,10 +45,30 @@ parameters:
...
-	require.NoError(t, err)
 	require.Equal(t, aggregates[0].Definition, expectedAggregate.Definition)
 }
+
+func Test_CleanupExpiredEntriesLoop(t *testing.T) {
+
+	expiryTime = 1
+	aggregates := initAggregates(t)
+	expectedAggregate := GetMockAggregate()
+	require.Equal(t, aggregates[0].Definition, expectedAggregate.Definition)
```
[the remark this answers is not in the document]
eranra: fixed

PR #182, opened 2022-04-14
kafka e2e test

eranra:


pkg/test/e2e/kafka/kafka_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func postFLPDeploy(ctx context.Context, cfg *envconf.Config, namespace string) error {
+	fmt.Printf("====> Waiting for FLP Deployment\n")
+	client, err := cfg.NewClient()
+	if err != nil {
+		return fmt.Errorf("postStrimziDeploy:NewClient error:%v", err)
```
KalmanMeth, line 44: s/postStrimziDeploy/postFLPDeploy/

pkg/test/e2e/utils.go
```diff
@@ -48,36 +44,54 @@ import (
...
-		envfuncs.DestroyKindCluster(kindClusterName),
+		e2eDeleteKindCluster(kindClusterName),
 	)
 	os.Exit((*testEnv).Run(m))
 }
 
 func e2eCreateKindCluster(clusterName string) env.Func {
 	return func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
 		fmt.Printf("====> Creating KIND cluster - %s\n", clusterName)
+		gexe.RunProc(fmt.Sprintf(`kind delete cluster --name %s`, clusterName))
```
KalmanMeth, line 78: Perhaps print "Creating KIND cluster" after the delete operation (in which you print "Deleting KIND cluster").

KalmanMeth (approved):

pkg/test/e2e/kafka/kafka_test.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	dep := appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{Name: "flowlogs-pipeline", Namespace: namespace},
+	}
+	// wait for the deployment to finish becoming available
+	err = wait.For(
+		conditions.New(client.Resources()).DeploymentConditionMatch(&dep, appsv1.DeploymentAvailable, v1.ConditionTrue), wait.WithTimeout(time.Minute*3))
+	if err != nil {
+		return fmt.Errorf("postStrimziDeploy:wait.For error:%v", err)
```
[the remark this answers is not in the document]
eranra: fixed

pkg/test/e2e/kafka/kafka_suite_test.go
```diff
@@ -0,0 +1,48 @@
...
+	},
+	e2e.ManifestDeployDefinition{
+		YamlFile: "strimzi.yaml",
+	},
+	e2e.ManifestDeployDefinition{
+		YamlFile:     "kafka.strimzi.yaml",
+		PostFunction: postStrimziDeploy,
+	},
+}
+
```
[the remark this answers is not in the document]
eranra: moved

pkg/test/e2e/kafka/kafka_test.go
eranra: fixed

pkg/test/e2e/utils.go
eranra: I wanted to encapsulate all the process of creation with those prints ... the "delete" is somehow part of the creation or recreation of the kind cluster. I renamed to `Recreating` so this is more simple to understand

pkg/test/e2e/kafka/kafka_test.go
```diff
@@ -0,0 +1,152 @@
...
+			}
+			errorsExist := strings.Contains(logs, "ERROR")
+			if errorsExist {
+				t.Fatal("flowlogs-pipeline reports errors ")
+			}
+
+			command := "kubectl get kafka my-cluster -o=jsonpath='{.status.listeners[?(@.type==\"external\")].bootstrapServers}'"
+			kafkaAddr := test.RunCommand(command)
+			// strip the quotation marks
+			kafkaServer = kafkaAddr[1 : len(kafkaAddr)-1]
```
ronensc, line 128: `kafkaServer` is a variable defined in `kafka_utils.go`. Do we want to add characters to its name to indicate that it's a package global variable rather than a local variable?

pkg/test/e2e/utils.go
```diff
@@ -48,36 +44,54 @@ import (
...
 	)
 	os.Exit((*testEnv).Run(m))
 }
 
-func e2eCreateKindCluster(clusterName string) env.Func {
+func e2eRecreateKindCluster(clusterName string) env.Func {
 	return func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
-		fmt.Printf("====> Creating KIND cluster - %s\n", clusterName)
+		fmt.Printf("====> Recreating KIND cluster - %s\n", clusterName)
+		gexe.RunProc(fmt.Sprintf(`kind delete cluster --name %s`, clusterName))
```
ronensc, line 78: Why don't you use `envfuncs.DestroyKindCluster()`?

pkg/test/e2e/utils.go
```diff
@@ -48,36 +44,54 @@ import (
...
 		fmt.Printf("====> Done.\n")
 		return newCtx, err
 	}
 }
 
+func e2eDeleteKindCluster(clusterName string) env.Func {
+	return func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
+		fmt.Printf("====> Deleting KIND cluster - %s\n", clusterName)
+		newCtx, err := envfuncs.DestroyKindCluster(clusterName)(ctx, cfg)
+		gexe.RunProc(fmt.Sprintf(`kind delete cluster --name %s`, clusterName))
```
ronensc, line 89: Isn't `envfuncs.DestroyKindCluster()` enough?

ronensc (commented):
Looks good

Makefile
```diff
@@ -103,7 +103,7 @@ tests-unit: validate_go ## Unit tests
 
 .PHONY: tests-e2e
 tests-e2e: validate_go $(KIND)  ## End-to-end tests
-	go test -v -timeout 1200s -race $$(go list ./... | grep  /e2e)
+	go test -p 1 -v -timeout 1200s -race $$(go list ./... | grep  /e2e)
```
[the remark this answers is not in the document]
eranra: Not allowing parallel execution of tests ... this is needed because we use kind and we do parallel context for k8s is problematic for `kubectl` and client commands. Theoretically, we can enable parallel executions and e2e test time but not for this PR. Maybe in the future.

pkg/test/e2e/utils.go
```diff
@@ -152,65 +173,41 @@ func e2eRecreateNamespace(name string) env.Func {
...
-		}
-
-		if !acceptedK8sTypes.MatchString(groupVersionKind.Kind) {
-		} else {
-			resources = append(resources, resource)
-		}
+func deployManifest(ctx context.Context, cfg *envconf.Config, namespace string, yamlFileName string) error {
+	fmt.Printf("====> Deploying k8s manifest - %s\n", yamlFileName)
+	cmd := fmt.Sprintf("cat %s | kubectl apply -n %s -f -", yamlFileName, namespace)
+	out, err := exec.Command("bash", "-c", cmd).Output()
```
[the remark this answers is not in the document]
eranra: " Can't we achieve the effect of kubectl apply with client-go?" ---> unfortunately no, that was the original code and I had to change this due to the need to deploy CR. We can't parse generic CR's and use the client ... so the best and probably only way to deploy is using the `kubectl apply` command. 

Two different libs that are doing plus-minus the same thing. Eventually this is calling the same underlined code.

pkg/test/e2e/kafka/kafka_test.go
```diff
@@ -0,0 +1,152 @@
...
+			if !startExist {
+				t.Fatal("can't find flowlogs-pipeline start message")
+			}
+			errorsExist := strings.Contains(logs, "ERROR")
+			if errorsExist {
+				t.Fatal("flowlogs-pipeline reports errors ")
+			}
+
+			command := "kubectl get kafka my-cluster -o=jsonpath='{.status.listeners[?(@.type==\"external\")].bootstrapServers}'"
+			kafkaAddr := test.RunCommand(command)
```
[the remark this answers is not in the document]
eranra: I am not sure ... this is reading a CR ... not part of the k8s corev1 client lib. We can probably switch to some other lib but I prefer not to change this code (from @KalmanMeth ) at this stage. If needed we will do that in a following PR.

Makefile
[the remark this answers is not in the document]
ronensc: I see. Do you think it worth adding a comment to explain this?

pkg/test/e2e/kafka/kafka_test.go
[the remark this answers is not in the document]
eranra: fixed

eranra: Change to `theKafkaServer`

pkg/test/e2e/utils.go
eranra: It should ... I removed that line (this was part of my tests)

pkg/test/e2e/utils.go
eranra: Because it is not in the `ctx` yet, so we can't use this function .... in the end of the day if you look on the underline implementation ... it is doing the same thing .... just running a process ... so I do not see any value in using that

pkg/test/e2e/kafka/kafka_test.go
```diff
@@ -0,0 +1,152 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			logs := e2e.LogsFromPods(pods, coreV1Client, cfg.Namespace())
+			fmt.Print(logs)
+			startExist := strings.Contains(logs, "Starting flowlogs-pipeline")
+			if !startExist {
+				t.Fatal("can't find flowlogs-pipeline start message")
+			}
```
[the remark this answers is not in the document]
eranra: Maybe in some sense, but since we want to dump logs, this is actually looking for the start log lines .... we do not want to check here constantly the health, just one time that it started.

PR #188, opened 2022-04-22
Fix shutdown time

mariomac:
Some nodes, such as the GRPC+Protobuf server, are not very friendly to shutdown signals from Kubernetes, and may take up to 1 minute before Kubernetes eventually kills the pods. This makes undeployments and redeployments to be very slow.

This PR captures the SIGTERM signals and stops the main goroutine when they are received. The pipeline runs in background so it is automatically stopped when the main goroutine ends.

### DIFF
--- a/pkg/pipeline/encode/encode_prom.go
+++ b/pkg/pipeline/encode/encode_prom.go
@@ -82,7 +82,7 @@ type encodeProm struct {
 	expiryTime int64
 	mList      *list.List
 	mCache     metricCache
-	exitChan   chan bool
+	exitChan   chan struct{}
 }
 
 var metricsProcessed = operationalMetrics.NewCounter(prometheus.CounterOpts{
@@ -336,7 +336,7 @@ func NewEncodeProm(params config.StageParam) (Encoder, error) {
 		}
 	}
 
-	ch := make(chan bool, 1)
+	ch := make(chan struct{})
 	utils.RegisterExitChannel(ch)
 
 	log.Debugf("metrics = %v", metrics)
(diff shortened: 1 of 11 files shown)


eranra:
@mariomac thanks for adding this PR ... there is a centralized process to exit all threads in FLP ... the code is here: https://github.com/netobserv/flowlogs-pipeline/blob/main/pkg/pipeline/utils/exit.go#L46 can you use the `RegisterExitChannel` function here https://github.com/netobserv/flowlogs-pipeline/blob/main/pkg/pipeline/utils/exit.go#L34  so that when we get `SIGTERM` signal the pipeline will exit (as well as all other threads in the FLP process). This should be very similar to what you added just using that centralized process exit code

cmd/flowlogs-pipeline/main.go
```diff
@@ -177,7 +179,15 @@ func run() {
...
 	// Starts the flows pipeline
-	mainPipeline.Run()
+	stopper := make(chan os.Signal, 1)
+	go func() {
+		mainPipeline.Run()
+		close(stopper)
+	}()
+
+	// Subscribe to signals for terminating the program.
+	signal.Notify(stopper, os.Interrupt, syscall.SIGTERM)
```
KalmanMeth, line 189: This has to be integrated to work with exiting all the other threads that exit when SIGINT or SIGTERM are received. See SetupElegantExit() in flowlogs-pipeline/pkg/pipeline/utils/exit.go

mariomac:
Thanks for the info, @eranra & @KalmanMeth ! I didn't realized this elegant shutdown mechanism.

I adapted the GRPC ingester to work with it and also changed the mechanism itself to use the standard practices in Golang:

Instead of a buffered bool channel, I created an unbuffered `struct{}` channel (the empty type, as the actual value is not imporant).

Also, the common practice when you want to use a channel as a barrier is not to send a message and read it, but close it. When you try to read a closed channel, the read operation will continue. This actually allows to share a single channel that is closed once and will allow the rest of instances to block until it is closed.

In this PR, I am still using one single channel for ingester, but in a later PR we can share the same channel for all the instances that want the "elegant" shutdown.

eranra (approved):
Hi @mariomac ... thanks for improving the mechanism ... very nice and clean ;-)

PR #190, opened 2022-04-25
simplify elegant shutdown

mariomac:
This PR simplifies elegant shutdown by sharing a single channel for all the components that want to listen for a shutdown signal.

Instead of a global variable, ideally this could be achieved by passing a cancellable context to all the running operations but the pipes library is not yet prepared for that (in my to-do list).

### DIFF
--- a/pkg/pipeline/ingest/ingest_kafka_test.go
+++ b/pkg/pipeline/ingest/ingest_kafka_test.go
@@ -124,10 +124,6 @@ func Test_IngestKafka(t *testing.T) {
 	require.Equal(t, record1, receivedEntries[0])
 	require.Equal(t, record2, receivedEntries[1])
 	require.Equal(t, record3, receivedEntries[2])
-
-	// make the ingest thread exit
-	close(ingestKafka.exitChan)
-	time.Sleep(time.Second)
 }
 
 type fakeKafkaReader struct {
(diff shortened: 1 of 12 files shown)


eranra (approved):

pkg/pipeline/ingest/ingest_kafka_test.go
```diff
@@ -124,10 +124,6 @@ func Test_IngestKafka(t *testing.T) {
 	require.Equal(t, record1, receivedEntries[0])
 	require.Equal(t, record2, receivedEntries[1])
 	require.Equal(t, record3, receivedEntries[2])
-
-	// make the ingest thread exit
-	close(ingestKafka.exitChan)
```
eranra, line 129: @mariomac This might cause failure in e2e test

pkg/pipeline/ingest/ingest_kafka_test.go
```diff
@@ -177,9 +173,4 @@ func Test_KafkaListener(t *testing.T) {
 
 	require.Equal(t, 1, len(receivedEntries))
 	require.Equal(t, string(fakeRecord), receivedEntries[0])
-
-	// make the ingest thread exit
-	close(ingestKafka.exitChan)
```
eranra, line 182: @mariomac This might cause failure in e2e test

eranra: @mariomac I reran the e2e and I think they failed due to glitch in github networking (access and not due to this) but we need to keep eyes open.

mariomac: Yeah it looked like an issue with networking but let's keep an eye.

In theory, when the test is finished the channels should be closed automatically by the runtime. We can do this operation explicit but it would make the test setup a bit less clear.

Issue #148, opened 2022-03-16
Improve local_deploy


eranra:
Improve the deployment so that it builds the latest version of the docker image locally ... push into the local kind cluster and use this image instead of the image from the internet. This will allow easy deployment of the latest version on Kind

Note: Need to change https://github.com/netobserv/flowlogs-pipeline/blob/main/contrib/kubernetes/deployment.yaml#L26 from `Always` to `Never` to be on the safe side that we are using the latest version.

PR #192, opened 2022-04-27
Load local image in local-deploy

ronensc:
Solves #148

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,9 +148,19 @@ push-image: build-image ## Push latest image
 
 # note: to deploy with custom image tag use: DOCKER_TAG=test make deploy
 .PHONY: deploy
-deploy: ## Deploy the image
+deploy: build-image ## Deploy the image
 	sed 's|%DOCKER_IMG%|$(DOCKER_IMG)|g;s|%DOCKER_TAG%|$(DOCKER_TAG)|g' contrib/kubernetes/deployment.yaml > /tmp/deployment.yaml
 	kubectl create configmap flowlogs-pipeline-configuration --from-file=flowlogs-pipeline.conf.yaml=$(FLP_CONF_FILE)
+# The following "if" statement is still part of the "deploy" recipe. It couldn't be indented with tabs because it will
+# be considered as part of the recipe. It could be indented with spaces but I find it confusing.
+# https://stackoverflow.com/a/28720186/2749989
+ifeq ($(OCI_RUNTIME),$(shell which docker))
+# This is an optimization for docker provider. "kind load docker-image" can load an image directly from docker's
+# local registry. For other providers (i.e. podman), we must use "kind load image-archive" instead.
+	$(KIND) load docker-image $(DOCKER_IMG):$(DOCKER_TAG)
+else
+	$(OCI_RUNTIME) save $(DOCKER_IMG):$(DOCKER_TAG) | kind load image-archive -
+endif
 	kubectl apply -f /tmp/deployment.yaml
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -148,9 +148,19 @@ push-image: build-image ## Push latest image
 
 # note: to deploy with custom image tag use: DOCKER_TAG=test make deploy
 .PHONY: deploy
-deploy: ## Deploy the image
+deploy: build-image ## Deploy the image
 	sed 's|%DOCKER_IMG%|$(DOCKER_IMG)|g;s|%DOCKER_TAG%|$(DOCKER_TAG)|g' contrib/kubernetes/deployment.yaml > /tmp/deployment.yaml
 	kubectl create configmap flowlogs-pipeline-configuration --from-file=flowlogs-pipeline.conf.yaml=$(FLP_CONF_FILE)
+# The following "if" statement is still part of the "deploy" recipe. It couldn't be indented with tabs because it will
+# be considered as part of the recipe. It could be indented with spaces but I find it confusing.
+# https://stackoverflow.com/a/28720186/2749989
```
ronensc, line 156: From @eranra: This comment is obvious and should be removed.

contrib/kubernetes/deployment.yaml
```diff
@@ -23,7 +23,7 @@ spec:
             - containerPort: 6343
             - containerPort: 2055
             - containerPort: 2056
-          imagePullPolicy: Always
+          imagePullPolicy: Never
```
ronensc, line 26: This should be set for `Never` only for local-deployment.

eranra: @ronensc maybe "IfNotPresent" fits both use-cases ... and actually if the code for OCP like in here https://cookbook.openshift.org/image-registry-and-image-streams/how-do-i-import-an-image-from-an-external-image.html works than you can keep the `Never` approach and add the code for OCP deployment to also load an image ... I really prefer the second, but it does require a bit of more work to check that this works also for OCP.

eranra (approved):
Looks very good to me

contrib/kubernetes/deployment.yaml
ronensc: I tried loading the FLP image into an OCP cluster and then started a pod with that image
```
$ oc import-image quay.io/netobserv/flowlogs-pipeline:latest --confirm
$ oc run  mypod --image=quay.io/netobserv/flowlogs-pipeline:latest --image-pull-policy=Never
```

But the pod couldn't find the image
```
$ oc get po
NAME    READY   STATUS              RESTARTS   AGE
mypod   0/1     ErrImageNeverPull   0          4s
```


Also, AFAIU, image streams are an OpenShift feature and there is no `import-image` equivalent in `kubectl`.
Do we want to introduce a new `oc` command to our Makefile?

eranra: @ronensc I was looking on https://medium.com/@adilsonbna/importing-an-external-docker-image-into-red-hat-openshift-repository-c25894cd3199 

This seems to solve this ... but this is also not trivial and makes things even a bit more complicated 

So, what I suggest is a keep things as is and move to `IfNotPresent` mode ... this should work for both.
Also, small remark in the makefile of OCP that states that this uses the latest version from `quay.io` 

Hope this makes sense to you

We might improve in one of the future PR's

eranra: @jotak @mariomac @jpinsonneau @OlivierCazade if any of you guys know how to push an image from local docker into OCP (above links are almost working, but not really :-) ))) we can use that to improve the process of ocp-deploy

PR #193, opened 2022-04-28
added topk functionality to aggregate

KalmanMeth:


eranra (commented):
@KalmanMeth I asked @ronensc to review --- this needs to be aligned to the connection tracking code that he started to write ...

pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -203,6 +203,52 @@ func (aggregate Aggregate) Evaluate(entries []config.GenericMap) error {
...
+			// find updated minRecentCount
+			newMinRecentCount := currentRecentCount
+			for i, m := range outputMetrics {
+				count := m["recent_count"].(int)
+				if count < newMinRecentCount {
+					newMinRecentCount = count
+					indexMin = i
+				}
+			}
+			minRecentCount = newMinRecentCount
```
ronensc, line 245: Finding the new min could be faster using a min heap. 
Alternatively, we can use Selection algorithm to find the k-th value in O(n) and then find all the values that are larger.
Do you think it worth the extra work?
https://stackoverflow.com/a/4961130/2749989

KalmanMeth: Assuming the topk value is small, I don't think it is necessary to overly complicate the algorithm. We can discuss further.

pkg/api/extract_aggregate.go
```diff
@@ -8,4 +8,5 @@ type AggregateDefinition struct {
 	By        AggregateBy        `yaml:"By" doc:"list of fields on which to aggregate"`
 	Operation AggregateOperation `yaml:"Operation" doc:"sum, min, max, avg or raw_values"`
 	RecordKey string             `yaml:"RecordKey" doc:"internal field on which to perform the operation"`
+	TopK      int                `yaml:"TopK" doc:"number of hightest incidence to report (default - report all)"`
```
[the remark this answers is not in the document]
KalmanMeth: fixed

pkg/pipeline/extract/aggregate/aggregate.go
KalmanMeth: I changed the code to use a heap of size topk.

pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -250,3 +255,66 @@ func (aggregate Aggregate) cleanupExpiredEntries() {
...
+	h := &topkHeap{}
+	for index, metricMap := range inputMetrics {
+		item := heapItem{
+			metrics: &inputMetrics[index],
+			value:   metricMap["total_value"].(float64),
+		}
+		heap.Push(h, item)
+		if h.Len() > topk {
+			heap.Pop(h)
+		}
```
ronensc, line 306: Not super critical but this could be optimized a bit further.
we "pay" `O(log (k))` for every Push and Pop operation. If we wrap this section with `if itemValue > minimum value of the heap` then we can save some cycles.

pkg/pipeline/extract/aggregate/aggregate_test.go
```diff
@@ -138,6 +136,6 @@ func Test_GetMetrics(t *testing.T) {
 
 	require.Equal(t, len(metrics), 1)
 	require.Equal(t, metrics[0]["name"], aggregate.Definition.Name)
-	valueFloat64, _ := strconv.ParseFloat(fmt.Sprintf("%s", metrics[0]["total_value"]), 64)
+	valueFloat64 := metrics[0]["total_value"].(float64)
```
ronensc, line 139: 👍

ronensc (commented):
Looks good

pkg/api/enum.go
```diff
@@ -36,9 +37,13 @@ type enumNameCacheKey struct {
 }
 
 var enumNamesCache = map[enumNameCacheKey]string{}
+var enumMutex = &sync.Mutex{}
 
 // GetEnumName gets the name of an enum value from the representing enum struct based on `TagYaml` tag.
 func GetEnumName(enum interface{}, operation string) string {
+	enumMutex.Lock()
+	defer enumMutex.Unlock()
+
```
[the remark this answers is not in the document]
KalmanMeth: This is to avoid a race condition reported by `go test -race`.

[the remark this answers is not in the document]
ronensc: I see. Although, I don't understand why there is a race condition if we run the unit tests with `-p 1`.
https://github.com/netobserv/flowlogs-pipeline/blob/be8aebf613331797935035be7cad816a0d8ba891/Makefile#L103
Anyway, I'm a bit concerned (from performance PoV) about acquiring the lock for every call to this low level function. Maybe we can use `RWMutex` to mitigate this issue?

ronensc (approved):

pkg/api/enum.go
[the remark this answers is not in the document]
KalmanMeth: Removed mutex and implemented populateEnumCache() during init().

ronensc (approved):

PR #195, opened 2022-05-01
use ubi8 container base image

KalmanMeth:


KalmanMeth:
During testing:
#19 [builder 11/13] RUN rm -rf bin
#19 sha256:e43a290424418aed9f6c6df124a2219c17fe689013c400100d48d81d458e665e
#19 0.983 rm: cannot remove 'bin/golangci-lint-v1.43.0': Permission denied
#19 0.983 rm: cannot remove 'bin/kind-v0.11.1': Permission denied
#19 ERROR: executor failed running [/bin/sh -c rm -rf bin]: exit code: 1

Not sure why permission is denied.

Commented out the command `RUN rm -rf bin`.

eranra:
@jotak can you please check this PR ... is there a better way to get golang 1.17 ???

contrib/docker/Dockerfile
```diff
@@ -12,7 +21,7 @@ RUN go mod download
 RUN go mod download -modfile=.bingo/golangci-lint.mod
 
 COPY . ./
-RUN rm -rf bin
+#RUN rm -rf bin
```
eranra, line 24: The need to delete the bin folder is to make sure that we are re-building/downloading the binaries using .bingo. when we build the container image and not use the downloaded ones from git

ronensc: I think a better way to get rid of the bin directory in the docker image is to add a `.dockerignore` file to the repository with a line to exclude `bin/`.
https://docs.docker.com/engine/reference/builder/#dockerignore-file

contrib/docker/.dockeringnore
```diff
@@ -0,0 +1,2 @@
+.git
+/bin/
```
ronensc, line 2: I found that there is an inconsistency between `docker` and `podman` regarding handling `.dockerignore`.
While both process it when it is in the repo's root directory, `docker` also allows `<dockerfile-name>.dockerignore`
https://stackoverflow.com/a/57774684/2749989

ronensc (commented):
Regarding the timeouts in the e2e tests, I recommend adding logging of timestamps in strategic places so we could measure which parts takes the most time to run.

jotak:
about the go-toolset image and go1.17: we have the same workaround on other images. Maybe we can have our own "netobserv/go-toolset" image to avoid fixing all components one by one ...

jotak:
@KalmanMeth @eranra the final stage here is still using ubuntu, I think we need either to change all stages to ubi8, or none (the whole purpose being to have a common base across components to make it easier to maintain / check vulnerabilities & upgrades).

But to be honest, I see also that there's included some net tools with the image, I don't want to force moving to ubi8 if you have the tooling you want with this current Dockerfile, the goal is not to make troubleshooting harder.

eranra:
> @KalmanMeth @eranra the final stage here is still using ubuntu, I think we need either to change all stages to ubi8, or none (the whole purpose being to have a common base across components to make it easier to maintain / check vulnerabilities & upgrades).
> 
> But to be honest, I see also that there's included some net tools with the image, I don't want to force moving to ubi8 if you have the tooling you want with this current Dockerfile, the goal is not to make troubleshooting harder.

@KalmanMeth what we want to make sure is that even with ubi8 we still have the simple tools to understand processes and networks. I also assume that the ubi8 image do not include or allow any package manager such as yum or apt so what I propose is to copy get those tools in the builder stage and make sure that we copy them into the ubi container. Exacely like what we are doing to the executables we compile https://github.com/netobserv/flowlogs-pipeline/pull/195/files#diff-e818f6b11598d2656922413d3912abb820c175e2f739549dbda55cc9559bd6fdL28 

This will allow us on one hand to work with ubi8 as needed for security but have those tools handy if needed. 

@jotak does this make sense and acceptable?

eranra (approved):

jotak:
thanks @KalmanMeth @eranra and sorry for the delay 😬

/lgtm

PR #196, opened 2022-05-02
NETOBSERV-253 Enrich Host informations

jpinsonneau:
This PR adds HostName when available. 
This will be usefull for topology grouping and ResourceLinks

image

### DIFF
--- a/pkg/pipeline/transform/kubernetes/kubernetes.go
+++ b/pkg/pipeline/transform/kubernetes/kubernetes.go
@@ -83,6 +83,7 @@ func (k *KubeData) GetInfo(ip string) (*Info, error) {
 					Labels:          pod.Labels,
 					OwnerReferences: pod.OwnerReferences,
 					HostIP:          pod.Status.HostIP,
+					HostName:        k.getHostName(pod.Status.HostIP),
 				}
 			case typeNode:
 				node := objs[0].(*v1.Node)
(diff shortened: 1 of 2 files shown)


pkg/pipeline/transform/kubernetes/kubernetes.go
```diff
@@ -102,6 +103,7 @@ func (k *KubeData) GetInfo(ip string) (*Info, error) {
 			}
 
 			info.Owner = k.getOwner(info)
+			info.HostName = k.getHostName(info)
```
eranra, line 106: You can move this under the `typePod` case because you output only if `HostIP` is not empty --- here: https://github.com/netobserv/flowlogs-pipeline/pull/196/files#diff-77b720060ce5b2e61d001e832eac1bf4d94ffdeb9734bbd4ee678d9398ad3892R155 

And this is set only for pods

eranra (approved):

pkg/pipeline/transform/kubernetes/kubernetes.go
jpinsonneau: Yes I was wondering if we could do the same for `services` but it can run on multiple pods / hosts and we'll need to manage arrays.

Let's keep it simple for now, thanks for the feedback :+1:

PR #198, opened 2022-05-06
PoC with generic pipes

mariomac:
From my hack'n'hustle day: this is just an incomplete PoC/demo about replacing the current pipeline builder by the upper-upperstream pipes library, in order to remove some code and allow easier extension of the current pipeline stages.


pkg/config/config.go
```diff
@@ -19,106 +19,68 @@ package config
...
+	Filter  []api.TransformFilter
+	Network []api.TransformNetwork
+	// Extractors
+	Aggregates []api.AggregateDefinition
+	// Encoders
+	Prom        []api.PromEncode
+	KafkaEncode []api.EncodeKafka
+	// Writers
+	Loki   []api.WriteLoki
+	Stdout []api.WriteStdout
```
mariomac, line 64: In the current PoC, the current YAML configuration format would be slightly different. You wouldn't need anymore to specify the "type" parameter and some structures disappear.

Off course this could be reworked to avoid breaking changes.

pkg/pipeline/pipeline.go
```diff
@@ -21,58 +21,53 @@ import (
...
-	log.Debugf("configParams = %v ", configParams)
+	graph.RegisterMiddle(builder, decode.AwsProvider)
+	graph.RegisterMiddle(builder, decode.JSONProvider)
+	graph.RegisterMiddle(builder, decode.ProtobufProvider)
 
-	build := newBuilder(configParams, stages)
-	if err := build.readStages(); err != nil {
+	/** etc **/
+
+	graph, err := builder.Build(options.PipeLine)
```
mariomac, line 55: the whole pipeline builder is replaced by this... we should add more `RegisterMiddle` and `RegisterTerminal` for the rest of stages.

The encoder/decoders could be replaced by `RegisterCodec` invocations, so you don't need to explicitly instantiate them in the configuration.

The `builder.Build` method magically instantiates and connects all the nodes according to the passed pipeline configuration.

pkg/api/decode_aws.go
```diff
@@ -17,6 +17,9 @@
 
 package api
 
+import "github.com/mariomac/pipes/pkg/graph/stage"
+
 type DecodeAws struct {
+	stage.Instance
```
mariomac, line 23: https://github.com/mariomac/pipes/tree/main/examples/tutorial/02-highlevel-nodes#defining-providers-for-each-node

pkg/config/config.go
```diff
@@ -19,106 +19,68 @@ package config
...
-}
-
-type Ingest struct {
-	Type      string
-	File      File
-	Collector api.IngestCollector
-	Kafka     api.IngestKafka
-	GRPC      api.IngestGRPCProto
+type PipelineDefinition struct {
+	graph.Connector
```
mariomac, line 43: https://github.com/mariomac/pipes/tree/main/examples/tutorial/02-highlevel-nodes#defining-the-graph-configuration

eranra:
@mariomac does this also provide us value in the vision to make the pipeline more dynamic and allow the plug-ins approach?

eranra:
Hi @mariomac, I just realized after talking with @ronensc and @KalmanMeth that not all of the pipeline stages are working with channels (a.k.a.  github.com/mariomac/pipes ) + we had some issues concurrency issues with FLP (i.e. @jotak ). It will be good to get all stages into channels ASAP and not drag forward, right? ... is this connected to this PR??? (if not maybe need to put this comment somewhere else).

PR #201, opened 2022-05-11
Connection tracking - hash mechanism

ronensc:
This is the first PR of the new Connection tracking module. Only the hash mechanism is implemented here. many more PRs are to come...


pkg/pipeline/conntrack/hash.go
```diff
@@ -0,0 +1,87 @@
+/*
```
KalmanMeth, line 1: For the directory structure, should conntrack be directly under pipeline or under transform, like the other network transforms?

pkg/pipeline/conntrack/hash.go
```diff
@@ -0,0 +1,87 @@
...
+		h, err := computeHashFields(flowLog, fg.Fields)
+		if err != nil {
+			return nil, fmt.Errorf("compute hash: %w", err)
+		}
+		fieldGroup2hash[fg.Name] = h
+	}
+
+	// Compute the total hash
+	hash := fnv.New32a()
+	for _, fgName := range keyFields.Hash.FieldGroups {
```
KalmanMeth, line 47: Can you explain somewhere why we need different FieldGroups in KeyFields and in Hash? What is the purpose of these additional FieldGroups?

KalmanMeth: One is of type []FieldGroup and one is of type []string. Using the same name for these 2 different fields is a bit confusing. I assume that the []string is derived from Name field of the []FieldGroup.

pkg/api/conn_track.go
```diff
@@ -0,0 +1,47 @@
...
+
+type FieldGroup struct {
+	Name   string   `yaml:"name" doc:"field group name"`
+	Fields []string `yaml:"fields" doc:"list of fields in the group"`
+}
+
+type ConnTrackHash struct {
+	FieldGroups []string `yaml:"fieldGroups" doc:"list of field groups"`
+	FieldGroupA string   `yaml:"fieldGroupA" doc:"field group A"`
+	FieldGroupB string   `yaml:"fieldGroupB" doc:"field group B"`
```
KalmanMeth, line 39: Do we need a description somewhere of what is meant by A and B?

pkg/pipeline/conntrack/hash_test.go
```diff
@@ -0,0 +1,212 @@
...
+			"Same IP, port and protocol",
+			NewFlowLog(ipA, portA, ipB, portB, protocolA, 111, 22),
+			NewFlowLog(ipA, portA, ipB, portB, protocolA, 222, 11),
+			true,
+		},
+		{
+			"Alternating ip+port",
+			NewFlowLog(ipA, portA, ipB, portB, protocolA, 111, 22),
+			NewFlowLog(ipB, portB, ipA, portA, protocolA, 222, 11),
+			false,
```
KalmanMeth, line 90: Why is this false? Won't we get the same hash for the 2 entries?

pkg/pipeline/conntrack/hash_test.go
```diff
@@ -0,0 +1,212 @@
...
+			{
+				Name: "protocol",
+				Fields: []string{
+					"Proto",
+				},
+			},
+		},
+		Hash: api.ConnTrackHash{
+			FieldGroups: []string{"protocol"},
+			FieldGroupA: "src",
```
KalmanMeth, line 152: The use of FieldGroupA field (non-blank) should be explained somewhere.

pkg/pipeline/conntrack/hash.go
ronensc: I don't think of the connection tracking as a network transform, but as a new pipeline stage.
It will emit new types of records (connection related) in addition to the flow log records.

pkg/pipeline/conntrack/hash.go
ronensc: You are correct.
`KeyFields.FieldGroups` is where you define the field groups
and
`KeyFields.Hash.FieldGroups` is where you reference the defined field groups (by name). I'll rename it and add some comments to make it a bit clearer.


Please take a look at these configuration examples:
https://gist.github.com/ronensc/0136cd1c3a761e2d6067679362fa72c6
https://gist.github.com/ronensc/c6d0a07b0503bbd7a36bc4ef0b6b5b29

pkg/api/conn_track.go
ronensc: I'll add a comment to describe their meaning

pkg/pipeline/conntrack/hash.go
```diff
@@ -0,0 +1,88 @@
...
+	"hash/fnv"
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/api"
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+)
+
+// ComputeHash computes the hash of a flow log according to keyFields.
+// Two flow logs will have the same hash if they belong to the same connection.
+func ComputeHash(flowLog config.GenericMap, keyFields api.KeyFields) ([]byte, error) {
+	type hashType []byte
```
eranra, line 33: @ronensc If you externalize the  `hashType` type you will be able to also return `hashType` and not []byte

pkg/api/conn_track.go
```diff
@@ -0,0 +1,53 @@
...
+
+package api
+
+type ConnTrack struct {
+	KeyFields         KeyFields     `yaml:"keyFields" doc:"fields that are used to identify the connection"`
+	OutputRecordTypes []string      `yaml:"outputRecordTypes" doc:"output record types to emit"`
+	OutputFields      []OutputField `yaml:"outputFields" doc:"list of output fields"`
+}
+
+type KeyFields struct {
```
eranra, line 26: I advice changing the name to `KeyDefinition` or `HashDefinition` because the term fields is already used inside `FieldGroup` and this defines more than just the `fields`

pkg/pipeline/conntrack/hash_test.go
```diff
@@ -0,0 +1,212 @@
...
+			require.NoError(t, err1)
+			require.NoError(t, err2)
+			if test.sameHash {
+				require.Equal(t, h1, h2)
+			} else {
+				require.NotEqual(t, h1, h2)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
eranra: What are we going to do with interfaces? if we want to see the flow only once, and we do not care from which interface we got it ???

pkg/pipeline/conntrack/hash.go
```diff
@@ -0,0 +1,88 @@
...
+	// Compute the total hash
+	hash := fnv.New32a()
+	for _, fgName := range keyFields.Hash.FieldGroupRefs {
+		hash.Write(fieldGroup2hash[fgName])
+	}
+	if keyFields.Hash.FieldGroupARef != "" {
+		hashA := fieldGroup2hash[keyFields.Hash.FieldGroupARef]
+		hashB := fieldGroup2hash[keyFields.Hash.FieldGroupBRef]
+		// Determine order between A's and B's hash to get the same hash for both flow logs from A to B and from B to A.
+		if bytes.Compare(hashA, hashB) < 0 {
```
eranra, line 54: nice :-)

pkg/pipeline/conntrack/hash.go
```diff
@@ -0,0 +1,88 @@
...
+			hash.Write(hashA)
+		}
+	}
+	return hash.Sum([]byte{}), nil
+}
+
+func computeHashFields(flowLog config.GenericMap, fieldNames []string) ([]byte, error) {
+	h := fnv.New32a()
+	for _, fn := range fieldNames {
+		// TODO: How should we handle a missing fieldName?
```
eranra, line 68: Info log and skip ( for now) --- this will cause in different Hash anyway ... maybe later we will make that a parameter

pkg/pipeline/conntrack/hash.go
```diff
@@ -0,0 +1,88 @@
...
+	for _, fg := range keyFields.FieldGroups {
+		h, err := computeHashFields(flowLog, fg.Fields)
+		if err != nil {
+			return nil, fmt.Errorf("compute hash: %w", err)
+		}
+		fieldGroup2hash[fg.Name] = h
+	}
+
+	// Compute the total hash
+	hash := fnv.New32a()
```
eranra, line 46: Can we "trick" somehow (maybe interface?) so that it will be a parameter to the functions what length of Hash to use >??? 

@mariomac maybe you have an idea ??? ^^^

eranra (approved):
Look good to me ... more than good actually :-) I added some comments but nothing really prevents merging.,

pkg/pipeline/conntrack/hash.go
mariomac: Yeah, actually the `fnv.New32a` and rest of hashers implement the `io.Writer` interface. You can change the function signature to something like:

```go
func computeHashFields(flowLog config.GenericMap, fieldNames []string, hasher io.Writer) ([]byte, error) {
```

and invoke it like:

```go
h, err := computeHashFields(flowLog, fg.Fields, fnv.New32a())
```

pkg/pipeline/conntrack/hash_test.go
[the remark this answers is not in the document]
ronensc: That's still an open question. In my current mental model, if we get multiple flow logs for the same flow from different interfaces, we can either add the interface field into the hash and get multiple connections or do not add it and aggregate the flow logs in the connection (which might result in wrong aggregation values)

pkg/pipeline/conntrack/hash.go
ronensc: 👍

pkg/pipeline/conntrack/hash.go
```diff
@@ -65,8 +66,11 @@ func ComputeHash(flowLog config.GenericMap, keyFields api.KeyFields) ([]byte, er
 func computeHashFields(flowLog config.GenericMap, fieldNames []string) ([]byte, error) {
 	h := fnv.New32a()
 	for _, fn := range fieldNames {
-		// TODO: How should we handle a missing fieldName?
-		f := flowLog[fn]
+		f, ok := flowLog[fn]
+		if !ok {
+			log.Warningf("Missing field %v", fn)
```
stleerh, line 72: This could be risky.  It could end up flooding the log file.  Is there a way to avoid repeated logs?

ronensc: We can replace it with a prometheus metric. WDYT?

pkg/pipeline/conntrack/hash_test.go
```diff
@@ -0,0 +1,215 @@
...
+				Fields: []string{
+					"SrcAddr",
+					"SrcPort",
+				},
+			},
+			{
+				Name: "dst",
+				Fields: []string{
+					"DstAddr",
+					"DstPort",
```
stleerh, line 57: If you include the ephemeral port in the hash, it will count a lot of connections.  We need to get an agreement on what a "connection" is.  Example: Let's say you access a web page.  Is that one connection?  With this implementation, it could be anywhere from 1 to 6 connections.  
  
A typical web page will refer to JavaScript files, CSS files, images, etc.  The browser will need to fetch these files.  Pretty much all modern browsers today will allow up to 6 simultaneous connections to the same domain, and will reuse these connections to fetch all the files.  Each of these connections will have a different ephemeral port.

pkg/pipeline/conntrack/hash.go
stleerh: I don't believe this needs to be tracked as a metric.  It doesn't seem like it's likely to occur but can if someone wants to be malicious.  For now, maybe let it go, but we should have a general solution to avoid spamming the log file.

pkg/pipeline/conntrack/hash_test.go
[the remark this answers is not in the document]
eranra: @ronensc so let's merge this as-is and think on that later. There are many options, and we will need to start with something that works. Let's plan to add the interface to the bi-directional hash. So this will get us separate flows for each of the interfaces. I think that for now, this is good. We can always decide to hook to only a sub-set of the interface if needed.

From my PoV there is not need to change this PR for this point ... if needed we will do that later.

pkg/pipeline/conntrack/hash_test.go
ronensc: I see. I tried to make the decision of what is a connection configurable. But it's still not flexible enough to support this use-case.

This specific unit test configures the classic 5-tuple to distinguish between connections. The other unit test uses a slightly different configuration. It uses the same 5-tuple but includes both flow directions in the same connection.

KalmanMeth (approved):

PR #203, opened 2022-05-13
[breaking change] Fix protobuf decode for new time fields in ms

jotak:


mariomac (approved):
LGTM. This is a breaking change that I don't know whether we should document. I guess people using version 0.1.2 of the NOO will still deploy the older flowlogs-pipeline version.

jotak:
Yes, true, actually this was breaking changes also in the console plugin and in the operator, as they all rely now on a new field that didn't exist before.
We must not forget to mention it in the next release note. Maybe we can use github labels to flag breaking changes...

PR #204, opened 2022-05-16
Common cache with expiry time

KalmanMeth:
Refactor multiple instances of LRU cache into shared code.

### DIFF
--- a/pkg/pipeline/encode/encode_prom_test.go
+++ b/pkg/pipeline/encode/encode_prom_test.go
@@ -124,7 +124,7 @@ func Test_NewEncodeProm(t *testing.T) {
 	require.Equal(t, gEntryInfo1["value"], int(bytesA))
 
 	// verify entries are in cache; one for the gauge and one for the counter
-	entriesMap := encodeProm.mCache.LruCacheMap
+	entriesMap := encodeProm.mCache.CacheMap
 	require.Equal(t, 2, len(entriesMap))
 
 	eInfo := entrySignature{
@@ -134,14 +134,14 @@ func Test_NewEncodeProm(t *testing.T) {
 
 	eInfoBytes := generateCacheKey(&eInfo)
 	encodeProm.mCache.Mu.Lock()
-	_, found := encodeProm.mCache.LruCacheMap[string(eInfoBytes)]
+	_, found := encodeProm.mCache.CacheMap[string(eInfoBytes)]
 	encodeProm.mCache.Mu.Unlock()
 	require.Equal(t, true, found)
 
 	// wait a couple seconds so that the entry will expire
 	time.Sleep(2 * time.Second)
 	encodeProm.mCache.CleanupExpiredEntries(encodeProm.expiryTime, encodeProm)
-	entriesMap = encodeProm.mCache.LruCacheMap
+	entriesMap = encodeProm.mCache.CacheMap
 	encodeProm.mCache.Mu.Lock()
 	require.Equal(t, 0, len(entriesMap))
 	encodeProm.mCache.Mu.Unlock()
@@ -171,9 +171,9 @@ func Test_EncodeAggregate(t *testing.T) {
 				labelNames: []string{"by", "aggregate"},
 			},
 		},
-		mCache: &utils.TimedLruCache{
-			LruCacheList: list.New(),
-			LruCacheMap:  make(utils.LruCacheMap),
+		mCache: &utils.TimedCache{
+			CacheList: list.New(),
+			CacheMap:  make(utils.TimedCacheMap),
 		},
 	}
 
--- a/pkg/pipeline/encode/encode_prom_test.go
+++ b/pkg/pipeline/encode/encode_prom_test.go
@@ -18,7 +18,6 @@
 package encode
 
 import (
-	"container/list"
 	"testing"
 	"time"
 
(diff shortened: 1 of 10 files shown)


pkg/pipeline/extract/aggregate/aggregate.go
```diff
@@ -47,7 +46,7 @@ type NormalizedValues string
 type Aggregate struct {
 	Definition api.AggregateDefinition
 	GroupsMap  map[NormalizedValues]*GroupState
-	GroupsList *list.List
+	Cache      *utils.TimedCache
```
ronensc, line 49: This change, replaced a linked list with a `TimedCache` (which is a linked list and a map). Can we remove `GroupsMap` and use the new cache instead?

pkg/pipeline/utils/timed_cache.go
```diff
@@ -0,0 +1,122 @@
...
+	e               *list.Element
+	SourceEntry     interface{}
+}
+
+type TimedCacheMap map[string]*CacheEntry
+
+type TimedCache struct {
+	Mu        sync.Mutex
+	CacheList *list.List
+	CacheMap  TimedCacheMap
```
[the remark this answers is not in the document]
KalmanMeth: CacheList is set and inspected by some cases.

pkg/pipeline/extract/aggregate/aggregate.go
KalmanMeth: The GroupsMap is used in several places in aggregate.go, and I think it would be messy to use the cache instead. If you see a clean way to do it, please share.

pkg/pipeline/encode/encode_prom_test.go
```diff
@@ -170,8 +171,10 @@ func Test_EncodeAggregate(t *testing.T) {
 				labelNames: []string{"by", "aggregate"},
 			},
 		},
-		mList:  list.New(),
-		mCache: make(metricCache),
+		mCache: &utils.TimedCache{
+			CacheList: list.New(),
+			CacheMap:  make(utils.TimedCacheMap),
+		},
```
ronensc, line 177: I think it's better to call `NewTimedCache()` instead.

pkg/pipeline/utils/timed_cache.go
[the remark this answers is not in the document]
ronensc: I found only one external access to `CacheList` and it's for querying its length. We can add a method for that in `timed_cache.go` and unexport the field.
https://github.com/netobserv/flowlogs-pipeline/blob/9783cf63f9545c32e1124f48a1e4faa74be5438e/pkg/pipeline/extract/aggregate/aggregate_test.go#L124

pkg/pipeline/extract/aggregate/aggregate.go
ronensc: I found 2 uses of `GroupsMap` in `aggregate.go`:
1. https://github.com/netobserv/flowlogs-pipeline/blob/9783cf63f9545c32e1124f48a1e4faa74be5438e/pkg/pipeline/extract/aggregate/aggregate.go#L138
I think it can be replaced with a call to `UpdateCacheEntry()`

2. https://github.com/netobserv/flowlogs-pipeline/blob/9783cf63f9545c32e1124f48a1e4faa74be5438e/pkg/pipeline/extract/aggregate/aggregate.go#L208
We need to add a new method to support iterating over the items in the cache.
I suggest adding a method like this:
```
func (l *TimedCache) Iterate(f func(key string, value interface{})) {
	for k, v := range l.CacheMap {
		f(k, v.SourceEntry)
	}
}
```

and replace the for loop with something like:
```
aggregate.Cache.Iterate(func(key string, value interface{}) {
	group := value.(GroupState)
	metrics = append(metrics, config.GenericMap{
	....
})
```

pkg/pipeline/utils/timed_cache.go
```diff
@@ -0,0 +1,121 @@
...
+
+type TimedCacheMap map[string]*cacheEntry
+
+type TimedCache struct {
+	Mu        sync.Mutex
+	CacheList *list.List
+	CacheMap  TimedCacheMap
+}
+
+func (l *TimedCache) GetCacheEntry(key string) (interface{}, bool) {
```
ronensc, line 50: Why the identifier of the receiver of all the methods of `*TimedCache` is `l`? (and not `t` or `tc`...)

pkg/pipeline/utils/timed_cache.go
```diff
@@ -77,25 +77,38 @@ func (l *TimedCache) UpdateCacheEntry(key string, entry interface{}) *cacheEntry
...
 	return cEntry
 }
 
+func (tc *TimedCache) GetCacheLen() int {
+	return tc.cacheList.Len()
+}
+
+// We expect that the function calling Iterate might make updates to the entries by calling UpdateCacheEntry()
+// We therefore cannot take the lock at this point since it will conflict with the call in UpdateCacheEntry()
+// TODO: Question: Do we need to make a copy of the map so that we are unaffected by changes to the original map while we iterate?
```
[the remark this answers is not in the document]
KalmanMeth: Added mutex to Iterate() with comment that in the future if needed to make updates to cache during interation then we need to implement the ability.

ronensc (approved):

PR #207, opened 2022-05-19
NETOBSERV-341 configure netflow v5 separately (breaking change if using v5)

### LINKED ISSUE #200: Make it possible to disable legacy ingester (netflow v5)
cf https://issues.redhat.com/browse/NETOBSERV-341
In the operator I configured the ingester with a given (single) port, assuming it would result in a single server listening. But it turns out it still listens for legacy netflow in addition to what I configured.
I think the configuration to listen for both v9/ipfix and v5 should me more explicit

jotak:
Fixes #200

@eranra @KalmanMeth 
Is it ok with you to have a breaking changes here? (people who used legacy v5 now have to configure it explicitly)
If that's a problem, we could default to the previous behaviour (legacy port = main port +1) and have another setting to disable v5, but I think it's more simple to keep default as 0 = legacy disabled.

### DIFF
--- a/contrib/kubernetes/flowlogs-pipeline.conf.yaml
+++ b/contrib/kubernetes/flowlogs-pipeline.conf.yaml
@@ -5,6 +5,7 @@ parameters:
     collector:
       hostname: 0.0.0.0
       port: 2055
+      portLegacy: 2056
     type: collector
   name: ingest_collector
 - decode:
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Flags:
 ```
 <!---END-AUTO-flowlogs-pipeline_help--->
 
-> Note: for API details refer to  [docs/api.md](docs/api.md).
+> Note: for API details refer to [docs/api.md](docs/api.md).
 > 
 ## Configuration generation
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -39,12 +39,13 @@ Following is the supported API format for kafka encode:
          batchSize: limit on how many messages will be buffered before being sent to a partition
 </pre>
 ## Ingest collector API
-Following is the supported API format for the netflow collector:
+Following is the supported API format for the NetFlow / IPFIX collector:
 
 <pre>
  collector:
          hostName: the hostname to listen on
-         port: the port number to listen on
+         port: the port number to listen on, for IPFIX/NetFlow v9. Omit or set to 0 to disable IPFIX/NetFlow v9 ingestion
+         portLegacy: the port number to listen on, for legacy NetFlow v5. Omit or set to 0 to disable NetFlow v5 ingestion
          batchMaxLen: the number of accumulated flows before being forwarded for processing
 </pre>
 ## Ingest Kafka API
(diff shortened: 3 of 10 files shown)


eranra:
@jotak I reviews the PR and it looks clean to me and can be merged. I think that this is better to be restrictive here, so totally ok with that.

BUT:: Merging this PR will break both FLP and NOO testing/deployment over kind::

Need to add the legacy port here: 

https://github.com/netobserv/flowlogs-pipeline/blob/main/contrib/kubernetes/flowlogs-pipeline.conf.yaml#L7

So that this will work with the Netflow simulator 

https://github.com/netobserv/flowlogs-pipeline/blob/main/contrib/kubernetes/deployment-netflow-simulator.yaml#L22

A similar thing will happen also to NOO:

In https://github.com/netobserv/network-observability-operator/blob/main/.mk/local.mk#L34 We will not expose port 2056 anymore so need to add that to the Operator code ???

eranra (approved):

jotak:
@eranra thanks, good catches. I've updated the confgen, see the second commit.
I'm not able to test on my machine (not enough RAM for kind), however, I wonder why the e2e tests passed in CI if it was relying on netflow v5 traffic.

About the netobserv operator, we can talk about that, but it's not supposed to support netflow v5, we only support ipfix, so I'd rather remove the v5 stuff that is used in make targets, rather than updating the code to handle non-supported v5

contrib/kubernetes/flowlogs-pipeline.conf.yaml
```diff
@@ -5,6 +5,7 @@ parameters:
     collector:
       hostname: 0.0.0.0
       port: 2055
+      portLegacy: 2056
```
eranra, line 8: @joel

This file is automatically generated by running `confgenerator`  from the Makefile `https://github.com/netobserv/flowlogs-pipeline/blob/main/Makefile#L243` 

The parameter needs to be added to https://github.com/netobserv/flowlogs-pipeline/blob/main/pkg/confgen/flowlogs2metrics_config.go#L56 

And then running the makefile should generate exactly this output 

BTW: It is using the same API structures:: look here:  https://github.com/netobserv/flowlogs-pipeline/blob/main/pkg/confgen/config.go#L41


Sorry, a bit more work :-(

eranra:
> @eranra thanks, good catches. I've updated the confgen, see the second commit. I'm not able to test on my machine (not enough RAM for kind), however, I wonder why the e2e tests passed in CI if it was relying on netflow v5 traffic.
> 
> About the netobserv operator, we can talk about that, but it's not supposed to support netflow v5, we only support ipfix, so I'd rather remove the v5 stuff that is used in make targets, rather than updating the code to handle non-supported v5

**Looks good, LGTM  ... :-)** 

As for the e2e tests ... I looked and we are not really using the simulator in the tests ... just in https://github.com/netobserv/flowogs-pipeline/blob/main/Makefile#L253 

so actually this does call for better e2e tests ... we might decide to do that just in the level of NOO and not in FLP ... need to decide.

jotak:
Thanks Eran

PR #208, opened 2022-05-22
 Add basic functionality of connection tracking

ronensc:
This PR adds the basic functionality of connection tracking.
Currently it outputs only new connection records and flow logs records.

Notes:
- There are still missing parts that will be added in later PRs such as adding the hash field in the output records,  reporting on connection status/end and deleting ended connections
- I struggled finding names for the identifiers (methods, variables, etc) and deciding what should be exported (PascalCase). Please advice on better names.
- ~~Once #201 is merged, this PR should be rebased.~~

### DIFF
--- a/pkg/pipeline/conntrack/aggregator.go
+++ b/pkg/pipeline/conntrack/aggregator.go
@@ -43,10 +43,10 @@ type aggregateBase struct {
 	initVal     float64
 }
 
-type aggregateSum struct{ aggregateBase }
-type aggregateCount struct{ aggregateBase }
-type aggregateMin struct{ aggregateBase }
-type aggregateMax struct{ aggregateBase }
+type aSum struct{ aggregateBase }
+type aCount struct{ aggregateBase }
+type aMin struct{ aggregateBase }
+type aMax struct{ aggregateBase }
 
 // TODO: think of adding a more complex operation such as Average Packet Size which involves 2 input fields: Bytes/Packets
 
@@ -66,16 +66,16 @@ func newAggregator(of api.OutputField) (aggregator, error) {
 	switch of.Operation {
 	case api.ConnTrackOperationName("Sum"):
 		aggBase.initVal = 0
-		agg = &aggregateSum{aggBase}
+		agg = &aSum{aggBase}
 	case api.ConnTrackOperationName("Count"):
 		aggBase.initVal = 0
-		agg = &aggregateCount{aggBase}
+		agg = &aCount{aggBase}
 	case api.ConnTrackOperationName("Min"):
 		aggBase.initVal = math.MaxFloat64
-		agg = &aggregateMin{aggBase}
+		agg = &aMin{aggBase}
 	case api.ConnTrackOperationName("Max"):
 		aggBase.initVal = -math.MaxFloat64
-		agg = &aggregateMax{aggBase}
+		agg = &aMax{aggBase}
 	default:
 		return nil, fmt.Errorf("unknown operation: %q", of.Operation)
 	}
@@ -118,7 +118,7 @@ func (agg *aggregateBase) addField(conn connection) {
 	}
 }
 
-func (agg *aggregateSum) update(conn connection, flowLog config.GenericMap, d direction) {
+func (agg *aSum) update(conn connection, flowLog config.GenericMap, d direction) {
 	outputField := agg.getOutputField(d)
 	v, err := agg.getInputFieldValue(flowLog)
 	if err != nil {
@@ -130,14 +130,14 @@ func (agg *aggregateSum) update(conn connection, flowLog config.GenericMap, d di
 	})
 }
 
-func (agg *aggregateCount) update(conn connection, flowLog config.GenericMap, d direction) {
+func (agg *aCount) update(conn connection, flowLog config.GenericMap, d direction) {
 	outputField := agg.getOutputField(d)
 	conn.updateAggValue(outputField, func(curr float64) float64 {
 		return curr + 1
 	})
 }
 
-func (agg *aggregateMin) update(conn connection, flowLog config.GenericMap, d direction) {
+func (agg *aMin) update(conn connection, flowLog config.GenericMap, d direction) {
 	outputField := agg.getOutputField(d)
 	v, err := agg.getInputFieldValue(flowLog)
 	if err != nil {
@@ -150,7 +150,7 @@ func (agg *aggregateMin) update(conn connection, flowLog config.GenericMap, d di
 	})
 }
 
-func (agg *aggregateMax) update(conn connection, flowLog config.GenericMap, d direction) {
+func (agg *aMax) update(conn connection, flowLog config.GenericMap, d direction) {
 	outputField := agg.getOutputField(d)
 	v, err := agg.getInputFieldValue(flowLog)
 	if err != nil {
--- a/pkg/pipeline/conntrack/aggregator.go
+++ b/pkg/pipeline/conntrack/aggregator.go
@@ -1,0 +1,164 @@
+/*
+ * Copyright (C) 2022 IBM, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package conntrack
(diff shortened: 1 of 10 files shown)


pkg/pipeline/conntrack/aggregator_test.go
```diff
@@ -0,0 +1,152 @@
...
+	require.NotNil(t, err)
+
+	// unknown Operation
+	_, err = newAggregator(api.OutputField{
+		Name:      "MyAgg",
+		Operation: "unknown",
+		SplitAB:   true,
+		Input:     "Input",
+	})
+}
```
KalmanMeth, line 47: missing `require.NotNil(t, err)`

ronensc: thanks

pkg/pipeline/conntrack/aggregator_test.go
```diff
@@ -0,0 +1,152 @@
...
+		},
+		{
+			name:        "Operation max",
+			outputField: api.OutputField{Name: "MyAgg", Operation: "max"},
+			expected:    &aggregateMax{aggregateBase{"MyAgg", "MyAgg", false, -math.MaxFloat64}},
+		},
+		{
+			name:        "Operation min",
+			outputField: api.OutputField{Name: "MyAgg", Operation: "min"},
+			expected:    &aggregateMin{aggregateBase{"MyAgg", "MyAgg", false, math.MaxFloat64}},
```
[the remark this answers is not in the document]
ronensc: The previous test cases use `sum`, but I'll add another one with an explicit name for that.

pkg/pipeline/conntrack/aggregator.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+type aggregateBase struct {
+	inputField  string
+	outputField string
+	splitAB     bool
+	initVal     float64
+}
+
+type aggregateSum struct{ aggregateBase }
```
eranra, line 46: you can drop the  `aggregate` prefix in all those types ... if those are reserved words you can start with `a` or `_` ... no read to repeat the package name

pkg/pipeline/conntrack/conn.go
```diff
@@ -0,0 +1,112 @@
+/*
```
eranra, line 1: consider renaming the file from conn.go to conntrack.go

pkg/pipeline/conntrack/conntrack_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+func TestTrack(t *testing.T) {
+
+	ipA := "10.0.0.1"
+	ipB := "10.0.0.2"
+	portA := 9001
+	portB := 9002
+	protocol := 6
+
+	flAB1 := NewFlowLog(ipA, portA, ipB, portB, protocol, 111, 11)
```
eranra, line 113: The NewFlowLog seems much more generic function than n the scope of hash/conntrtack consider moving that to test utils.

pkg/pipeline/conntrack/conntrack_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"testing"
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/api"
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+	"github.com/stretchr/testify/require"
+)
+
+func buildConnTrackConfig(isBidirectional bool, outputRecordType []string) *api.ConnTrack {
+	splitAB := isBidirectional
+	var h api.ConnTrackHash
```
eranra, line 30: s/h/hash

pkg/pipeline/conntrack/conntrack_test.go
eranra: Also it will be good to add the term mock in the name of the function

pkg/pipeline/conntrack/conntrack_test.go
```diff
@@ -0,0 +1,175 @@
...
+			buildConnTrackConfig(false, []string{"newConnection"}),
+			[]config.GenericMap{flAB1, flAB2, flBA3, flBA4},
+			[]config.GenericMap{
+				newMockRecordConn(ipA, portA, ipB, portB, protocol, 111, 11, 1),
+				newMockRecordConn(ipB, portB, ipA, portA, protocol, 333, 33, 1),
+			},
+		},
+		{
+			"unidirectional, output new connection and flow log",
+			buildConnTrackConfig(false, []string{"newConnection", "flowLog"}),
```
eranra, line 154: s/buildConnTrackConfig/buildMockConnTrackConfig

pkg/pipeline/conntrack/aggregator.go
```diff
@@ -0,0 +1,164 @@
...
+// newAggregator returns a new aggregator depending on the output field operation
+func newAggregator(of api.OutputField) (aggregator, error) {
+	if of.Name == "" {
+		return nil, fmt.Errorf("empty name %v", of)
+	}
+	var inputField string
+	if of.Input != "" {
+		inputField = of.Input
+	} else {
+		inputField = of.Name
```
eranra, line 62: @ronensc please make sure to document that if the name of the input field is not explicitly set we use the name of the output field also for input. This is very nice from usage point of view but might not be trivial to someone that uses this. Another option is to force explicit ... but I think that I agree with you here to make the configuration easier by reusing the name .... so from my PoV we just need good documentation here.

pkg/pipeline/conntrack/aggregator.go
```diff
@@ -0,0 +1,164 @@
...
+
+type aggregateSum struct{ aggregateBase }
+type aggregateCount struct{ aggregateBase }
+type aggregateMin struct{ aggregateBase }
+type aggregateMax struct{ aggregateBase }
+
+// TODO: think of adding a more complex operation such as Average Packet Size which involves 2 input fields: Bytes/Packets
+
+// newAggregator returns a new aggregator depending on the output field operation
+func newAggregator(of api.OutputField) (aggregator, error) {
```
eranra, line 54: @ronensc The code for `newAggregator` and for aggregator, in general, can be reused also for extract aggregate .... no in this PR but maybe in following ??? do you want to open an issue so we will remember to look on that?

pkg/pipeline/conntrack/conntrack.go
```diff
@@ -0,0 +1,149 @@
...
+func (ct *conntrackImpl) Track(flowLogs []config.GenericMap) []config.GenericMap {
+	log.Debugf("Entering Track")
+	log.Debugf("Track none, in = %v", flowLogs)
+
+	var outputRecords []config.GenericMap
+	for _, fl := range flowLogs {
+		// TODO: think of returning a string rather than []byte
+		computedHash, err := ComputeHash(fl, ct.config.KeyDefinition, ct.hasher)
+		if err != nil {
+			log.Warningf("skipping flow log: %v", err)
```
eranra, line 64: Can you add also the flowlog to the warning --- so we will be able to debug and understand what is broken

pkg/pipeline/conntrack/conntrack.go
```diff
@@ -0,0 +1,149 @@
...
+			continue
+		}
+		hashStr := hex.EncodeToString(computedHash.hashTotal)
+		conn, exists := ct.hash2conn[hashStr]
+		if !exists {
+			builder := NewConnBuilder()
+			conn = builder.
+				Hash(computedHash).
+				KeysFrom(fl, ct.config.KeyDefinition).
+				Aggregators(ct.aggregators).
```
eranra, line 74: Why do we need to keep the `aggregators` in each connection? we need just the fields right? the `aggregators` are the same for all the connections

eranra (approved):

pkg/pipeline/conntrack/conntrack_test.go
ronensc: done

pkg/pipeline/conntrack/conntrack_test.go
ronensc: Done.

pkg/pipeline/conntrack/conntrack_test.go
ronensc: done

pkg/pipeline/conntrack/aggregator.go
ronensc: done

pkg/pipeline/conntrack/aggregator.go
```diff
@@ -0,0 +1,164 @@
...
+	var inputField string
+	if of.Input != "" {
+		inputField = of.Input
+	} else {
+		inputField = of.Name
+	}
+	aggBase := aggregateBase{inputField: inputField, outputField: of.Name, splitAB: of.SplitAB}
+	var agg aggregator
+	switch of.Operation {
+	case api.ConnTrackOperationName("Sum"):
```
mariomac, line 67: What's the point of using this instead of:
```go
case "sum":
```
?

mariomac (approved):
LGTM, despite a minor question.

PR #209, opened 2022-05-23
Enum race

KalmanMeth:
Use separate enum cache tables for each of the different stages. This avoids the race condition we sometimes see in the tests without having to take a mutex.


ronensc (commented):
I think a simpler (and arguably cleaner) approach would be to stick with a single `enumNamesCache` and to initialize it with all the possible values in an `init()` function rather than populating it `lazily` as done currently in `GetEnumName()`

Issue #212, opened 2022-05-30
Avoid json encoding on flow ingestion


jotak:
As commented here https://github.com/netobserv/flowlogs-pipeline/blob/main/pkg/pipeline/ingest/ingest_collector.go#L168-L169, it seems like there's an easy perf improvement we can do by avoiding json encoding during flow ingestion, especially as it kinda forces users to use json decoding as the following stage.

It seems like we can "just" remove the json encode instruction, update related tests and voila. But that's a breaking change of course, consumers need to update their pipeline to remove the following json decode stage.

What I'm not sure, is if it's still necessary to allow json encoding, as all following stages use a native map format anyway. If we think we must maintain a json-encoded option, there's several possible approaches I can see so far:
- add a json-encode option as part of the ingest-collector stage params
- create a new json-encode stage (oddly, this stage should be a "transform" stage and not an "encode" stage, because transform has input & output, whereas "encode" has no output (??) )

ping @eranra @KalmanMeth @ronensc

jotak:
Fixed by #225

PR #215, opened 2022-06-02
Avoid json encoding on flow ingestion #212 (breaking change)

jotak:
I had to create a "fake" stage `DecodeCast`, just to fix incompatible types `map[string]interface{}` and `GenericMap` (which are actually the same).
However I think we could do better by merging all Ingest and Decode stages (reach out on slack to discuss)

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -25,6 +25,7 @@ const (
 	GRPCType              = "grpc"
 	KafkaType             = "kafka"
 	JSONType              = "json"
+	CastType              = "cast"
 	PBType                = "protobuf"
 	AWSType               = "aws"
 	StdoutType            = "stdout"
--- a/pkg/config/pipeline_builder.go
+++ b/pkg/config/pipeline_builder.go
@@ -83,6 +83,11 @@ func (b *PipelineBuilderStage) DecodeJSON(name string) PipelineBuilderStage {
 	return b.next(name, StageParam{Name: name, Decode: &Decode{Type: api.JSONType}})
 }
 
+// DecodeCast chains the current stage with a Cast decode stage and returns that new stage.
+func (b *PipelineBuilderStage) DecodeCast(name string) PipelineBuilderStage {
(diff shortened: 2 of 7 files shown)


mariomac (commented):
Mostly LGTM but I have few minor change suggestions.

I agree that it would be better to merge ingest and decode so we don't leak an implementation detail to our potential users (who should define this pipeline stage even if it does nothing visible to them).

In the future, we could also integrate automatic "codecs" that detect where two stages with different types are connected and transparently add this conversor when the stage types are convertible, so the definer of the pipeline shouldn't need to be aware of this details. In my hack and hustle days, I tried this idea and it worked: https://github.com/mariomac/pipes/tree/main/examples/tutorial/03-codecs

jotak:
> we could also integrate automatic "codecs" that detect where two stages with different types are connected and transparently add this conversor when the stage types are convertible

but would it work in this situation, where ingest output is `interface{}` and next input is `GenericMap` ? How could it infer the codec for `interface{}` ?

mariomac:
> but would it work in this situation, where ingest output is interface{} and next input is GenericMap ? How could it infer the codec for interface{} ? Doing reflection?

you should add a codec like:
```
func InterfaceMapCodec(in <-chan []interface{}, out chan<- []GenericMap) {
	for i := range in {
		out <- // your conversion code here
	}
}
```
The code to write from our side is the same, but from a user perspective, they shouldn't define any "decode" step.

mariomac:
Yes, yes, I think that the first option should be improve the current interface and remove the necessity of this intermediate step.

Off-topic: In the original library, since it uses generics, you wouldn't need to do that in most cases because e.g. for GRPC you wouldn't forward interfaces but actual `pfblow.Record`. instances.

jotak:
I created an alternative PR that I think is better: https://github.com/netobserv/flowlogs-pipeline/pull/221
The advantage of that alternative is that it hides the implementation details from the users.

jotak:
closing: superseded by #225

PR #219, opened 2022-06-02
create custom flp image with custom metric

KalmanMeth:
Perform the following operations:

make create-kind-cluster
make build-image-custom
make kind-load-image
make deploy
make deploy-prometheus
make deploy-netflow-simulator

The metrics defined in the directory `network_definitions_custom` will be run through the confgenerator and built into a container extracted from git.
The next step is to do this in the network-observability-operator repository and integrate it with the operator.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,6 +129,10 @@ run: build ## Run
 build-image:
 	DOCKER_BUILDKIT=1 $(OCI_RUNTIME) build -t $(DOCKER_IMG):$(DOCKER_TAG) -f contrib/docker/Dockerfile .
 
+.PHONY: build-image-custom
+build-image-custom:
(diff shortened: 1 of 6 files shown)


eranra:
@KalmanMeth this PR seems like a good practice for using different folders to include custom definitions but I do not think that we need that as part of the FLP project. In FLP we need to have just one set of example network definitions and actually all of the content of this PR should be probably be PR'ed into NOO

KalmanMeth:
> @KalmanMeth this PR seems like a good practice for using different folders to include custom definitions but I do not think that we need that as part of the FLP project. In FLP we need to have just one set of example network definitions and actually all of the content of this PR should be probably be PR'ed into NOO
Correct. This is the exercise to perform what is needed, but it needs to be integrated in NOO. I am still working on how to integrate it directly into the NOO. I expect this PR will be be closed once the details are worked out to integrate with NOO.

PR #220, opened 2022-06-02
conntrack: Implement connection end records

ronensc:
- Add a connection store that stores the connections sorted by last update time
- Add a dependency on github.com/benbjohnson/clock to allow mocking time

Notes for reviewer:
- Please start reviewing from the added tests
- Does `connectionStore` deserve a file of its own?

### DIFF
--- a/pkg/pipeline/conntrack/conntrack.go
+++ b/pkg/pipeline/conntrack/conntrack.go
@@ -18,11 +18,14 @@
 package conntrack
 
 import (
+	"container/list"
 	"encoding/hex"
 	"fmt"
 	"hash"
 	"hash/fnv"
+	"time"
 
+	"github.com/benbjohnson/clock"
 	"github.com/netobserv/flowlogs-pipeline/pkg/api"
 	"github.com/netobserv/flowlogs-pipeline/pkg/config"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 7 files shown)


pkg/pipeline/conntrack/conntrack.go
```diff
@@ -42,14 +45,80 @@ type ConnectionTracker interface {
...
+		connList:  list.New(),
+	}
+}
+
+//////////////////////////
+
+type conntrackImpl struct {
+	clock                     clock.Clock
+	config                    api.ConnTrack
+	hasher                    hash.Hash
```
mariomac, line 116: Some points:

- As you always use `fnv.New32a()` in both production and testing, maybe you don't need to inject it here and just instantiate it directly in `ComputeHash`.
- Instead of `hash.Hash`, I'd use the `hash.Hash32` or `hash.Hash64` interfaces, which also provide a `Sum32() uint32` or `Sum64() uint64` methods. Using them would avoid having to convert byte slices to strings to store/compare the hashes, and we wouldn't need the `areHashEqual` and `copyTotalHash` functions
- Reusing a single hasher instance (using the `Reset()` method) could corrupt the hashes if it is used concurrently. I'd avoid injecting it as a property, but if you still want to inject the hasher for other reasons, I'd change this attribute by:
```go
hasherProvider  func() hash.Hash32
```
(or `hash.Hash64`). This way, when you invoke e.g. `ComputeHash` you can do it like:
```go
 		computedHash, err := ComputeHash(fl, ct.config.KeyDefinition, ct.hasherProvider())
```
So each invocation of `ComputeHash` will use a fresh/unique instance of a hasher.

When you instantiate the `connTrackImpl`, you should set the value as:
```go
hasherProvider: fnv.New32a,
```

mariomac (changes requested):

ronensc (commented):
@mariomac thanks for the constructive feedback :heart:

pkg/pipeline/conntrack/conntrack.go
```diff
@@ -42,14 +45,80 @@ type ConnectionTracker interface {
...
+type processConnF func(connection) (shouldDelete, shouldStop bool)
+
+func (ct *connectionStore) addConnection(hashStr string, conn connection) {
+	_, ok := ct.getConnection(hashStr)
+	if ok {
+		log.Errorf("BUG. connection with hash %v already exists in store. %v", hashStr, conn)
+	}
+	e := ct.connList.PushBack(conn)
+	ct.hash2conn[hashStr] = e
+}
```
KalmanMeth, line 69: Will this be ported later to use the common timed-cache we recently implemented?

pkg/pipeline/conntrack/conntrack_test.go
```diff
@@ -166,10 +168,187 @@ func TestTrack(t *testing.T) {
...
+			nil,
+		},
+		{
+			"51s: end conn BA",
+			startTime.Add(51 * time.Second),
+			nil,
+			[]config.GenericMap{
+				newMockRecordConn(ipB, portB, ipA, portA, protocol, 777, 77, 2),
+			},
+		},
```
KalmanMeth, line 341: What identifying mark is there for connection entries that are completed? Where do you test for that?

pkg/pipeline/conntrack/conntrack.go
ronensc: I think so

pkg/pipeline/conntrack/conntrack_test.go
ronensc: Good points. There isn't a mark yet. I'm working on adding a field to indicate the record type and submit it in a separate PR.

pkg/pipeline/conntrack/conntrack.go
```diff
@@ -86,24 +154,45 @@ func (ct *conntrackImpl) Track(flowLogs []config.GenericMap) []config.GenericMap
...
+	}
+
 	return outputRecords
 }
 
-func (ct *conntrackImpl) addConnection(hashStr string, conn connection) {
-	ct.hash2conn[hashStr] = conn
+func (ct *conntrackImpl) popEndConnections() []config.GenericMap {
+	var outputRecords []config.GenericMap
+	ct.connStore.iterateOldToNew(func(conn connection) (shouldDelete, shouldStop bool) {
```
eranra, line 168: This need to be async based on the timeout of the connection and not "just" based on the assumption that we get once in a while a new flow-log that will trigger.

@mariomac is this doable? do we need to use your channel PR for this to happen ???

PR #221, opened 2022-06-03
New IngestDecode hidden stage (breaking change)

jotak:
A new IngestDecode stage is added, that produces GenericMap output.
This stage is almost transparent to the users, as they'll continue to use
"ingest" like a usual ingester, but without a following "decode" stage. Impacted ingesters are
Collector and GRPC (which always decodes protobuf)

### DIFF
--- a/pkg/pipeline/ingest_decode/ingest_collector.go
+++ b/pkg/pipeline/ingest_decode/ingest_collector.go
@@ -107,8 +107,8 @@ func (w *TransportWrapper) Send(_, data []byte) error {
 	return err
 }
 
-// Ingest ingests entries from a network collector using goflow2 library (https://github.com/netsampler/goflow2)
-func (ingestC *ingestCollector) Ingest(out chan<- []config.GenericMap) {
+// IngestDecode ingests entries from a network collector using goflow2 library (https://github.com/netsampler/goflow2)
+func (ingestC *ingestCollector) IngestDecode(out chan<- []config.GenericMap) {
 	ctx := context.Background()
 	ingestC.in = make(chan map[string]interface{}, channelSize)
 
--- a/pkg/pipeline/ingest_decode/ingest_collector.go
+++ b/pkg/pipeline/ingest_decode/ingest_collector.go
@@ -15,7 +15,7 @@
  *
  */
 
-package ingest
+package ingest_decode
 
 import (
 	"context"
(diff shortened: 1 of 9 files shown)


mariomac (commented):
good improvement. Thanks!

mariomac (approved):

pkg/pipeline/ingest_decode/ingest_collector.go
```diff
@@ -109,7 +108,7 @@ func (w *TransportWrapper) Send(_, data []byte) error {
 }
 
 // Ingest ingests entries from a network collector using goflow2 library (https://github.com/netsampler/goflow2)
-func (ingestC *ingestCollector) Ingest(out chan<- []interface{}) {
+func (ingestC *ingestCollector) Ingest(out chan<- []config.GenericMap) {
 	ctx := context.Background()
```
KalmanMeth, line 112: Should we change the name of the interface function (Ingest) so we don't confuse it with an Ingestor? Or do we now dispose with the distinction between stage names, and any first stage is called Ingest?

pkg/pipeline/ingest_decode/ingest_decode.go
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ *
+ */
+
+package ingest_decode
+
+import "github.com/netobserv/flowlogs-pipeline/pkg/config"
+
+type IngesterDecoder interface {
+	Ingest(out chan<- []config.GenericMap)
```
KalmanMeth, line 23: Perhaps IngestDecode().

KalmanMeth (approved):

pkg/pipeline/ingest_decode/ingest_decode.go
jotak: done

pkg/pipeline/ingest_decode/ingest_collector.go
jotak: sure, done

pkg/pipeline/ingest_decode/ingest_collector.go
```diff
@@ -109,7 +108,7 @@ func (w *TransportWrapper) Send(_, data []byte) error {
 }
 
 // Ingest ingests entries from a network collector using goflow2 library (https://github.com/netsampler/goflow2)
```
eranra, line 110: s/Ingest/IngestDecode

jotak:
closing: superseded by #225

PR #223, opened 2022-06-03
(WIP) NETOBSERV-333: GRPC ingest metrics

mariomac:
Can't be merged until we merge this PR: https://github.com/netobserv/netobserv-ebpf-agent/pull/31

Implements the following metrics that are already implemented by the IPFIX ingestor:

```
flow_decoder_count
flow_process_nf_delay_summary_seconds
flow_process_nf_delay_summary_seconds_sum
flow_process_nf_delay_summary_seconds_count
flow_traffic_summary_size_bytes
flow_traffic_summary_size_bytes_sum
flow_traffic_summary_size_bytes_count
ingest_collector_flow_logs_processed
```

It does not implement the following metrics:

* `flow_traffic_bytes`, because it's equivalent to existing `flow_traffic_summary_size_bytes_sum`.
* `flow_process_nf_count`, because it's equivalent to existing `ingest_collector_flow_logs_processed`.
* `flow_process_nf_flowset_sum` and `flow_process_nf_flowset_records_sum` because the GRPC+Protobuf service does not have different message types as IPFIX (DataFlowSet, OptionsDataFlowSet, TemplateFlowSet, etc...). All the values here would be equivalent to `ingest_collector_flow_logs_processed`.
* `flow_process_nf_templates_count` because GRPC+Protobuf does not use templates.
* `flow_traffic_packets_counter` because GRPC does not work at packet level and there is no affordable way to inspect GRPC communication at IP/TCP level.
* `flow_summary_decoding_time` because GRPC+Protobuf does not provide an affordable way to inspect the raw-->object decoding time.
* `ingest_collector_queue_length` value will be always 0, since GRPC ingestor does not implement any internal queue to batch/forward the flow records.

Also, the `flow_decoder_count` might have very different values as for the IPFIX scenario. This metric specifies how many messages have been decoded. In IPFIX, its value would be the same as `ingest_collector_flow_logs_processed`, as Goflow2 decodes each flow record individually. In GRPC+Protobuf (eBPF agent), this value would be much lower because flow records are sent in big batches.


memodi:
@mariomac - is it possible to include `flow_process_nf_errors_count ` metric as well?

mariomac:
@memodi adding it

go.mod
```diff
@@ -25,39 +25,51 @@ require (
...
-	k8s.io/api v0.23.4
-	k8s.io/apimachinery v0.23.4
-	k8s.io/client-go v0.23.4
+	k8s.io/api v0.24.0
+	k8s.io/apimachinery v0.24.0
+	k8s.io/client-go v0.24.0
 	sigs.k8s.io/e2e-framework v0.0.6
 )
 
+replace github.com/netobserv/netobserv-ebpf-agent v0.1.0 => github.com/mariomac/netobserv-agent v0.0.0-20220602132418-53e4a4d79cd6
```
mariomac, line 38: This will be reverted back when the related PR is merged.

pkg/pipeline/ingest/ingest_grpc.go
```diff
@@ -57,3 +77,58 @@ func (no *GRPCProtobuf) Close() error {
...
+				if portIdx := strings.IndexByte(auth[0], ':'); portIdx > 0 {
+					remoteIP = auth[0][:portIdx]
+				} else {
+					remoteIP = auth[0]
+				}
+			}
+		}
+		trafficLabels := prometheus.Labels{
+			"type":       "protobuf",
+			"remote_ip":  remoteIP,
```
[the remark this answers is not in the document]
mariomac: This does not refer to the IPs in the flow but the IP of the eBPF agents. The number of remote IPs should be small (one agent per node).

I added this field to replicate the behavior of the equivalent metric in the IPFIX ingestor.

pkg/pipeline/ingest/ingest_grpc.go
```diff
@@ -1,17 +1,36 @@
...
 )
 
 const defaultBufferLen = 100
 
+// Prometheus metrics describing the performance of the eBPF ingest
+var (
+	flowDecoderCount = flow.DecoderStats.With(
+		prometheus.Labels{"worker": "", "name": "protobuf"})
+	processDelaySummary = flow.NetFlowTimeStatsSum.With(
+		prometheus.Labels{"version": "protobuf1.0", "router": ""})
```
eranra, line 29: Maybe move `protobuf1.0` to const ???

pkg/pipeline/ingest/ingest_grpc.go
```diff
@@ -1,17 +1,36 @@
...
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/proto"
 )
 
 const defaultBufferLen = 100
 
+// Prometheus metrics describing the performance of the eBPF ingest
+var (
+	flowDecoderCount = flow.DecoderStats.With(
+		prometheus.Labels{"worker": "", "name": "protobuf"})
```
eranra, line 27: maybe move `protobuf` to const ... or change to `GRPC` instead

eranra (approved):
Minor comments. LGTM

PR #225, opened 2022-06-07
Remove Decoder stage (breaking change)

jotak:
Decoder is removed as a stage, and added wherever it makes sense as
Ingest config: relevant only for File ingester and Kafka ingester.

Also remove AWS decoder (after discussion)


pkg/pipeline/pipeline_builder_test.go
```diff
@@ -47,42 +45,46 @@ func TestConnectionVerification(t *testing.T) {
...
+- { follows: ingest1, name: transform1 }
+- { follows: ingest1, name: transform2 }
+- { follows: transform2, name: write1 }
 `,
-		failingNodeName: "decode1",
+		failingNodeName: "transform1",
 	}, {
 		description: "terminal node write1 does not have inputs",
 		config: baseConfig + `- transform:
     type: none
```
KalmanMeth, line 84: It looks like transform1 does not have outputs.

KalmanMeth (approved):

pkg/pipeline/pipeline_builder_test.go
jotak: here again I think you're looking at the wrong pipeline, it's the one defined below

ronensc (approved):
LGTM

PR #226, opened 2022-06-09
NETOBSERV-334 concurrent map iteration & write fix

jpinsonneau:
This fix `fatal error: concurrent map iteration and map write` when calling `stdout` and `loki` stages in parallel

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,8 +24,6 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-type GenericMap map[string]interface{}
-
 var (
 	Opt        = Options{}
 	PipeLine   []Stage
@@ -117,15 +115,4 @@ func ParseConfig() error {
 	}
 	logrus.Debugf("params = %v ", Parameters)
 	return nil
-}
-
-// Copy will create a flat copy of GenericMap
-func (m GenericMap) Copy() map[string]interface{} {
-	result := map[string]interface{}{}
-
-	for k, v := range m {
-		result[k] = v
-	}
-
-	return result
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -118,3 +118,14 @@ func ParseConfig() error {
 	logrus.Debugf("params = %v ", Parameters)
 	return nil
 }
+
+// Copy will create a flat copy of GenericMap
+func (m GenericMap) Copy() map[string]interface{} {
+	result := map[string]interface{}{}
+
+	for k, v := range m {
+		result[k] = v
+	}
+
+	return result
+}
--- a/pkg/pipeline/write/write_loki.go
+++ b/pkg/pipeline/write/write_loki.go
@@ -231,7 +231,8 @@ func (l *Loki) processRecords() {
 			log.Debugf("exiting writeLoki because of signal")
 			return
 		case record := <-l.in:
-			err := l.ProcessRecord(record)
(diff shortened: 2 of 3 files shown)


pkg/config/config.go
```diff
@@ -118,3 +118,14 @@ func ParseConfig() error {
...
+// Copy will create a flat copy of GenericMap
+func (m GenericMap) Copy() map[string]interface{} {
+	result := map[string]interface{}{}
+
+	for k, v := range m {
+		result[k] = v
+	}
+
+	return result
+}
```
jpinsonneau, line 131: I do a simple flat copy here for performances purpose but we can implement a deep copy if needed

mariomac (approved):

pkg/config/config.go
KalmanMeth: Should this function be here or in utils?

pkg/config/config.go
```diff
@@ -118,3 +118,14 @@ func ParseConfig() error {
 	logrus.Debugf("params = %v ", Parameters)
 	return nil
 }
+
+// Copy will create a flat copy of GenericMap
+func (m GenericMap) Copy() map[string]interface{} {
```
eranra, line 123: Should the input and output be both `GenericMap` ??? 

i.e. s/ `map[string]interface{}` / `GenericMap`

pkg/config/config.go
eranra: @KalmanMeth @jpinsonneau I had the same thought .... maybe the right thing to do is open a new GenericMap.go file

pkg/pipeline/write/write_loki.go
```diff
@@ -231,7 +231,8 @@ func (l *Loki) processRecords() {
 			log.Debugf("exiting writeLoki because of signal")
 			return
 		case record := <-l.in:
-			err := l.ProcessRecord(record)
+			// copy record before process to avoid alteration on parallel stages
+			err := l.ProcessRecord(record.Copy())
```
eranra, line 235: @jpinsonneau I suggest moving the copy process to be inside the `ProcessRecord` function --- the rule should be that we are not allowed to change the inputs --- and we need to maybe check also in other places that we are not doing that.

eranra (approved):
Some suggestions for improvement but from my PoV this can merge as-is. so LGTM

pkg/config/config.go
jpinsonneau: I can move this :+1: 

Does a new file `generic_map.go` in config folder is ok for both of you ? @KalmanMeth @eranra

pkg/config/config.go
jpinsonneau: sounds good :+1:

pkg/config/config.go
jpinsonneau: https://github.com/netobserv/flowlogs-pipeline/pull/226/commits/2648b82480b2f4a2515e3bee5055568afa7a47be#diff-2eff304122c33ebdc65982528ca9efc9fd8061e6c75fda0beffd1b3ce0cdc046

pkg/pipeline/write/write_loki.go
eranra: @KalmanMeth ... so should we add copy also there ??? do you want to open a PR for that?

KalmanMeth (approved):

ronensc (approved):

jpinsonneau:
Thank you all for the feedback :+1:

PR #228, opened 2022-06-09
Refactor hashType: []byte -> uint64

ronensc:
This PR changes the type under the hood of the hash from `[]byte` to `uint64`.

`uint64` simplifies the code since its easier to copy than `[]byte` and can serve as a key in a map (in contrast to `[]byte`)

This PR is a follow up on the 2nd point from this review: https://github.com/netobserv/flowlogs-pipeline/pull/220#discussion_r893270514

I also changed the hash to use 64 bits instead of 32 bits to reduce the chance of collisions (different connections with the same hash).
Assuming a perfect hash function with uniform distribution of the has values,
 for 32 bits we need only 9,300 connections to have 1% chance for collision.
But, for 64 bits we need ~61,000,000 connections to have 1% chance for collision.
I based this on the table from [Birthday attack](https://en.wikipedia.org/wiki/Birthday_attack). @kalman Am I correct with my analysis?

### DIFF
--- a/pkg/pipeline/conntrack/conn.go
+++ b/pkg/pipeline/conntrack/conn.go
@@ -83,7 +83,7 @@ func (c *connType) toGenericMap() config.GenericMap {
 }
 
 func (c *connType) getHash() totalHashType {
-	return copyTotalHash(c.hash)
+	return c.hash
 }
 
 // TODO: Should connBuilder get a file of its own?
@@ -100,8 +100,8 @@ func NewConnBuilder() *connBuilder {
 	}
 }
 
-func (cb *connBuilder) Hash(h *totalHashType) *connBuilder {
-	cb.conn.hash = copyTotalHash(*h)
+func (cb *connBuilder) Hash(h totalHashType) *connBuilder {
+	cb.conn.hash = h
 	return cb
 }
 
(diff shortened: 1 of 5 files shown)


pkg/pipeline/conntrack/conn.go
```diff
@@ -68,7 +68,7 @@ func (c *connType) toGenericMap() config.GenericMap {
 }
 
 func (c *connType) getHash() totalHashType {
-	return copyTotalHash(c.hash)
+	return c.hash
```
mariomac, line 71: Since this is a private type and attribute, maybe the `getHash()` method is not needed anymore and you can invoke directly `c.hash`

pkg/pipeline/conntrack/hash.go
```diff
@@ -97,11 +82,17 @@ func computeHashFields(flowLog config.GenericMap, fieldNames []string, hasher ha
...
 		}
 		hasher.Write(bytes)
 	}
-	return hasher.Sum([]byte{}), nil
+	return hasher.Sum64(), nil
+}
+
+func uint64ToBytes(data uint64) []byte {
+	b := make([]byte, 8)
+	binary.BigEndian.PutUint64(b, data)
```
mariomac, line 94: I guess the encoding order (little vs big endian) does not matter here, right?

mariomac (approved):
LGTM. Just a minor comment.

KalmanMeth (approved):

pkg/pipeline/conntrack/hash.go
ronensc: right.

PR #229, opened 2022-06-10
Add omitempty flag to stage config

OlivierCazade:
We recently started using the config API in the NOO to configure FLP.

Some attributes do not have the omitempty flag resulting in them getting converted in json with an empty string or 0 as value instead of using the internal default value of FLP.


ronensc (approved):

jotak (approved):
lgtm

PR #230, opened 2022-06-12
Conntrack: Add RecordType and HashId fields

ronensc:
This PR adds 2 fields to the output records of the connection tracking module:
1. connection hash id as a string in hex format (internally, it's stored as uint64 so it's size is 8 bytes)
2. RecordType with 3 value options: `NewConnection`, `EndConnection`, `FlowLog`.

- [x] This PR is based on top of #220 and should be merged after that one is merged.

### DIFF
--- a/pkg/api/conn_track.go
+++ b/pkg/api/conn_track.go
@@ -19,6 +19,8 @@ package api
 
 import "time"
 
+const HashIdFieldName = "HashId"
+
 type ConnTrack struct {
 	// TODO: should by a pointer instead?
 	KeyDefinition        KeyDefinition `yaml:"keyDefinition" doc:"fields that are used to identify the connection"`
--- a/pkg/api/conn_track.go
+++ b/pkg/api/conn_track.go
@@ -19,7 +19,10 @@ package api
 
 import "time"
 
-const HashIdFieldName = "HashId"
+const (
+	HashIdFieldName     = "HashId"
+	RecordTypeFieldName = "RecordType"
+)
 
 type ConnTrack struct {
 	// TODO: should by a pointer instead?
--- a/pkg/api/conn_track.go
+++ b/pkg/api/conn_track.go
@@ -20,8 +20,8 @@ package api
 import "time"
 
 const (
-	HashIdFieldName     = "HashId"
-	RecordTypeFieldName = "RecordType"
+	HashIdFieldName     = "NetObs-HashId"
+	RecordTypeFieldName = "NetObs-RecordType"
 )
 
 type ConnTrack struct {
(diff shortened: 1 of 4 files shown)


mariomac (approved):

pkg/api/conn_track.go
```diff
@@ -19,6 +19,11 @@ package api
 
 import "time"
 
+const (
+	HashIdFieldName     = "NetObs-HashId"
```
eranra, line 23: Please remove "NetObs"

pkg/pipeline/conntrack/conntrack.go
```diff
@@ -141,14 +142,20 @@ func (ct *conntrackImpl) Track(flowLogs []config.GenericMap) []config.GenericMap
 			ct.connStore.addConnection(computedHash.hashTotal, conn)
 			ct.updateConnection(conn, fl, computedHash)
 			if ct.shouldOutputNewConnection {
-				outputRecords = append(outputRecords, conn.toGenericMap())
+				record := conn.toGenericMap()
+				addHashField(record, computedHash.hashTotal)
```
eranra, line 146: @ronensc what happened to the famous "builder" methodology ( consider making the code a bit nicer with such  :-)   )

eranra (approved):

pkg/api/conn_track.go
ronensc: I talked to @eranra offline and he is not fond of the mention of "NetObs" in the prefix.
@mariomac What about using underscore "_" as the prefix? Similar practice is employed in elasticsearch.
```go
const (
	HashIdFieldName     = "_HashId"
	RecordTypeFieldName = "_RecordType"
)
```

pkg/pipeline/conntrack/conntrack.go
ronensc: I think the builder design pattern is a bit overkill in this use case

pkg/pipeline/conntrack/conntrack.go
```diff
@@ -141,14 +142,20 @@ func (ct *conntrackImpl) Track(flowLogs []config.GenericMap) []config.GenericMap
...
 			}
 		} else {
 			ct.updateConnection(conn, fl, computedHash)
 		}
 
 		if ct.shouldOutputFlowLogs {
-			outputRecords = append(outputRecords, fl)
+			record := fl.Copy()
+			addHashField(record, computedHash.hashTotal)
+			addTypeField(record, api.ConnTrackOutputRecordTypeName("FlowLog"))
```
KalmanMeth, line 157: Should the values of ConnTrackOutputRecordTypeName be made into an enum and included in the documentation?

PR #235, opened 2022-06-16
NETOBSERV-258: add TimeReceived in Kafka transformer

mariomac:
If a flow is directly ingested by the kafka transformer instance (e.g. the eBPF agent submits there directly), it needs to add the `TimeReceived` field as soon as it unmarshalls the flow data.

### DIFF
--- a/pkg/pipeline/decode/decode_json.go
+++ b/pkg/pipeline/decode/decode_json.go
@@ -19,6 +19,7 @@ package decode
 
 import (
 	"encoding/json"
+	"time"
 
 	"github.com/netobserv/flowlogs-pipeline/pkg/config"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


KalmanMeth (approved):

pkg/pipeline/decode/decode_json.go
```diff
@@ -42,6 +43,9 @@ func (c *DecodeJson) Decode(in []interface{}) []config.GenericMap {
 			continue
 		}
 		decodedLine2 := make(config.GenericMap, len(decodedLine))
+		// flows directly ingested by flp-transformer won't have this field, so we need to add it
+		// here. If the received line already contains the field, it will be overridden later
+		decodedLine2["TimeReceived"] = time.Now().Unix()
```
OlivierCazade, line 48: May be we should add a check to ensure there is no TimeReceived field?

mariomac: In that case, if there is no `TimeReceived` field, it will be added to the current payload. If there is an actual `TimeReceived` field, it will be added after this, overriding the value that we set here.

OlivierCazade: Is it what we want? To override TimeReceived if the field is already here?

PR #237, opened 2022-06-21
Netobserv 390: fix kafka transformer

OlivierCazade:
Multiple small changes in this PR:
- Fix timer in kafka ingester which was causing the FLP transformer to stop processing flows under too much traffic
- Add batch limit in kafka ingester
- Fix concurrent map access that could cause some crash when loglevel was set to debug
- Expose Kafka commit interval and default to 500ms for performances

### DIFF
--- a/pkg/api/ingest_kafka.go
+++ b/pkg/api/ingest_kafka.go
@@ -25,4 +25,5 @@ type IngestKafka struct {
 	StartOffset      string   `yaml:"startOffset,omitempty" json:"startOffset,omitempty" doc:"FirstOffset (least recent - default) or LastOffset (most recent) offset available for a partition"`
 	BatchReadTimeout int64    `yaml:"batchReadTimeout,omitempty" json:"batchReadTimeout,omitempty" doc:"how often (in milliseconds) to process input"`
 	Decoder          Decoder  `yaml:"decoder,omitempty" json:"decoder" doc:"decoder to use (E.g. json or protobuf)"`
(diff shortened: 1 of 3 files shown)


mariomac (commented):
I'd add unit tests or integration tests, if possible

mariomac (commented):
I'd add a test that actually verifies the current additions. Something like:

- Abstract current kafka listener behind an injectable interface to provide mocks/fakes.
Test 1:
- Set a very high batch read timeout (e.g. 1h)
- Send mocked kafka messages with a total size lower than batchMaxLength
- Verify that messages aren't forwarded by the output channel
- Send messages until the total size is higher than batchMaxLength
- Verify that messages are forwarded by the output channel
Test 2:
- Set a very high batchMaxlength
- set a reasonably low batch timeout: e.g. 100 ms
- Send few mocked kafka messages (size < batchMatLength), making sure they are sent before this 100ms. If we can't, increase the timeout e.g. to 200ms or ideally, also allow injecting a ticker that can be triggered on demand.
- Verify that the kafka messages are sent after the timeout

KalmanMeth:
> Abstract current kafka listener behind an injectable interface to provide mocks/fakes.
> Test 1:  Set a very high batch read timeout (e.g. 1h)
> Test 2: Set a very high batchMaxlength

Such tests could be added to test/e2e/kafka/kafka_test.go or the existing tests could be adjusted to include these.

KalmanMeth (approved):

OlivierCazade:
I added the suggested tests.

PR #239, opened 2022-06-22
Copy maps before entering Transform stages

### LINKED ISSUE #234: Error: concurrent map iteration and map write
I got this race today:

```
fatal error: concurrent map iteration and map write

goroutine 149 [running]:
runtime.throw({0x1c07bf2, 0x462959})
	/usr/local/go/src/runtime/panic.go:1198 +0x71 fp=0xc000032f48 sp=0xc000032f18 pc=0x4378b1
runtime.mapiternext(0x14)
	/usr/local/go/src/runtime/map.go:858 +0x4eb fp=0xc000032fb8 sp=0xc000032f48 pc=0x41168b
reflect.mapiternext(0x195)
	/usr/local/go/src/runtime/map.go:1346 +0x19 fp=0xc000032fd0 sp=0xc000032fb8 pc=0x462899
reflect.(*MapIter).Next(0xc001183de0)
	/usr/local/go/src/reflect/value.go:1631 +0x99 fp=0xc000032ff0 sp=0xc000032fd0 pc=0x49ba59
internal/fmtsort.Sort({0x19c0e20, 0xc001183d78, 0xc00106e620})
	/usr/local/go/src/internal/fmtsort/sort.go:63 +0x159 fp=0xc0000330b0 sp=0xc000032ff0 pc=0x4b0b99
fmt.(*pp).printValue(0xc0006fa4e0, {0x19c0e20, 0xc001183d78, 0x4d41c5}, 0x76, 0x1)
	/usr/local/go/src/fmt/print.go:769 +0x445 fp=0xc000033298 sp=0xc0000330b0 pc=0x4ecde5
fmt.(*pp).printValue(0xc0006fa4e0, {0x183e020, 0xc001c51e90, 0x1210}, 0x76, 0x0)
	/usr/local/go/src/fmt/print.go:865 +0x17ae fp=0xc000033480 sp=0xc000033298 pc=0x4ee14e
fmt.(*pp).printArg(0xc0006fa4e0, {0x183e020, 0xc001c51e90}, 0x76)
	/usr/local/go/src/fmt/print.go:712 +0x74c fp=0xc000033520 sp=0xc000033480 pc=0x4ec90c
fmt.(*pp).doPrintf(0xc0006fa4e0, {0x1be2a59, 0x10}, {0xc0000337b0, 0x47301e, 0xc0000336c8})
	/usr/local/go/src/fmt/print.go:1026 +0x288 fp=0xc000033618 sp=0xc000033520 pc=0x4ef108
fmt.Sprintf({0x1be2a59, 0x10}, {0xc0000337b0, 0x1, 0x1})
	/usr/local/go/src/fmt/print.go:219 +0x59 fp=0xc000033670 sp=0xc000033618 pc=0x4e9519
github.com/sirupsen/logrus.(*Entry).Logf(0xc0004915e0, 0x5, {0x1be2a59, 0xc0000336f8}, {0xc0000337b0, 0x18, 0x18f6ac0})
	/app/vendor/github.com/sirupsen/logrus/entry.go:338 +0x49 fp=0xc0000336b8 sp=0xc000033670 pc=0x8f72a9
github.com/sirupsen/logrus.(*Logger).Logf(0xc000227180, 0x5, {0x1be2a59, 0x10}, {0xc0000337b0, 0x1, 0x1})
	/app/vendor/github.com/sirupsen/logrus/logger.go:151 +0x85 fp=0xc000033708 sp=0xc0000336b8 pc=0x8f92a5
github.com/sirupsen/logrus.(*Logger).Debugf(...)
	/app/vendor/github.com/sirupsen/logrus/logger.go:161
github.com/sirupsen/logrus.Debugf(...)
	/app/vendor/github.com/sirupsen/logrus/exported.go:189
github.com/netobserv/flowlogs-pipeline/pkg/pipeline/ingest.(*ingestKafka).processLogLines(0xc00018a240, 0x5)
	/app/pkg/pipeline/ingest/ingest_kafka.go:100 +0x2a5 fp=0xc000033838 sp=0xc000033708 pc=0xa73945
```

I know @jpinsonneau fixed a similar one recently (in loki write), but this is another one (I do have Julien's patch while seeing this).
As far as I can tell, this one happened between a map iteration (in kafka ingest / logger) and probably a map write in `enrich`.
We should review all map writes that we do and secure them (basically, do a copy on write). Not sure if there is such a thing out of the box in golang but it would be nice to have immutable maps, explicitly needing to copy to a mutable one before writing.

cc @mariomac @eranra

jotak:
Fixes #234

### DIFF
--- a/pkg/config/generic_map.go
+++ b/pkg/config/generic_map.go
@@ -29,3 +29,12 @@ func (m GenericMap) Copy() GenericMap {
 
 	return result
 }
+
+// CopyMaps will copy a slice of GeneriMaps
+func CopyMaps(in []GenericMap) []GenericMap {
+	var result []GenericMap
+	for _, i := range in {
+		result = append(result, i.Copy())
+	}
+	return result
+}
--- a/pkg/pipeline/pipeline_builder.go
+++ b/pkg/pipeline/pipeline_builder.go
@@ -237,7 +237,7 @@ func (b *builder) getStageNode(pe *pipelineEntry, stageID string) (interface{}, 
 	case StageTransform:
 		stage = node.AsMiddle(func(in <-chan []config.GenericMap, out chan<- []config.GenericMap) {
 			for i := range in {
-				out <- pe.Transformer.Transform(i)
+				out <- pe.Transformer.Transform(config.CopyMaps(i))
 			}
 		})
 	case StageExtract:


jotak:
Opening as draft because I haven't tested yet.

I'm only copying maps before Transform, because it's really where we can assume maps will be written, but to be honest I think it would be better to copy between every stage.

cc @eranra @OlivierCazade @mariomac @KalmanMeth what do you think ; does it seem like the right place to do it?

pkg/config/generic_map.go
```diff
@@ -29,3 +29,12 @@ func (m GenericMap) Copy() GenericMap {
 
 	return result
 }
+
+// CopyMaps will copy a slice of GeneriMaps
+func CopyMaps(in []GenericMap) []GenericMap {
+	var result []GenericMap
```
mariomac, line 35: for more efficiency:

```go
result := make([]GenericMap, 0, len(in))
```

eranra (approved):
@jotak I think that this is good as is to merge ... as for the "big question" on doing that for all stages ... I thought that this is connected to https://github.com/netobserv/flowlogs-pipeline/pull/198 ... I do not know what @mariomac thinks on that.

KalmanMeth:
I think the copy should be performed by exactly those stages that make changes to the configMap. Each transform receives configMaps, and should treat them as read-only. If any change is made, then the stage should either make a copy or should create its own new array of configMaps.

OlivierCazade:
I agree with Kalman proposal. For performances, we should avoid as much unnecessary copy as possible.

mariomac:
The way the pipeline library works, only when a stage finishes processing a map, it forwards it to the next stage. So in principle we should be sure that only one stage/goroutine has the map at the same time.

However, I was not considering the scenario in which a stage sends the map to two stages, and one of the stages modify the contents. E.g. the ingester sends the map to the transformer and to a stdout logger. This is the only scenario in which I could imagine we could have this kind of concurrent modification.

I guess that for now might be Ok, and for a further optimization, we could improve the pipeline builder to inject a cloning node only if this kind of path forks happen. E.g. normally it would build a map like:

```
ingest/decode --> transform --> write
```

But if it detects a fork where one of the destination nodes transform the path, inject a cloner like:

```
ingest/decode --> stdout_log
              \-> clone --> transform --> write
```

jotak:
Closing in favor of #240

PR #247, opened 2022-06-29
conntrack: Integrate module with the pipeline

ronensc:
This PR makes the connection tracking module usable as part of the pipeline in the form of a new stage.
I added an integration test to demonstrate its usage and a new fake ingester to facilitate testing.


pkg/config/config.go
```diff
@@ -46,12 +46,13 @@ type Stage struct {
...
-	Name      string     `json:"name"`
-	Ingest    *Ingest    `json:"ingest,omitempty"`
-	Transform *Transform `json:"transform,omitempty"`
-	Extract   *Extract   `json:"extract,omitempty"`
-	Encode    *Encode    `json:"encode,omitempty"`
-	Write     *Write     `json:"write,omitempty"`
+	Name      string         `json:"name"`
+	Ingest    *Ingest        `json:"ingest,omitempty"`
+	Transform *Transform     `json:"transform,omitempty"`
+	ConnTrack *api.ConnTrack `json:"conntrack,omitempty"`
```
jpinsonneau, line 52: Should we consider something more generic for consistency like `track` ?

pkg/api/utils.go
```diff
@@ -0,0 +1,55 @@
...
+		return nil
+	case string:
+		var err error
+		d.Duration, err = time.ParseDuration(value)
+		if err != nil {
+			return err
+		}
+		return nil
+	default:
+		return errors.New("invalid duration")
```
jpinsonneau, line 36: value would be usefull in the error message

jpinsonneau (commented):
Looks great @ronensc ! Thanks

pkg/config/config.go
ronensc: The inconsistency bothers me too. At first I added an abstraction layer, but then I removed it because the config ended up with consecutive levels of `conntrack` which looked funny to me:
```yaml
pipeline: ...
parameters: ...
  - name: conntrack
    contrack:
      conntrack:
        ...
      type: ...
```
https://github.com/netobserv/flowlogs-pipeline/pull/247/commits/30aa7e8382f13e1c422a35581af2efb63cbbb1ec#diff-81d2b1554fae3c0d5186eeb17c96d0dbb9373fd8867220245e2612b4bd26849eL45-L50

I didn't think of making the added layer a generic layer for tracking. I'll take your advice :)

pkg/api/conn_track.go
```diff
@@ -29,7 +27,7 @@ type ConnTrack struct {
 	KeyDefinition        KeyDefinition `yaml:"keyDefinition,omitempty" doc:"fields that are used to identify the connection"`
 	OutputRecordTypes    []string      `yaml:"outputRecordTypes,omitempty" enum:"ConnTrackOutputRecordTypeEnum" doc:"output record types to emit"`
 	OutputFields         []OutputField `yaml:"outputFields,omitempty" doc:"list of output fields"`
-	EndConnectionTimeout time.Duration `yaml:"endConnectionTimeout,omitempty" doc:"duration of time to wait from the last flow log to end a connection"`
+	EndConnectionTimeout Duration      `yaml:"endConnectionTimeout,omitempty" doc:"duration of time to wait from the last flow log to end a connection"`
```
KalmanMeth, line 30: duration in time in seconds?

ronensc: Not necessary. The config value is passed to `time.ParseDuration()` during unmarshalling.
https://pkg.go.dev/time#ParseDuration
So for example `30s` will be parsed as 30 seconds and `2m` will be parsed as 2 minutes.

pkg/config/config.go
jpinsonneau: yeah let's keep it simple :smile:

jotak: but conn_track couldn't be an implementation of a standard Transform stage ?

KalmanMeth (approved):

pkg/pipeline/conntrack_integ_test.go
```diff
@@ -0,0 +1,154 @@
...
+            fieldGroupARef: src 
+            fieldGroupBRef: dst 
+        outputFields:
+          - name: Bytes
+            operation: sum
+            SplitAB: true
+          - name: Packets
+            operation: sum
+            SplitAB: true
+          - name: numFlowLogs
```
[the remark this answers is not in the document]
ronensc: It's the only outputField that doesn't depend on any input field. But that's not a good reason 😅

pkg/pipeline/conntrack_integ_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	in := mainPipeline.pipelineStages[0].Ingester.(*ingest.IngestFake).In
+	writer := mainPipeline.pipelineStages[2].Writer.(*write.WriteFake)
+
+	ingestFile(t, in, "../../hack/examples/ocp-ipfix-flowlogs.json")
+	writer.Wait()
+	writer.ResetWait()
+
+	// Wait a moment to make the connections expired
+	time.Sleep(2 * time.Second)
```
ronensc, line 112: I'm not very pleased with this 2 second sleep. It obviously slows the test and on if the running platforms is loaded it might fail.
I can remediate this by injecting a fake clock and simulate the sleep. But, on the other hand, I think of this test as an integration test and I'm not sure if we want to fake the clock. WDYT?

pkg/config/config.go
ronensc: I'll follow @KalmanMeth and your suggestion and make `conntrack` a type of `extract`

PR #251, opened 2022-07-07
adjustments to confgenerator for Operator

KalmanMeth:
Include only extract-aggregate and encode-prom rules in the produced config file.

### DIFF
--- a/pkg/confgen/flowlogs2metrics_config.go
+++ b/pkg/confgen/flowlogs2metrics_config.go
@@ -25,7 +25,7 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-func (cg *ConfGen) generateFlowlogs2PipelineConfig() map[string]interface{} {
+func (cg *ConfGen) GenerateFlowlogs2PipelineConfig() map[string]interface{} {
 	config := map[string]interface{}{
 		"log-level": "error",
 		"pipeline": []map[string]string{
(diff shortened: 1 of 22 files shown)


pkg/confgen/flowlogs2metrics_config.go
```diff
@@ -97,7 +98,84 @@ func (cg *ConfGen) generateFlowlogs2PipelineConfig(fileName string) error {
...
+			parameters[i] = map[string]interface{}{
+				"name": "ingest_collector",
+				"ingest": map[string]interface{}{
+					"type": "collector",
+					"collector": map[string]interface{}{
+						"port":       cg.config.Ingest.Collector.Port,
+						"portLegacy": cg.config.Ingest.Collector.PortLegacy,
+						"hostname":   cg.config.Ingest.Collector.HostName,
+					},
+				},
```
ronensc, line 118: I think we now have all these data structures defined in https://github.com/netobserv/flowlogs-pipeline/tree/main/pkg/api
Maybe we should consider reusing them rather than having them repeated here?

ronensc: The high-level of StageParam is defined in a different file:
https://github.com/netobserv/flowlogs-pipeline/blob/467e409460cef7c7f408cc3ab38476a5da1e03aa/pkg/config/config.go#L48-L55

And perhaps the pipeline config builder could become handy
https://github.com/netobserv/flowlogs-pipeline/blob/main/pkg/config/pipeline_builder.go

KalmanMeth: If we want to redo this, let's put it in a separate issue and PR.

ronensc (approved):

pkg/confgen/flowlogs2metrics_config.go
ronensc: I created an issue for this
https://github.com/netobserv/flowlogs-pipeline/issues/254

PR #259, opened 2022-07-19
NETOBSERV-465: Kafka metrics 2

OlivierCazade:
Added more metrics to the kafka ingester and added timeout mecanics to the kafka tests.

### DIFF
--- a/pkg/pipeline/ingest/ingest_kafka.go
+++ b/pkg/pipeline/ingest/ingest_kafka.go
@@ -206,9 +206,9 @@ func NewIngestKafka(params config.StageParam) (Ingester, error) {
 		commitInterval = jsonIngestKafka.CommitInterval
 	}
 
-	dialer := &kafka.Dialer{
-		Timeout:   kafka.DefaultDialer.Timeout,
-		DualStack: kafka.DefaultDialer.DualStack,
+	dialer := &kafkago.Dialer{
+		Timeout:   kafkago.DefaultDialer.Timeout,
+		DualStack: kafkago.DefaultDialer.DualStack,
 	}
 	if jsonIngestKafka.TLS != nil {
 		log.Infof("Using TLS configuration: %v", jsonIngestKafka.TLS)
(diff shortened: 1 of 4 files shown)


pkg/pipeline/ingest/ingest_kafka.go
```diff
@@ -74,14 +76,61 @@ func (ingestK *ingestKafka) kafkaListener() {
...
 
 }
 
+func processRecordDelay(record config.GenericMap) {
+	TimeFlowEndInterface, ok := record["TimeFlowEnd"]
+	if !ok {
+		flowErrors.With(prometheus.Labels{"router": "", "error": "No TimeFlowEnd found"}).Inc()
+		return
+	} else {
+		fmt.Println(TimeFlowEndInterface)
```
KalmanMeth, line 101: Do you really want to print to stdout for each end of flow? or is this only for debug purposes?

OlivierCazade: Definitly not, this is a debug trace that I forgot to remove.
Thanks

pkg/test/utils.go
```diff
@@ -163,3 +165,14 @@ func DeserializeJSONToMap(t *testing.T, in string) config.GenericMap {
 	require.NoError(t, err)
 	return m
 }
+
+func WaitFromChannel(in chan []config.GenericMap, timeout int64) ([]config.GenericMap, error) {
+	duration := time.Duration(timeout) * time.Millisecond
```
ronensc, line 170: Won't it be more idiomatic to change the type of `timeout` from `int64` to `time.Duration`?

Then, in the invocation side, one can argue that it is more apparent what the unit of time is:
```go
test.WaitFromChannel(ingestOutput, 2 * time.Second)
```

ronensc (approved):

pkg/test/utils.go
OlivierCazade: Good point thanks!
I changed it.

jpinsonneau (approved):

PR #264, opened 2022-07-22
vendor dependencies and speedup container build

mariomac:
Having to download vendor dependencies on each build complicates the downstream compilation process. This PR vendors them.

It also slightly optimizes the container build process, separating the linting stage from the build_code stage. Having to download golangci-lint on each build made the process too slow.

Now the build time for a container is ~100seconds, which could be halved if we provided two different container images: a "production" FLP that only compiles and includes the flowlogs-pipeline binary (as we will have downstream), and a "debug" image that also includes the confgenerator, as well as the iputils, curl and net-tools packages.


jotak:
/lgtm
About production image versus debug: maybe that's a difference we can leave between upstream and downstream (with the production image only built for downstream) ?

eranra:
@mariomac I very much like the idea of debug vs. production ... I also had discussions with @ronensc on how to use remote debug

PR #266, opened 2022-07-26
NETOBSERV-497 allow direct metrics pipeline

jotak:
This PoC creates a new "SimpleProm" Encode stage that directly extracts
metrics from flows without an intermediate Aggregate stage

The rationale is that Prometheus already manages labels aggregations.
It leaves more responsibilities (and more power) to the query side.
For instance, aggregations such as sum/avg etc. would be performed in
PromQL at query time rather than upstream.

Pipelines to generate are simpler as they don't need an "Aggregate" stage, and the prom-encode stage itself has less parameters, e.g.:

```go
	enrichedStage.SimpleEncodePrometheus("prometheus", api.SimplePromEncode{
		Port:   int(b.desired.PrometheusPort),
		Prefix: "netobserv_",
		Metrics: []api.SimplePromMetricsItem{{
			Name:      "bytes_total",
			RecordKey: "Bytes",
			Type:      "counter",
			Labels:    []string{"SrcK8S_Namespace", "DstK8S_Namespace"},
		}, {
			Name:      "packets_total",
			RecordKey: "Packets",
			Type:      "counter",
			Labels:    []string{"SrcK8S_Namespace", "DstK8S_Namespace"},
		}},
	})
```


jotak:
This pipeline (cf go code above) results in creating this metrics, with both src/dest namespaces as labels :

Capture d’écran du 2022-07-26 13-48-36

eranra:
@jotak how many metrics will be sent to Prometheus? don't you think that this will exhaust Prometheus with too many metrics? the idea of aggregate was to pre-process metrics so that the only thing Prometheus will need to handle is the aggregates. For example in the case of the above Prometheus will get namespace metrics without the need to worry about the underly metrics. Maybe I am missing something here ... we can find time to talk tomorrow maybe?

jotak:
@eranra I already proposed a call on thursday (check your mail) :)

eranra:
@jotak I see that now ... can we move that to an earlier time ... morning EU time --- this overlaps multiple meetings for me

jotak:
btw we would also need to check how to combine that with confgen ; maybe adding a config flag in metrics definitions to tell if we want to use the "direct prom" approach or the "aggregate + prom" approach.

jotak:
Also, for the record: we need to check if there is an internal cache in prom client to clean from time to time (similar to the expiry mechanism implemented in FLP caches)

KalmanMeth:
In encode_prom, we implemented a cache to clean up items that are inactive, and we encapsulated it in utils.timed_cache.

KalmanMeth:
The encode_prom stage does not depend on the extract_aggregate stage. The confgenerator builds a config that uses  extract_aggregate  that feeds encode_prom , but encode_prom can be used by itself without aggregation. Once you add in the cache to SimpleEncodeProm it becomes essentially the existing encode_prom.

jotak:
@KalmanMeth yes, I see that now. Maybe with the exception of the histogram values, right?

jotak:
@KalmanMeth I think the existing `PromEncode` is _almost_ working well as a independent stage, like what I was trying to do, I see however two issues:
- The `PromMetricsFilter` in PromEncode api prevents from exposing any metrics we'd like
- Histogram are not working

I believe these are changes added later, without having in mind that the prom-encode stage could be used independently?

In my last commit I've added twice a test of exposed metrics, see https://github.com/netobserv/flowlogs-pipeline/pull/266/commits/eb99bc0bd094b3904d6ff8927cd5ff0b7a0df621 ; one time for the existing PromEncode, one time for the new PoC prom encode.
I would expect the test should work on both, but it only works on the SimpleProm one because of the two issues mentioned.

jotak:
So, rather than having a new Encode implementation, I will try to fix the existing one.

jotak:
last commit: now focusing on fixing the existing PromEncode to work fine without Aggregation. Also spending some time on optimization (e.g. decreasing number of allocs)

jotak:
I see another issue with the current Aggregation stage: we cannot mix "direct" PromEncode stages and "non-direct" (ie. following Aggregation) in confgen: we would need to fork before Aggregation, something like : 
```
transform -> aggregate
aggregate -> prom1
transform -> prom2
```

But confgen doesn't support that. Also, having 2 prom stages means we need to deal with port conflicts.
So at the moment, it's something like all or nothing : either we have only "aggregate+prom" in the pipeline, or we have only "direct-prom", but not a mix of them.

jotak:
FYI, jira created : https://issues.redhat.com/browse/NETOBSERV-497

pkg/pipeline/encode/encode_prom.go
```diff
@@ -34,153 +36,190 @@ import (
...
 var metricsProcessed = operationalMetrics.NewCounter(prometheus.CounterOpts{
 	Name: "encode_prom_metrics_processed",
 	Help: "Number of metrics processed",
 })
 
+var errorsCounter = operationalMetrics.NewCounterVec(prometheus.CounterOpts{
+	Name: "encode_prom_errors",
+	Help: "Total errors during metrics generation",
+}, []string{"error", "metric", "key"})
+
```
ronensc, line 74: 👍

pkg/pipeline/encode/encode_prom_test.go
```diff
@@ -62,124 +66,238 @@ parameters:
...
-	require.NotNil(t, v)
-
-	newEncode, err := NewEncodeProm(cfg.Parameters[0])
-	require.Equal(t, err, nil)
-	return newEncode
+func initPromWithServer(params *api.PromEncode) (*EncodeProm, func(), error) {
+	reg := prometheus.NewRegistry()
+	prometheus.DefaultRegisterer = reg
+	prometheus.DefaultGatherer = reg
+	http.DefaultServeMux = http.NewServeMux()
```
ronensc, line 79: What's the purpose of this setting?

ronensc (approved):
This looks promising. It feels like the code is cleaner.
@jotak `I'm amazed by how fast you'v`e done this work 🤩

pkg/pipeline/encode/encode_prom_test.go
jotak: this is because in `encode_prom`, we run `http.Handle("/metrics", promhttp.Handler())`
which creates a handler on the default mux/router. When this is called several times, an error would be fired (something like "cannot register /metrics, route already exists").

I don't remember exactly in which case that happened, maybe it was just in the benchmark I created below.

BTW it also shows that the prom_encode stage will need to be refactored if I some point we want to be able to define more than one prom-encodes in a pipeline. I mentioned that in this jira: https://issues.redhat.com/browse/NETOBSERV-498

pkg/pipeline/encode/encode_prom.go
```diff
@@ -196,44 +235,48 @@ func (e *EncodeProm) cleanupExpiredEntriesLoop() {
...
-	for _, mInfo := range jsonEncodeProm.Metrics {
-		var pMetric PromMetric
-		fullMetricName := promPrefix + mInfo.Name
+	counters := []counterInfo{}
+	gauges := []gaugeInfo{}
+	histos := []histoInfo{}
+	aggHistos := []histoInfo{}
+
+	for i := range config.Metrics {
+		mInfo := config.Metrics[i]
```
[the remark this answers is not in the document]
jotak: yes, I don't know why I wrote it that way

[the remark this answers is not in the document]
jotak: Now I remember: that's because a pointer to `mInfo` is use below, so we need to do a copy somewhere. But I'll do differently, I'll change the model to not use a pointer, so the copy is implicit

pkg/pipeline/encode/encode_prom.go
```diff
@@ -34,153 +36,190 @@ import (
...
-	switch entry.PromMetric.metricType {
-	case api.PromEncodeOperationName("Gauge"):
-		entry.PromMetric.promGauge.Delete(entry.eInfo.Labels)
-	case api.PromEncodeOperationName("Counter"):
-		entry.PromMetric.promCounter.Delete(entry.eInfo.Labels)
-	case api.PromEncodeOperationName("Histogram"):
-		entry.PromMetric.promHist.Delete(entry.eInfo.Labels)
-	}
+func (e *EncodeProm) Cleanup(cleanupFunc interface{}) {
+	cleanupFunc.(func())()
```
KalmanMeth, line 222: Where do we now cleanup old entries?

pkg/pipeline/encode/encode_prom_test.go
```diff
@@ -62,124 +66,241 @@ parameters:
...
-		},
-		"value": float64(7),
+	exposed := test.ReadExposedMetrics(t)
+
+	require.Contains(t, exposed, `test_bytes_total{dstIP="10.0.0.1",srcIP="20.0.0.2"} 8`)
+	require.Contains(t, exposed, `test_bytes_total{dstIP="30.0.0.3",srcIP="10.0.0.1"} 12`)
+	require.Contains(t, exposed, `test_packets_total{dstIP="10.0.0.1",srcIP="20.0.0.2"} 2`)
+	require.Contains(t, exposed, `test_packets_total{dstIP="30.0.0.3",srcIP="10.0.0.1"} 2`)
+	require.Contains(t, exposed, `test_latency_seconds_bucket{dstIP="10.0.0.1",srcIP="20.0.0.2",le="0.025"} 0`)
+	require.Contains(t, exposed, `test_latency_seconds_bucket{dstIP="10.0.0.1",srcIP="20.0.0.2",le="0.05"} 1`)
```
KalmanMeth, line 192: Where are these buckets defined? Are they a default?

pkg/pipeline/encode/encode_prom_test.go
```diff
@@ -62,124 +66,241 @@ parameters:
...
+	}
+
+	prom, cleanup, err := initPromWithServer(&params)
+	require.NoError(b, err)
+	defer cleanup()
+
+	for i := 0; i < b.N; i++ {
+		prom.Encode(hundredFlows())
+	}
 }
```
KalmanMeth, line 306: @jotak I learned a lot from this re-write of encode_prom and its test.

pkg/pipeline/encode/encode_prom_test.go
jotak: yes, prom client defines default: `DefBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}`

This hasn't changed from the previous implementation

KalmanMeth (approved):

pkg/confgen/encode.go
```diff
@@ -41,6 +41,15 @@ func (cg *ConfGen) parseEncode(encode *map[string]interface{}) (*api.PromEncode,
 		return nil, err
 	}
 
+	// Histograms built from Aggregate need to be flagged as they are handled in a different way in PromEncode
```
jotak, line 44: I must remove this part, the same is done in `flowlogs2metrics_config.go`

pkg/pipeline/encode/encode_prom_test.go
ronensc: @jotak thanks. I tried commenting out that line and see what happens. It failed on the second unit test of `encode_prom_test.go` because of the multiple registration that you have described. So resetting the `DefaultServeMux` to a new instance on each unit test solves this problem.

Issue #265, opened 2022-07-26
Make FLP able to generate labels per aggregate


KalmanMeth:
https://issues.redhat.com/browse/NETOBSERV-490 

Currently, FLP merges all aggregate keys in an unique prom label called "aggregate".

It restricts what we can further query with promQL. To have full querying capabilities, we should have 1 label per aggregate.

E.g. for aggregates defined for SrcNamespace, DstNamespace, SrcSubnet, DstSubnet we should be able to get each of these as individual prom labels (instead of all concatenated in "by" and "aggregate" clauses).

PR #269, opened 2022-07-28
enable individual labels from aggregate

KalmanMeth:
Addresses #265 
https://issues.redhat.com/browse/NETOBSERV-490

### DIFF
--- a/pkg/pipeline/extract/aggregate/aggregate_test.go
+++ b/pkg/pipeline/extract/aggregate/aggregate_test.go
@@ -97,8 +97,13 @@ func Test_FilterEntry(t *testing.T) {
 	entry := test.GetIngestMockEntry(false)
 
 	err, _, _ := aggregate.FilterEntry(entry)
-
 	require.Equal(t, err, nil)
+
+	err, normalizedLabels, labels := aggregate.FilterEntry(entry)
+	require.Equal(t, err, nil)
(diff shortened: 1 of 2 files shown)


jotak (approved):
LGTM
Maybe we can think about a follow-up, because it creates redundant information in metrics (there are individual labels AND concatenated labels) => maybe a bool flag, or an enum in Aggregation api to let the user choose how they want the result to look like? E.g. `labelsMode: enum(copy|concat)`

pkg/pipeline/extract/aggregate/aggregate_test.go
```diff
@@ -96,12 +96,12 @@ func Test_FilterEntry(t *testing.T) {
 	aggregate := GetMockAggregate()
 	entry := test.GetIngestMockEntry(false)
 
-	err, _ := aggregate.FilterEntry(entry)
+	err, _, _ := aggregate.FilterEntry(entry)
 
 	require.Equal(t, err, nil)
```
ronensc, line 101: I would add a few more checks to test the new behavior.
Something like:
```go
	err, normalizedLabels, labels := aggregate.FilterEntry(entry)
	require.Equal(t, err, nil)
	require.Equal(t, Labels{"srcIP": "10.0.0.1", "dstIP": "20.0.0.2"}, labels)
	require.Equal(t, NormalizedValues("20.0.0.2,10.0.0.1"), normalizedLabels)
```

ronensc (approved):

PR #271, opened 2022-07-28
Make yaml.Unmarshal strict

ronensc:
Using `yaml.UnmarshalStrict()` will prevent us from making typos in our yamls without being noticed.

### DIFF
--- a/pkg/api/utils_test.go
+++ b/pkg/api/utils_test.go
@@ -37,6 +37,6 @@ func TestDuration_MarshalYAML(t *testing.T) {
 func TestDuration_UnmarshalYAML(t *testing.T) {
 	expectedMsg := testMessage{Elapsed: Duration{2 * time.Hour}}
 	var actualMsg testMessage
-	require.NoError(t, yaml.Unmarshal([]byte("elapsed: 2h\n"), &actualMsg))
+	require.NoError(t, yaml.UnmarshalStrict([]byte("elapsed: 2h\n"), &actualMsg))
 	require.Equal(t, expectedMsg, actualMsg)
 }
--- a/pkg/confgen/confgen_test.go
+++ b/pkg/confgen/confgen_test.go
@@ -55,7 +55,6 @@ encode:
     prefix: flp_
     port: 9102
 visualization:
-  type: grafana
   grafana:
     dashboards:
       - name: "details"
@@ -90,7 +89,6 @@ write:
     staticLabels:
       job: flowlogs-pipeline
 visualization:
-  type: grafana
   grafana:
     dashboards:
       - name: "details"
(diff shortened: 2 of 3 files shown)


pkg/api/utils_test.go
```diff
@@ -37,6 +37,6 @@ func TestDuration_MarshalYAML(t *testing.T) {
 func TestDuration_UnmarshalYAML(t *testing.T) {
 	expectedMsg := testMessage{Elapsed: Duration{2 * time.Hour}}
 	var actualMsg testMessage
-	require.NoError(t, yaml.Unmarshal([]byte("elapsed: 2h\n"), &actualMsg))
```
KalmanMeth, line 40: @ronensc What is your general rule when you want to use UnmarshalStrict as opposed to Unmarshal?

ronensc: Assume we have a type like this:
```go
type Person struct {
	Name string `yaml:"name"`
	Age  int    `yaml:"age"`
}
```

and data that we want to parse:
```yaml
person:
  name: John
  Age: 42
```

Notice that the age field is misspelled (it should be `age` with lowercase a).

Using unstrict `yaml.Unmarshal()` won't raise an error on this. It will simply ignore the unrecognized field `Age` and set the default value for the missing field `age`.

I think in our use-case, where we parse config files, we want to make sure that all fields are recognized and emit an error otherwise.

KalmanMeth: So aren't there numerous places in the code where we Marshal the config data where we should also use UnmarshalStrict ?

ronensc: I replaced all the `yaml.Unmarshal()` invocations that I found (I found only 3).
Please let me know if I missed something.

KalmanMeth: There are also lots of json.Unmarshal() calls.

jotak (approved):
lgtm

pkg/api/utils_test.go
ronensc: Unfortunately, the `json` library doesn't have a strict version of `Unmarshal()` like the `yaml` has.
But, we can get the strict behavior with `json.NewDecoder` and `DisallowUnknownFields()` as suggested here:
https://groups.google.com/g/golang-nuts/c/MJuJDte5YHg/m/JMs0S-4uAQAJ

KalmanMeth: Can't some of those be changed to to yaml.UnmarshalStrict?

ronensc:
**Note**: This PR deals only with `confgen`. That is, `flowlogs-pipeline` would start normally even if its config has unknown fields. I created an issue for it and will solve it in a later PR. #274

PR #272, opened 2022-07-31
conntrack: add conntrack to confgen and create connection size in bytes histogram metric

ronensc:
This PR adds the conntrack to the confgen. It's added between the `transform` stage and the `aggregate` stage.
In addition, this PR adds the first metric the uses conntrack data: connection length histogram.

image
image


contrib/kubernetes/flowlogs-pipeline.conf.yaml
```diff
@@ -1,15 +1,30 @@
 # This file was generated automatically by flowlogs-pipeline confgenerator
 log-level: error
+pipeline:
```
KalmanMeth, line 3: So the output now puts the data structures back in natural order (instead of the golang default of lexicographical order); `pipeline` before `parameters`; `name` before `follows`; etc. Good!

ronensc: I think that's a nice side effect of #267 rather than this PR 😅

contrib/kubernetes/flowlogs-pipeline.conf.yaml
```diff
@@ -71,78 +90,156 @@ parameters:
...
+        - name: protocol
+          fields:
+          - proto
+        hash:
+          fieldGroupRefs:
+          - protocol
+          fieldGroupARef: src
+          fieldGroupBRef: dst
+      outputRecordTypes:
+      - newConnection
```
KalmanMeth, line 116: Do we describe anywhere how to use these fields? in the README?

contrib/kubernetes/flowlogs-pipeline.conf.yaml
```diff
@@ -71,78 +90,156 @@ parameters:
...
+      endConnectionTimeout: 30s
+- name: extract_aggregate
+  extract:
+    type: aggregates
     aggregates:
     - name: bandwidth_network_service
       by:
       - service
+      - _RecordType
       operation: sum
```
KalmanMeth, line 149: Should we use this syntax ( _ prefix) also for other variables defined internally, such as `aggregate` and `by` in the prom stage?

docs/metrics.md
```diff
@@ -17,8 +17,8 @@ and the transformation to generate the exported metric.
 |:---|:---|
 | **Details** | Sum bytes for all traffic per network service | 
 | **Usage** | Evaluate network usage breakdown per network service | 
-| **Labels** | bandwidth, graph, rate, network-service |
-| **Operation** | aggregate by `service` and `sum` field `bytes` |
+| **Tags** | bandwidth, graph, rate, network-service |
+| **Operation** | aggregate by `service, _RecordType` and `sum` field `bytes` |
```
KalmanMeth, line 21: Is _RecordType specified anywhere in the documentation?

KalmanMeth (approved):

contrib/kubernetes/flowlogs-pipeline.conf.yaml
ronensc: Not yet

contrib/kubernetes/flowlogs-pipeline.conf.yaml
ronensc: We can open that for a discussion. I'm not against. @eranra @jotak WDYT?

docs/metrics.md
ronensc: not yet

contrib/kubernetes/flowlogs-pipeline.conf.yaml
```diff
@@ -71,78 +90,156 @@ parameters:
       - input: dstIP
         output: dstLocation
         type: add_location
-    type: network
-- extract:
+- name: extract_conntrack
```
jotak, line 93: do you know if this is going to be compatible with a "direct prometheus" pipeline, ie. without Aggregation stage?
Said differently, if a prom-encode is wired directly after extract_conntrack, will it work? (granted we include fixes from https://github.com/netobserv/flowlogs-pipeline/pull/266 )
I guess there will be some adjustments to do so that prom-encode picks only "flowLog" output, and ignore "new/endConnection"

ronensc: > do you know if this is going to be compatible with a "direct prometheus" pipeline, ie. without Aggregation stage?
I can't think why not, so I would say yes. I think it should work.
Instead of adding adjustments to prom-encode to pick only "flowLog" records, we can add a `Transform` step between `extract_conntrack` and `encode_prom` to filter the records.

contrib/kubernetes/flowlogs-pipeline.conf.yaml
jotak: +1 but this can be done outside of this PR

PR #276, opened 2022-08-02
Added assign_if transformation

### LINKED ISSUE #275: Add option for transforming to some custom values on the basis of some parameters
We have `InIf` and `OutIf` flags represents traffic direction and can have integer values. now we want to change to some custom values say `In` or `Out`, To do this we need a way so that in configuration we can tell on some parameters replace input value with some other value.

tkuldeep:
Fixes #275 
Signed-off-by: Kuldeep Singh Pal <tkuldeep.allen@gmail.com>

### DIFF
--- a/pkg/api/transform_network.go
+++ b/pkg/api/transform_network.go
@@ -43,6 +43,7 @@ type NetworkTransformRule struct {
 	Output     string `yaml:"output,omitempty" json:"output,omitempty" doc:"entry output field"`
 	Type       string `yaml:"type,omitempty" json:"type,omitempty" enum:"TransformNetworkOperationEnum" doc:"one of the following:"`
 	Parameters string `yaml:"parameters,omitempty" json:"parameters,omitempty" doc:"parameters specific to type"`
+	Assignee   string `yaml:"assignee,omitempty" json:"assignee,omitempty" doc:"value needs to assign to output field"`
 }
 
 type NetworkTransformRules []NetworkTransformRule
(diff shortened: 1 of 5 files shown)


pkg/api/transform_network.go
```diff
@@ -43,6 +44,7 @@ type NetworkTransformRule struct {
 	Output     string `yaml:"output,omitempty" json:"output,omitempty" doc:"entry output field"`
 	Type       string `yaml:"type,omitempty" json:"type,omitempty" enum:"TransformNetworkOperationEnum" doc:"one of the following:"`
 	Parameters string `yaml:"parameters,omitempty" json:"parameters,omitempty" doc:"parameters specific to type"`
+	Assignee   string `yaml:"assignee,omitempty" json:"assignee,omitempty" doc:"value needs to assign to output field"`
```
KalmanMeth, line 47: Please re-use existing Parameters field instead of new field.

pkg/pipeline/transform/transform_network.go
```diff
@@ -95,6 +95,18 @@ func (n *Network) TransformEntry(inputEntry config.GenericMap) config.GenericMap
 				outputEntry[rule.Output] = outputEntry[rule.Input]
 				outputEntry[rule.Output+"_Evaluate"] = true
 			}
+		case api.TransformNetworkOperationName("AssignIf"):
+			expressionString := fmt.Sprintf("val %s", rule.Parameters)
```
KalmanMeth, line 99: Eran asked to combine the `AssignIf` into the existing `AddIf`. The code is essentially the same, with the exception of what gets assigned to the the Output variable. If an assigned value is provided, it gets the new value; otherwise it gets the value of the Input variable.

eranra (approved):

KalmanMeth (approved):

PR #279, opened 2022-08-03
Make unmarshaling json config strict

### LINKED ISSUE #274: Crash FLP on an incorrect config file
Currently, FLP will start normally even if its config has unknown fields.
The problem is that misspelled fields will silently be ignored and have default values.
It's better to crash FLP with an indicative error.

ronensc:
Unmarshaling non-config related json still use the non-strict version.

This is a follow-up on #271 
Fix #274

### DIFF
--- a/pkg/test/utils.go
+++ b/pkg/test/utils.go
@@ -143,6 +143,7 @@ func RunCommand(command string) string {
 }
 
 func DeserializeJSONToMap(t *testing.T, in string) config.GenericMap {
+	t.Helper()
 	var m config.GenericMap
 	err := json.Unmarshal([]byte(in), &m)
 	require.NoError(t, err)
--- a/contrib/kubernetes/flowlogs-pipeline.conf.yaml
+++ b/contrib/kubernetes/flowlogs-pipeline.conf.yaml
@@ -1,15 +1,28 @@
 # This file was generated automatically by flowlogs-pipeline confgenerator
 log-level: error
+pipeline:
+- name: ingest_collector
+- name: transform_generic
+  follows: ingest_collector
+- name: transform_network
+  follows: transform_generic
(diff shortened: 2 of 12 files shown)


pkg/test/utils.go
```diff
@@ -143,6 +143,7 @@ func RunCommand(command string) string {
 }
 
 func DeserializeJSONToMap(t *testing.T, in string) config.GenericMap {
+	t.Helper()
```
KalmanMeth, line 146: What is the practical result of declaring this function as a Helper function?

KalmanMeth (approved):

pkg/test/utils.go
ronensc: @KalmanMeth 
By marking a function as helper, when a test fails in the helper function, the line number reported will be of the caller function rather than inside the test helper function.

Consider running the following test:

```golang
package b

import (
	"testing"
)

func Test(t *testing.T) {
	helperMarked(t, "hello", "hi") // line 8
	helperUnmarked(t, "hello", "hi")
}

func helperMarked(t *testing.T, expected, actual string) {
	t.Helper()
	if expected != actual {
		t.Errorf("expected (%v) != actual(%v)", expected, actual)
	}
}
func helperUnmarked(t *testing.T, expected, actual string) {
	if expected != actual {
		t.Errorf("expected (%v) != actual(%v)", expected, actual) // line 20
	}
}
```

The result is:
```
=== RUN   Test
    b_test.go:8: expected (hello) != actual(hi)
    b_test.go:20: expected (hello) != actual(hi)
--- FAIL: Test (0.00s)

FAIL
```

jotak: nice, I didn't know that

jotak (approved):
lgtm

PR #282, opened 2022-08-04
NETOBSERV-396: Expose TLS configuration for prometheus endpoint

OlivierCazade:
This enable TLS configuration for the prometheus endpoint

### DIFF
--- a/pkg/api/encode_prom.go
+++ b/pkg/api/encode_prom.go
@@ -18,9 +18,8 @@
 package api
 
 type PromTLSConf struct {
-	Enable   bool   `yaml:"enable,omitempty" json:"enable,omitempty" doc:"set to true to enable tls for the prometheus endpoint"`
-	CertFile string `yaml:"certFile,omitempty" json:"certFile,omitempty" doc:"path to the certificate file"`
-	KeyFile  string `yaml:"keyFile,omitempty" json:"keyFile,omitempty" doc:"path to the key file"`
+	CertPath string `yaml:"certPath,omitempty" json:"certPath,omitempty" doc:"path to the certificate file"`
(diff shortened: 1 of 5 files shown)


pkg/api/encode_prom.go
```diff
@@ -17,11 +17,18 @@
 
 package api
 
+type PromTLSConf struct {
+	Enable   bool   `yaml:"enable,omitempty" json:"enable,omitempty" doc:"set to true to enable tls for the prometheus endpoint"`
+	CertFile string `yaml:"certFile,omitempty" json:"certFile,omitempty" doc:"path to the certificate file"`
```
jotak, line 22: Nit: in other places we use "Path" rather than "File"
E.g.
```golang
	CACertPath         string `yaml:"caCertPath,omitempty" json:"caCertPath,omitempty" doc:"path to the CA certificate"`
```

Just for consistency, we should align here

pkg/api/encode_prom.go
```diff
@@ -17,11 +17,18 @@
 
 package api
 
+type PromTLSConf struct {
+	Enable   bool   `yaml:"enable,omitempty" json:"enable,omitempty" doc:"set to true to enable tls for the prometheus endpoint"`
```
jotak, line 21: In kafka/loki we don't use an `enable` flag, we use a pointer instead and check for nil. (I know we can't do that in the operator, but here it's different)
So here also I think we should be consistent

OlivierCazade: Changed it,
Thanks

jotak (approved):
lgtm, thanks Olivier

KalmanMeth (approved):

PR #286, opened 2022-08-09
NETOBSERV-466: Packet metrics

OlivierCazade:
Add two new metrics to the ingest stages.

### DIFF
--- a/pkg/pipeline/ingest/ingest_kafka.go
+++ b/pkg/pipeline/ingest/ingest_kafka.go
@@ -91,7 +91,21 @@ func (ingestK *ingestKafka) kafkaListener() {
 
 }
 
-func processRecordDelay(record config.GenericMap) {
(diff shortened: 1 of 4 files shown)


pkg/pipeline/ingest/ingest_kafka.go
```diff
@@ -91,7 +91,21 @@ func (ingestK *ingestKafka) kafkaListener() {
 
 }
 
-func processRecordDelay(record config.GenericMap) {
+func (ingestK *ingestKafka) processPacketMetrics(record config.GenericMap) {
```
OlivierCazade, line 94: I had to create a dedicate function for kafka because the field is a float64 while it is an uint64 in the other ingest.

We need to investigate this, may be for all fields.

OlivierCazade:
Closing, we can implement this metrics by configuration only.

PR #287, opened 2022-08-16
conntrack: Add update reports for long connections

ronensc:
This PR adds a new type of output records to the connection tracking module: update report for long connections.

There's a new setting in the conntrack config `UpdateConnectionInterval` to set the amount of time to wait between update reports.

This PR also extends the connectionStore data structure from an ordered map to a multi-ordered map. The actual implementation of the data structure were extracted to a new file which simplifies `connectionStore`. This extension was needed to optimize the search for connections that require update report - rather than going through all the connections and inspecting their `nextUpdateReport` field, we keep them ordered by this field (in addition to the order by expiry time) and going through only those that require update report.

There is still 1 issue with the suggested implementation:

The scan for connections that need update report is done for every incoming batch of flow logs. This could be a problem when the incoming batch frequency is too high. A possible solution would be to scan once in a while. We can set the scan interval to be 50% of `UpdateConnectionInterval`.
**Note:** The same problem exists with end connections

Additional refactors and minor changes:
1. Instead of storing the `lastUpdate` of a connection, we store the `expiryTime` (which is `lastUpdate + ConnectionTimeout`). Changing the word "update" reduces confusion with the new feature of update reports. In my opinion, this also makes the expiry condition more understandable.
2. `connectionStore` was extracted to a file of its own.
3. `require.Equal()` was replaced with `require.JSONEq()` in `pkg/config/pipeline_builder_test.go`

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -195,6 +195,7 @@ Following is the supported API format for specifying connection tracking:
          outputRecordTypes: (enum) output record types to emit
              newConnection: New connection
              endConnection: End connection
+             updateConnection: Update connection
              flowLog: Flow log
          outputFields: list of output fields
                  name: output field name
@@ -206,4 +207,5 @@ Following is the supported API format for specifying connection tracking:
                  splitAB: When true, 2 output fields will be created. One for A->B and one for B->A flows.
                  input: The input field to base the operation on. When omitted, 'name' is used
          endConnectionTimeout: duration of time to wait from the last flow log to end a connection
+         updateConnectionInterval: duration of time to wait between update reports of a connection
 </pre>
--- a/pkg/pipeline/extract/conntrack/conntrack.go
+++ b/pkg/pipeline/extract/conntrack/conntrack.go
@@ -113,13 +113,11 @@ func (ct *conntrackImpl) Extract(flowLogs []config.GenericMap) []config.GenericM
 
 func (ct *conntrackImpl) popEndConnections() []config.GenericMap {
 	var outputRecords []config.GenericMap
-	// TODO: The following comment isn't correct. it's last update time rather than expired time.
-	// Iterate over the connections by their expired time from old to new.
+	// Iterate over the connections by their expiry time from old to new.
 	ct.connStore.iterateFrontToBack(expiryOrder, func(conn connection) (shouldDelete, shouldStop bool) {
-		expireTime := ct.clock.Now().Add(-ct.config.EndConnectionTimeout.Duration)
-		lastUpdate := conn.getLastUpdate()
-		if lastUpdate.Before(expireTime) {
-			// The last update time of this connection is too old. We want to pop it.
+		expiryTime := conn.getExpiryTime()
+		if ct.clock.Now().After(expiryTime) {
+			// The connection has expired. We want to pop it.
 			record := conn.toGenericMap()
 			addHashField(record, conn.getHash().hashTotal)
 			addTypeField(record, api.ConnTrackOutputRecordTypeName("EndConnection"))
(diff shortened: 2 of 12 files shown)


docs/api.md
```diff
@@ -20,9 +20,6 @@ Following is the supported API format for prometheus encode:
          prefix: prefix added to each metric name
          expiryTime: seconds of no-flow to wait before deleting prometheus data item
          tls: TLS configuration for the prometheus endpoint
-             enable: set to true to enable tls for the prometheus endpoint
```
eranra, line 23: why this is removed, should not be connected to this PR???

pkg/pipeline/extract/conntrack/multiorderedmap.go
```diff
@@ -0,0 +1,135 @@
+/*
```
eranra, line 1: @ronensc this feels like a generic helper utility, can you move that to `utils` ??

eranra (commented):
@ronensc can you update the README file ... currently it is still talking on the old implementation

pkg/pipeline/extract/conntrack/conntrack.go
```diff
@@ -143,6 +71,7 @@ func (ct *conntrackImpl) Extract(flowLogs []config.GenericMap) []config.GenericM
 				KeysFrom(fl, ct.config.KeyDefinition).
 				Aggregators(ct.aggregators).
 				Build()
+			conn.setNextUpdateReportTime(ct.clock.Now().Add(ct.config.UpdateConnectionInterval.Duration))
```
eranra, line 74: Do you want to add this as another `build` step instead of explicitaly calling the function

docs/api.md
ronensc: That field was changed from struct to struct ptr and the api-to-doc script doesn't handle such types at the moment.
I've submitted a different PR to solve this: https://github.com/netobserv/flowlogs-pipeline/pull/288

pkg/pipeline/extract/conntrack/multiorderedmap.go
ronensc: Done

pkg/pipeline/extract/conntrack/conntrack.go
ronensc: Done

pkg/pipeline/utils/multiorderedmap.go
```diff
@@ -0,0 +1,135 @@
...
+	delete(mom.m, key)
+}
+
+// MoveToBack moves the record of key `key` to the back of orderID. If the key or the orderID doesn't exist, an error
+// is returned.
+func (mom MultiOrderedMap) MoveToBack(key Key, orderID OrderID) error {
+	if _, found := mom.GetRecord(key); !found {
+		return fmt.Errorf("can't MoveToBack non-existing key %x (order id %q)", key, orderID)
+	}
+	rw := mom.m[key]
```
KalmanMeth, line 105: Shouldn't GetRecord() already give the element?

pkg/pipeline/utils/multiorderedmap_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// Test invalid args
+	err = mom.MoveToBack(0, lengthOrder)
+	require.Error(t, err)
+	err = mom.MoveToBack(5, "ERROR")
+	require.Error(t, err)
+	assertLengthConsistency(t, mom)
+}
+
+func TestMultiOrderedMap_IterateFrontToBackIterateFrontToBack(t *testing.T) {
```
KalmanMeth, line 100: Is the name IterateFrontToBack doubled intentionally or is it a copy-paste error?

pkg/pipeline/utils/multiorderedmap.go
ronensc: `GetRecord()` indeed returns the element. But, here were interested in the element's wrapper.
I'll remove the call to `GetRecord()` to make it clearer.

pkg/pipeline/utils/multiorderedmap_test.go
ronensc: it's an error. good catch!

KalmanMeth (approved):

jotak (approved):
LGTM - I guess it's a step toward the option to entirely remove raw flowlogs from storage in the future.

At some point I'd like to measure the performance impact of using conntracking:
- on FLP (I guess more memory used, more compute)
- on Loki if we don't store raw flowlogs, we save a lot of space

ronensc:
> * on Loki if we don't store raw flowlogs, we save a lot of space

@jotak In theory, this is correct. 
But, it depends on the characteristics of the connections in the cluster and the output records settings of the connection tracking module.
i.e. If most of the connections are short (starts and ends in the same flow-log) and the connection tracking is set to output both `new connection` and `end connection` records,
then, we'll end up with 2 records in loki for each connection, rather than 1 in the current settings without conntrack.

PR #288, opened 2022-08-17
Make apitodoc support Ptrs and update docs

ronensc:
This PR enhances the apitodoc script to parse struct pointers in the API such as in the following cases:

https://github.com/netobserv/flowlogs-pipeline/blob/9b96359be572c2b9d0fcf5620d5be3386f8a955d/pkg/api/ingest_kafka.go#L30

https://github.com/netobserv/flowlogs-pipeline/blob/9b96359be572c2b9d0fcf5620d5be3386f8a955d/pkg/api/encode_prom.go#L30

### DIFF
--- a/cmd/apitodoc/main.go
+++ b/cmd/apitodoc/main.go
@@ -30,7 +30,7 @@ import (
 func iterate(output io.Writer, data interface{}, indent int) {
 	newIndent := indent + 1
 	dataType := reflect.ValueOf(data).Kind()
-	//dataTypeName := reflect.ValueOf(data).Type().String()
+	// DEBUG code: dataTypeName := reflect.ValueOf(data).Type().String()
 	d := reflect.ValueOf(data)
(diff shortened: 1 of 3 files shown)


cmd/apitodoc/main.go
```diff
@@ -35,7 +35,7 @@ func iterate(output io.Writer, data interface{}, indent int) {
 	if dataType == reflect.Slice || dataType == reflect.Map {
```
KalmanMeth, line 35: There are some commented out statements in this function (lines 33, 44). Should these be removed?

KalmanMeth (approved):

cmd/apitodoc/main.go
ronensc: I didn't write them. But, it seems that line 33 is used in the commented out debug code in lines 36 and 41.
I don't see the point of line 44. I'll remove it.

PR #289, opened 2022-08-24
Add copyright notice to code files that lack it

ronensc:
I used the following command to find files that lack the copyright notice
```bash
grep -L -R --exclude-dir=vendor --include \*.go "Copyright (C)"
```

### DIFF
--- a/pkg/pipeline/decode/decode_protobuf.go
+++ b/pkg/pipeline/decode/decode_protobuf.go
@@ -1,3 +1,20 @@
+/*
+ * Copyright (C) 2022 IBM, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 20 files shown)


pkg/pipeline/decode/decode_protobuf.go
```diff
@@ -1,3 +1,20 @@
+/*
+ * Copyright (C) 2022 IBM, Inc.
```
KalmanMeth, line 2: Do we have to distinguish between files created and copyright by IBM vs Redhat?

ronensc: I don't know
cc @eranra @jotak @jpinsonneau @OlivierCazade @mariomac

KalmanMeth (approved):

Issue #280, opened 2022-08-03
Extend conntrack API to allow both splitted and unsplitted aggregation


ronensc:
Currently, the configuration of an output field of a connection supports a boolean flag `splitAB` to control whether to split the aggregation by direction.

For example, the following configuration:
```
outputFields:
  - name: Bytes
	operation: sum
	splitAB: false
```
will add a single `Bytes` field to a connection which sums the bytes of flows from A to B and from B to A.

On the other hand, the following configuration
```
outputFields:
  - name: Bytes
	operation: sum
	splitAB: true
```
will add two fields (`Bytes_AB` and `Bytes_BA`) to the connection which sums the bytes separately for A->B and B->A respectively.

If we want to allow both splitted and unsplitted sums, and end up with 3 additional fields in the connection (`Bytes`, `Bytes_AB` and `Bytes_AB`), we need to write:
```
outputFields:
  - name: Bytes
	operation: sum
	splitAB: true
  - name: Bytes
	operation: sum
	splitAB: false
```

We need to think if we want to make the configuration more compact and clear.
Also, do we want to allow aggregation of only one direction but not the other?
If so, 2 options come to mind:
1. Replace `splitAB` with 3 booleans:
```
outputFields:
  - name: Bytes
	operation: sum
	aggregateAB: true
	aggregateBA: true
	aggregateBoth: true
```

2. Replace `splitAB` with a list:
```
outputFields:
  - name: Bytes
	operation: sum
	aggregateBy: [AB, BA, both]
```

eranra:
@ronensc can we go with an enum approach that for the `emit` field (instead of splitAB) supports the following "AB" (only split), "Totals" (only totals), "AB&Totlas" (that emits both the split values and the totals). I think that splitting just on one of the sides is not very reasonable.

PR #291, opened 2022-08-25
conntrack: Add conntrack docs to README.md

ronensc:


README.md
```diff
@@ -533,6 +533,122 @@ No aggregate operation is needed and it should be set `raw_values`. The `valueKe
...
+  extract:
+    type: conntrack
+    conntrack:
+      endConnectionTimeout: 30s
+      updateConnectionInterval: 10s
+      keyDefinition:
+        fieldGroups:
+        - name: src
+          fields:
+          - SrcIP
```
KalmanMeth, line 582: Above you use SrcAddr rather than SrcIP

README.md
```diff
@@ -533,6 +533,122 @@ No aggregate operation is needed and it should be set `raw_values`. The `valueKe
...
+        hash:
+          fieldGroupRefs:
+          - protocol
+          fieldGroupARef: src
+          fieldGroupBRef: dst
+      outputRecordTypes:
+      - newConnection
+      - endConnection
+      - updateConnection
+      - flowLog
```
KalmanMeth, line 600: The `flowlog` parameter is not described above.

README.md
```diff
@@ -533,6 +533,122 @@ No aggregate operation is needed and it should be set `raw_values`. The `valueKe
...
+    "DstPort": 80,
+    "Proto": 17,
+    "Bytes": 300,
+    "TimeReceived": 1661430300
+}
+```
+
+Notice that all output records contain `_RecordType` and `_HashId` fields.
+Output fields that set `splitAB: true` (like in `Bytes`) are split into 2 fields `Bytes_AB` and `Bytes_BA` which 
+aggregate values separately based on direction A->B and B->A respectively.
```
jpinsonneau, line 651: and also adds a `XX_total` field ?

Is there other fields than `Bytes` ?

KalmanMeth (approved):

README.md
ronensc: > and also adds a `XX_total` field ?

The `Bytes_total` field is created because it was specified explicitly in the [config snippet.](https://github.com/netobserv/flowlogs-pipeline/pull/291/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R603-R605)
It's not auto-generated by the conntrack module. I created an issue to discuss this: https://github.com/netobserv/flowlogs-pipeline/issues/280



> Is there other fields than `Bytes` ?

In this config snippet, `Bytes` is the only field that is set with `splitAB: true`.
For all the other fields, the default is set to `false`.
I'll add a sentence for the default case.

jotak (approved):
lgtm

PR #295, opened 2022-09-06
Log as debug: "Can't find service name for..."

jotak:
Currently, FLP logs show loads of these "Can't find service name for Port 57430 and protocol 6" logs
Problem is that they don't really refer to issues, it's totally expected to have unmatched ports (especially when looking at source ports). The fact that this is logged within the processing pipeline loop make them unsuitable for `Info` level.
This PR moves them to Debug level.


jotak:
Also removes `err` from the second log, as it's always nil

jotak:
Also, if we want another strategy for reporting this kind of things that are generated in the pipeline loop, using metrics could be a better alternative than logging.

mariomac (approved):

ronensc (approved):

PR #297, opened 2022-09-07
change field names for extract_aggregate

KalmanMeth:
Both extract-aggregate and extract-timebased use an `Operation` and a `key` on which to perform the operation. This PR changes the usage in extract-aggregate to match the usage in extract-timebased so that the field on which the `Operation` is performed is called `operationKey`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -499,13 +499,13 @@ parameters:
             - "dstIP"
             - "srcIP"
           operation: "avg"
-          recordKey: "value"
+          operationKey: "value"
 ```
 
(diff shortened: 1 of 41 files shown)


KalmanMeth:
In extract-aggregate:
Change RecordKey to OperationKey
Change Operation To OperationType
Change By to GroupByKeys

In extract-timebased
Change Recordkey to IndexKey
Change Operation To OperationType

ronensc (approved):

KalmanMeth:
code rebase

PR #303, opened 2022-09-09
NETOBSERV-577 Few FLP optimizations

jotak:
- parallel transforms
- minor loki writer optimizations (pre-compute a couple of things)
- add loki benchmark
- add stage measurment metric

### DIFF
--- a/pkg/pipeline/transform/transform_network.go
+++ b/pkg/pipeline/transform/transform_network.go
@@ -23,6 +23,7 @@ import (
 	"net"
 	"regexp"
 	"strconv"
+	"sync"
 	"text/template"
 
 	"github.com/Knetic/govaluate"
@@ -40,9 +41,23 @@ type Network struct {
 }
 
 func (n *Network) Transform(input []config.GenericMap) []config.GenericMap {
+	entriesChan := make(chan config.GenericMap, len(input))
+	var wg sync.WaitGroup
+	wg.Add(len(input))
+
+	for _, e := range input {
+		go func(entry config.GenericMap) {
+			defer wg.Done()
+			outputEntry := n.TransformEntry(entry)
+			entriesChan <- outputEntry
+		}(e)
+	}
+
+	wg.Wait()
+	close(entriesChan)
+
 	outputEntries := make([]config.GenericMap, 0)
-	for _, entry := range input {
-		outputEntry := n.TransformEntry(entry)
+	for outputEntry := range entriesChan {
 		outputEntries = append(outputEntries, outputEntry)
 	}
 	return outputEntries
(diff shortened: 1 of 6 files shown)


pkg/pipeline/transform/transform_network.go
```diff
@@ -40,9 +41,23 @@ type Network struct {
...
+	entriesChan := make(chan config.GenericMap, len(input))
+	var wg sync.WaitGroup
+	wg.Add(len(input))
+
+	for _, e := range input {
+		go func(entry config.GenericMap) {
+			defer wg.Done()
+			outputEntry := n.TransformEntry(entry)
+			entriesChan <- outputEntry
+		}(e)
```
mariomac, line 53: I'm afraid that the goroutine creation/management and channel management costs could repel any parallelization improvement. TransformEntry is a short process, and spawning 10K goroutines to hold each of them in parallel could not be effective.

I'd rather divide the work in less tasks carrying a bigger portion of the work. E.g. carry a subslice on each goroutine:

```go
outputEntries := make([]config.GenericMap, len(input))
parallelTasks := runtime.GOMAXPROCS(0) // or halve it, or do user-configurable 
for t := 0; t < parallelTasks ; t++ {
    si := t * len(input) / parallelTasks
    ei := (t + 1) * len(input) / parallelTasks
    go func() {
        defer wg.Done()
        for i, entry := range input[si: ei] {
            outputEntries[i + si] = n.TransformEntry(entry)
        }
    }()
}
```

mariomac: ("10K goroutines", thinking in the use case of one of our users)

jotak: yeah ... "goroutines are cheap", but maybe not THAT cheap :)

jotak: anyway, the need for parallel processing was assuming there was blocking calls on the kube API, but if that's not the case (still unclear to me), we can forget this

mariomac (approved):

pkg/pipeline/write/write_loki_test.go
```diff
@@ -285,3 +291,57 @@ parameters:
...
+	}
+
+	i := 0
+	for range lokiFlows {
+		i++
+		if i == 100*b.N {
+			close(lokiFlows)
+			break
+		}
+	}
```
ronensc, line 346: What's the purpose of this loop?

ronensc (approved):

pkg/pipeline/write/write_loki_test.go
jotak: Reading on the channel is necessary to not block the flows processing. I haven't looked much further in `test.FakeLokiHandler`, I'd agree this code is not super elegant maybe something better can be done, i'll look more into it

PR #304, opened 2022-09-09
Optimize netdb Services databas

mariomac:
Benchmarks comparing new vs old version:

```
BenchmarkGetProtoByNumber-12            162093156               74.11 ns/op
BenchmarkGetProtoByName-12              93009231               126.4 ns/op

BenchmarkGetProtoByNumber_Old-12          945291             12846 ns/op
BenchmarkGetProtoByName_Old-12          17034988               706.8 ns/op
```

To reproduce all the benchmarks, you should checkout commit `5ac2502081c1a47101f758030b988466cbf554ec`, as the old code is removed in the head commit.


pkg/pipeline/transform/netdb/fixtures/etcProtocols.txt
```diff
@@ -0,0 +1,162 @@
+# /etc/protocols:
+# $Id: protocols,v 1.12 2016/07/08 12:27 ovasik Exp $
```
ronensc, line 2: As noted in `go help test`, I think `testdata` is more idiomatic name for the fixtures directory.

```
The go tool will ignore a directory named "testdata", making it available
to hold ancillary data needed by the tests.
```
https://pkg.go.dev/cmd/go/internal/test

pkg/pipeline/transform/netdb/netdb_test.go
```diff
@@ -0,0 +1,82 @@
...
+func TestServicesDB(t *testing.T) {
+	db, err := testingServicesDB()
+	require.NoError(t, err)
+
+	assert.Equal(t, "netbios-dgm", db.ByPortAndProtocolNumber(138, 6))
+	assert.Equal(t, "netbios-dgm", db.ByPortAndProtocolName(138, "tcp"))
+	// verify it also finds service name by protocol alias
+	assert.Equal(t, "netbios-dgm", db.ByPortAndProtocolName(138, "TCP"))
+
+	// verify multiple ports can be associated to the same protocol
```
ronensc, line 36: I don't understand this comment. The cases below all have the same port but different protocols. In contrast to the comment that states multiple ports and the same protocol.

pkg/pipeline/transform/transform_network.go
```diff
@@ -213,16 +217,27 @@ func NewTransformNetwork(params config.StageParam) (Transformer, error) {
...
 
+	var servicesDB *netdb.ServiceNames
 	if needToInitNetworkServices {
-		err := netdb.Init(jsonNetworkTransform.ProtocolsFile, jsonNetworkTransform.ServicesFile)
+		var err error
+		protos, err := os.Open("/etc/protocols")
+		if err != nil {
+			return nil, fmt.Errorf("opening /etc/protocols: %w", err)
+		}
+		services, err := os.Open("/etc/services")
```
ronensc, line 227: Shouldn't these `os.Open()` be followed by `defer file.Close()` as done in the tests?

ronensc, line 227: In the old implementation, the paths to the protocols file and the services file were configurable (`jsonNetworkTransform.ProtocolsFile` and `jsonNetworkTransform.ServicesFile`). Don't we want to keep them configurable (rather than hardcoded)?

ronensc (commented):
Great optimization! :)

pkg/pipeline/transform/netdb/fixtures/etcProtocols.txt
mariomac: Good point. Didn't know it. Thanks!

pkg/pipeline/transform/netdb/netdb_test.go
mariomac: yeah you are right. Fixing

PR #308, opened 2022-09-15
Allow enabling Go Pprof

mariomac:
If the `profile.port` argument is set to something different than 0, it will add a pprof endpoint so we can inspect what's going on FLP in terms of memory and CPU.


jotak:
/lgtm
but a short documentation on how to use it would be nice

mariomac:
@jotak I'll add the actual documentation in the NOO repo, as a new PR.

jotak (approved):

PR #312, opened 2022-09-22
Fix crash on startup

jotak:
"failed to initialize pipeline opening "": open : no such file or directory"
Previous PR removed the default file names for service mapping

### DIFF
--- a/cmd/flowlogs-pipeline/main_test.go
+++ b/cmd/flowlogs-pipeline/main_test.go
@@ -18,9 +18,14 @@
 package main
 
 import (
+	"encoding/json"
 	"os"
 	"os/exec"
 	"testing"
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline"
+	"github.com/stretchr/testify/require"
 )
 
 func TestTheMain(t *testing.T) {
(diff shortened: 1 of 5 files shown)


pkg/api/transform_network.go
```diff
@@ -24,6 +24,18 @@ type TransformNetwork struct {
 	ProtocolsFile  string                `yaml:"protocolsFile,omitempty" json:"protocolsFile,omitempty" doc:"path to protocols file (optional, default: /etc/protocols)"`
 }
 
+func (tn *TransformNetwork) GetServiceFiles() (string, string) {
+	p := tn.ProtocolsFile
+	if p == "" {
+		p = "/etc/protocols"
```
jotak, line 30: by the way, I used the previous default, as you can see here: https://github.com/netobserv/flowlogs-pipeline/pull/304/files#diff-803cd3d555b315c0ffdf0ac4905fd947085c6edf178137e7734e7af8a1218214L55

But maybe this isn't super safe to write in /etc/protocols? maybe rather use /tmp/flp-protocols and /tmp/flp-services instead? @KalmanMeth @ronensc @mariomac  ?

jotak: never mind, I thought it was a file that we would write into, I didn't know it was a standardized thingy

pkg/pipeline/transform/transform_network.go
```diff
@@ -189,15 +189,16 @@ func NewTransformNetwork(params config.StageParam) (Transformer, error) {
...
 		if err != nil {
-			return nil, fmt.Errorf("opening %q: %w", jsonNetworkTransform.ProtocolsFile, err)
+			return nil, fmt.Errorf("opening protocols file %q: %w", pFilename, err)
 		}
 		defer protos.Close()
-		services, err := os.Open(jsonNetworkTransform.ServicesFile)
+		services, err := os.Open(sFilename)
 		if err != nil {
-			return nil, fmt.Errorf("opening %q: %w", jsonNetworkTransform.ServicesFile, err)
+			return nil, fmt.Errorf("opening services file %q: %w", sFilename, err)
```
ronensc, line 201: +1 for improving the error message

mariomac (approved):

ronensc (approved):
Thanks for the fix!
It's weird that the e2e tests didn't catch it...

pkg/pipeline/transform/transform_network.go
ronensc: I was able to find the error message in the e2e test of the PR that introduced the bug. But, despite that, the test passes.
https://github.com/netobserv/flowlogs-pipeline/actions/runs/3082582328/jobs/4982452400
image

cmd/flowlogs-pipeline/main_test.go
```diff
@@ -36,3 +41,25 @@ func TestTheMain(t *testing.T) {
...
+    }
+}`
+	var opts config.Options
+	err := json.Unmarshal([]byte(js), &opts)
+	require.NoError(t, err)
+	cfg, err := config.ParseConfig(opts)
+	require.NoError(t, err)
+	require.NotNil(t, cfg)
+	mainPipeline, err := pipeline.NewPipeline(&cfg)
+	require.NoError(t, err)
```
mariomac, line 63: This test is failing here with:
```
can't access kubenetes. Tried using config from: config parameter, KUBECONFIG env, homedir and InClusterConfig. Got: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
```
I wonder if using `os.Setenv` is enough...

ronensc: Another option would be to remove the `add_kubernetes` rules from the test config.

mariomac: yeah, Ronen's suggestion is much more clean for a unit test

jotak: I'm rather mocking the kube client loading

jotak: (else, kube client initialization would fail without a real kube cluster available)

pkg/pipeline/transform/transform_network.go
jotak: yeah seems like there's some improvement to bring to the e2e test too

pkg/pipeline/transform/kubernetes/kubernetes.go
```diff
@@ -47,7 +47,13 @@ const (
 	typeService           = "Service"
 )
 
+type KubeDataInterface interface {
```
mariomac, line 50: nitpick: this interface can be private.

jotak: done (I just had to move the mock into kubernetes package :) )

mariomac (approved):

PR #313, opened 2022-09-22
NETOBSERV-589: fix Kafka protobuf decoding

mariomac:
Fixed few parts of kafka protobuf decoding:
* Decoder interface now accepts byte arrays instead of strings, so it can accept both JSONs and Protobuf messages
* Avoid logging raw protobuf messages, as they are actually binary and pollute the logs.

Two unneeded modules and vendored files have been automatically removed after running `go mod tidy` and `go mod vendor`.


pkg/api/ingest_kafka.go
```diff
@@ -25,7 +25,7 @@ type IngestKafka struct {
 	StartOffset       string     `yaml:"startOffset,omitempty" json:"startOffset,omitempty" doc:"FirstOffset (least recent - default) or LastOffset (most recent) offset available for a partition"`
 	BatchReadTimeout  int64      `yaml:"batchReadTimeout,omitempty" json:"batchReadTimeout,omitempty" doc:"how often (in milliseconds) to process input"`
 	Decoder           Decoder    `yaml:"decoder,omitempty" json:"decoder" doc:"decoder to use (E.g. json or protobuf)"`
-	BatchMaxLen       int        `yaml:"batchMaxLen,omitempty" json:"batchMaxLen,omitempty" doc:"the number of accumulated flows before being forwarded for processing"`
+	BatchMaxLen       int        `yaml:"batchMaxLen,omitempty" json:"batchMaxLen,omitempty" doc:"the number of accumulated flows before being forwarded for processing. Does not apply for Protobuf ingestion"`
```
OlivierCazade, line 28: BatchMaxLen is used internally to send to the next stage, so I don't understand why this does not apply to Protobuf.

Looking at the second, I think this parameter still apply.

mariomac: The `processJSONLines` (previously `processLogLines`) receives a single JSON flow on each Kafka message so it batchs multiple flows before submitting them to the next stage.

When protobuf is chosen, the above method is replaced by `processProtobufLines`, because each protobuf message already contains multiple flows so there is no need to accumulate them. So the `processProtobufLines` just receives the message, decodes it, and forwards all the received flows to the next pipeline stage.

OlivierCazade: Yes but this BatchMaxLen is not about batching between Kafka and FLP.

Once a flow has been received by the ingester, this flow is stored until BatchMaxLen is reached or BatchReadTimout is reached.

This is not receiving flow per batch, this is sending flow to the next stage in batch.

mariomac: Yes, you are right, but observe that, when using Kafka, now the ingester behaves differently: https://github.com/netobserv/flowlogs-pipeline/pull/313/files#diff-7a58c31ce0acf1277ffd5e45a5e43a1faf33cf04e70800a348721d0c21beb4e8R174-R192

It just receives the batch of flows from Kafka directly as a batch, so they are decoded and forwarded directly to the next stage in the pipeline.

Why? There is no need to store the flows if they have been already received as a big batch. This will also avoid adding an extra delay to the pipeline.

mariomac: Done

OlivierCazade: Thanks!

OlivierCazade (approved):

PR #314, opened 2022-09-23
NETOBSERV-579 prefixing all metrics, add new ones

jotak:
- Allow to prefix operational metrics (it introduces a global settings
  for metrics, which will probably be extended later)
- Unify metrics usage across stages: e.g. channel size was tracked only
  in netflow ingester; now they share a common struct that generate
metrics. More metrics now include the "stage" label.
- Adding new gauges for tracking channel sizes
- Some renaming
- Updated metrics doc

Some refactoring to make doc generation work with metrics that are not defined as globals (need to decouple metrics definition from its instantiation in prom registry) . Update doc to include labels, and sort alphabetically.

Also I had to move health & operationalMetrics into the same package to avoid repetitions like `operationalmetrics.Metrics` , which resulted also in move health_test into the pipeline package (to avoid cycle)

**Breaking changes**: the breaking changes are only about metric names (not API), see below

### DIFF
--- a/pkg/pipeline/encode/encode_prom.go
+++ b/pkg/pipeline/encode/encode_prom.go
@@ -196,10 +196,14 @@ func (e *EncodeProm) prepareAggHisto(flow config.GenericMap, info *api.PromMetri
 
 func (e *EncodeProm) extractGenericValue(flow config.GenericMap, info *api.PromMetricsItem) interface{} {
 	if info.Filter.Key != "" {
-		val, found := flow[info.Filter.Key]
-		shouldKeepRecord := found && val == info.Filter.Value
-		if !shouldKeepRecord {
-			return nil
+		if val, found := flow[info.Filter.Key]; found {
+			sVal, ok := val.(string)
+			if !ok {
+				sVal = fmt.Sprint(val)
+			}
+			if sVal != info.Filter.Value {
+				return nil
+			}
 		}
 	}
 	if info.ValueKey == "" {
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/mariomac/guara v0.0.0-20220523124851-5fc279816f1f
 	github.com/mitchellh/mapstructure v1.4.3
 	github.com/netobserv/gopipes v0.1.1
-	github.com/netobserv/loki-client-go v0.0.0-20211018150932-cb17208397a9
+	github.com/netobserv/loki-client-go v0.0.0-20220927092034-f37122a54500
 	github.com/netobserv/netobserv-ebpf-agent v0.1.1-0.20220608092850-3fd4695b7cc2
 	github.com/netsampler/goflow2 v1.1.1-0.20220509155230-5300494e4785
 	github.com/pkg/errors v0.9.1
--- a/go.sum
+++ b/go.sum
@@ -701,6 +701,8 @@ github.com/netobserv/gopipes v0.1.1 h1:f8zJsvnMgRFRa2B+1siwRtW0Y4dqeBROmkcI/HgT1
 github.com/netobserv/gopipes v0.1.1/go.mod h1:eGoHZW1ON8Dx/zmDXUhsbVNqatPjtpdO0UZBmGZGmVI=
 github.com/netobserv/loki-client-go v0.0.0-20211018150932-cb17208397a9 h1:c2swm3EamzgjBq9idNbEs5bNz20FJo/HK6uxyigXekQ=
 github.com/netobserv/loki-client-go v0.0.0-20211018150932-cb17208397a9/go.mod h1:LHXpc5tjKvsfZn0pwLKrvlgEhZcCaw3Di9mUEZGAI4E=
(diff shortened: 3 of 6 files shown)


jotak:
Here's the details of the new metrics:

### conntrack_input_records (unchanged)
| **Name** | conntrack_input_records | 
|:---|:---|
| **Description** | The total number of input records per classification. | 
| **Type** | counter | 
| **Labels** | classification | 


### conntrack_memory_connections (unchanged)
| **Name** | conntrack_memory_connections | 
|:---|:---|
| **Description** | The total number of tracked connections in memory. | 
| **Type** | gauge | 
| **Labels** |  | 


### conntrack_output_records (unchanged)
| **Name** | conntrack_output_records | 
|:---|:---|
| **Description** | The total number of output records. | 
| **Type** | counter | 
| **Labels** | type | 


### encode_prom_errors (unchanged)
| **Name** | encode_prom_errors | 
|:---|:---|
| **Description** | Total errors during metrics generation | 
| **Type** | counter | 
| **Labels** | error, metric, key | 


### ingest_flows_processed (renamed from "ingest_collector_flow_logs_processed", added stage label)
| **Name** | ingest_flows_processed | 
|:---|:---|
| **Description** | Number of flow logs processed by the ingester | 
| **Type** | counter | 
| **Labels** | stage | 


### metrics_processed (renamed from "encode_prom_metrics_processed", added stage label)
| **Name** | metrics_processed | 
|:---|:---|
| **Description** | Number of metrics processed | 
| **Type** | counter | 
| **Labels** | stage | 


### records_written (renamed from "loki_records_written", added stage label, now works also with kafka write)
| **Name** | records_written | 
|:---|:---|
| **Description** | Number of output records written | 
| **Type** | counter | 
| **Labels** | stage | 


### stage_duration_ms (unchanged)
| **Name** | stage_duration_ms | 
|:---|:---|
| **Description** | Pipeline stage duration in milliseconds | 
| **Type** | histogram | 
| **Labels** | stage | 


### stage_in_queue_size (renamed from ingest_collector_queue_length, generalized for all stages)
| **Name** | stage_in_queue_size | 
|:---|:---|
| **Description** | Pipeline stage input queue size (number of elements in queue) | 
| **Type** | gauge | 
| **Labels** | stage | 


### stage_out_queue_size (new metric)
| **Name** | stage_out_queue_size | 
|:---|:---|
| **Description** | Pipeline stage output queue size (number of elements in queue) | 
| **Type** | gauge | 
| **Labels** | stage |

pkg/pipeline/write/write_loki.go
```diff
@@ -271,7 +266,9 @@ func NewWriteLoki(params config.StageParam) (*Loki, error) {
 		}
 	}
 
+	// TODO / FIXME / FIGUREOUT: seems like we have 2 input channels for Loki? (this one, and see also pipeline_builder.go / getStageNode / StageWrite)
```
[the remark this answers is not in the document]
mariomac: This channel seems to act as a buffer to avoid blocking the previous pipeline. Probably we'd need to verify that removing the `in` channel wouldn't affect performance.

When we are able to use Go 1.18, maybe we could upgrade to an [upper-upperstream version of the go-pipes](https://github.com/mariomac/pipes) library that allows defining the buffer length between pipeline stages.

pkg/pipeline/encode/encode_prom.go
```diff
@@ -196,10 +196,14 @@ func (e *EncodeProm) prepareAggHisto(flow config.GenericMap, info *api.PromMetri
...
 func (e *EncodeProm) extractGenericValue(flow config.GenericMap, info *api.PromMetricsItem) interface{} {
 	if info.Filter.Key != "" {
-		val, found := flow[info.Filter.Key]
-		shouldKeepRecord := found && val == info.Filter.Value
-		if !shouldKeepRecord {
-			return nil
+		if val, found := flow[info.Filter.Key]; found {
+			sVal, ok := val.(string)
+			if !ok {
+				sVal = fmt.Sprintf("%v", val)
```
mariomac, line 202: Nit: maybe `fmt.Sprint(val)` is slightly faster.

mariomac (approved):

pkg/pipeline/encode/encode_prom.go
jotak: done

OlivierCazade (approved):

pkg/operational/metrics.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sirupsen/logrus"
+)
+
+type MetricDefinition struct {
+	Name   string
+	Help   string
+	Type   metricType
+	Labels []string
+}
```
ronensc, line 35: @jotak IIUC, `Type metricType` is only used by the auto-documentation.
So, theoretically, one can create a `MetricDefinition` of type `Counter` and pass it to `NewGauge()` causing a misleading documentation.
If I'm correct, is it worth adding a check on `NewGauge()/NewCounter()/NewHistogram()` to validate the type?

jotak: that's correct, sure we can do what you suggest

PR #316, opened 2022-09-28
Unify ingest metrics too

jotak:
Except for netflow/ipfix which continue to use goflow2 metrics, other ingester now adopt the same "framework" for metrics as the other stages, hence also have new names with a prefix

New metrics are:

### ingest_errors
| **Name** | ingest_errors | 
|:---|:---|
| **Description** | Counter of errors during ingestion | 
| **Type** | counter | 
| **Labels** | stage, type, code | 

### ingest_flows_processed
| **Name** | ingest_flows_processed | 
|:---|:---|
| **Description** | Provides number of flows processed, batches processed, and batch size stats | 
| **Type** | summary | 
| **Labels** | stage | 

### ingest_latency_ms
| **Name** | ingest_latency_ms | 
|:---|:---|
| **Description** | Latency between flow end time and ingest time, in milliseconds | 
| **Type** | histogram | 
| **Labels** | stage | 

### ingest_packets_size_bytes
| **Name** | ingest_packets_size_bytes | 
|:---|:---|
| **Description** | Ingested packets size distribution, in bytes | 
| **Type** | summary | 
| **Labels** | stage |

### DIFF
--- a/pkg/pipeline/ingest/ingest_grpc.go
+++ b/pkg/pipeline/ingest/ingest_grpc.go
@@ -3,9 +3,6 @@ package ingest
 import (
 	"context"
 	"fmt"
-	"strconv"
-	"strings"
-	"time"
 
 	"github.com/netobserv/flowlogs-pipeline/pkg/api"
 	"github.com/netobserv/flowlogs-pipeline/pkg/config"
@@ -14,28 +11,14 @@ import (
 	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/utils"
 	"github.com/netobserv/netobserv-ebpf-agent/pkg/grpc"
 	"github.com/netobserv/netobserv-ebpf-agent/pkg/pbflow"
-	flow "github.com/netsampler/goflow2/utils"
-	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 	grpc2 "google.golang.org/grpc"
-	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 )
 
 const (
 	defaultBufferLen = 100
-	decoderName      = "protobuf"
-	decoderVersion   = "protobuf1.0"
-)
-
-// Prometheus metrics describing the performance of the eBPF ingest
-// This metrics are internal to the goflow2 library, we update them here manually to align with the IPFIX ingester
-var (
-	flowDecoderCount    = flow.DecoderStats
-	processDelaySummary = flow.NetFlowTimeStatsSum.With(
-		prometheus.Labels{"version": decoderVersion, "router": ""})
-	flowTrafficBytesSum = flow.MetricPacketSizeSum
-	flowErrors          = flow.NetFlowErrors
 )
 
 // GRPCProtobuf ingests data from the NetObserv eBPF Agent, using Protocol Buffers over gRPC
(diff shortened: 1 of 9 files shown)


jotak:
cc @memodi @Amoghrd

pkg/pipeline/ingest/ingest_grpc.go
```diff
@@ -109,41 +91,23 @@ func instrumentGRPC(port int, counter func(int)) grpc2.UnaryServerInterceptor {
...
-			}
-		}
-		trafficLabels := prometheus.Labels{
-			"type":       "protobuf",
-			"remote_ip":  remoteIP,
-			"local_ip":   "0.0.0.0",
-			"local_port": localPort,
-		}
+		// instrument batch size distribution (which also instruments total flows counter under the hood)
+		m.batchSize.Observe(float64(len(flowRecords.Entries)))
```
jotak, line 98: For information, I removed a couple of labels that I don't think are super useful, and it simplifies the code & process (no need to extract a remote IP from metadata)

Let me know if you think we should keep these labels (type, remote_ip,local_ip, local_port)

jotak:
@memodi answers below:

> @jotak - couple of questions:
> 
>     1. is ingestion and flow processing a separate stages in FLP? could there be a scenario where there are delays in flow processing since the time its ingested?

Ingestion is the initial stage, followed by "enrichment" and then "writes" (to loki and prometheus), roughly. "Flow processing" is not a stage per se.. I would say that every stages do some flow processing.

So yes, there can be delays after the ingestion, and there's a metric for that: `netobserv_stage_duration_ms` with a `stage` label.

To sum up:
- `netobserv_ingest_latency_ms` accounts for delay before ingestion, from the FlowEndTime (e.g. could be delay within the eBPF agent, network delay between agent and FLP, or kafka if it is involved)
- `netobserv_stage_duration_ms` is for the processing time inside FLP, for each individual stage

>     2. does `ingest_packets_size_bytes` denotes summary for pre-aggregated packet sizes ? or is it the summary for size _after_ aggregating packets into flows?

Good question and I realize the name is not really correct, it doesn't represent packets. It's the size of the batches sent by agents to FLP. It does not represent the observed cluster traffic, but the size of the flows themselves, potentially batched.

jotak:
@mariomac if you think we need to modify some metrics to make it easier for trying to remove batching between stages, let me know

mariomac (approved):

pkg/pipeline/ingest/metrics.go
```diff
@@ -6,21 +6,69 @@ import (
...
 	return &metrics{
 		Metrics:        opMetrics,
 		stage:          stage,
-		flowsProcessed: opMetrics.CreateFlowsProcessedCounter(stage),
+		stageType:      stageType,
+		latency:        opMetrics.NewHistogram(&latencyHistogram, []float64{.001, .01, .1, 1, 10, 100, 1000, 10000}, stage),
+		stageDuration:  opMetrics.GetOrCreateStageDurationHisto().WithLabelValues(stage),
+		batchSize:      opMetrics.NewSummary(&batchSizeSummary, stage),
+		batchSizeBytes: opMetrics.NewSummary(&batchSizeBytesSummary, stage),
+		errors:         opMetrics.NewCounterVec(&errorsCounter),
```
ronensc, line 57: What's the thumb rule for choosing Summary over Histogram?
According to [Histograms and summaries](https://prometheus.io/docs/practices/histograms/), considering client performance, summaries are more expensive:
```
Observations are expensive due to the streaming quantile calculation.
```

jotak: I don't know if there's a rule of thumb, I tend to prefer histograms also because they provide more flexibility in queries.
On the flip side, summaries tend to generate less metrics (although it depends on configuration, but that's generally speaking) which is also something the prometheus server appreciates.
But TBH, here I just didn't change their types: goflow2 uses summaries which, previously, were reused in ingest_grpc.

PR #317, opened 2022-09-29
NETOBSERV-578: decrease informers' memory consumption

mariomac:
This PR adds a Transformer for each Kubernetes informer that converts the watched Pods, Services, and Nodes to `Info` instances, that are stored by the informers cache and contain only the data that is required for the decoration.

This also alleviates the load on the Garbage Collector: before this patch, a new `Info` instance was created for each flow decoration. Now the same objects are reused on `GetInfo` invocations for the same IP.

In Flowlogs-Pipelines with mid-to-low loads, no improvements have been observed in terms of CPU and slight improvements in terms of memory. In the image below, the middle part corresponds to the new version of FLP.

image

### DIFF
--- a/pkg/pipeline/transform/kubernetes/kubernetes.go
+++ b/pkg/pipeline/transform/kubernetes/kubernetes.go
@@ -18,6 +18,7 @@
 package kubernetes
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"os"
@@ -29,6 +30,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -36,7 +38,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
-var Data kubeDataInterface = &KubeData{}
+var Data kubeDataInterface = &KubeData{ips: map[string]*Info{}}
 
 const (
 	kubeConfigEnvVariable = "KUBECONFIG"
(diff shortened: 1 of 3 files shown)


jotak:
I did some tests on higher load with `hey-ho`, I also see the memory usage being improved, of a slightly better amount that what you show:
- Spiking at 1.16GB (versus 1.39 before) => -16%
- Stabilizing at 642MB (versus 423 before) => -34%

jotak:
@mariomac I haven't reviewed the code in deep, but remember the discussion we had about potential blocking calls and fact the cache seemed pre-loaded? Does this PR change anything in that regard / should we try to parallelize transformer processing?

mariomac:
@jotak after working deeply with the informers' code, I can confirm that there aren't blocking calls (neither before nor now).

The initial consideration about parallelizing workers was to share the big amount of memory that the informers' spent so we expected to minimize the overall memory usage by scaling vertically and downscaling horizontally.

There are some reasons that changed my mind with respect to parallelizing (I'll annotate them also in the related JIRA to justify it):

* The memory spent by the Informers' cache does not look to be so critical as we initially thought. And after this PR, the memory usage will be noticeably lower.

* Since the flows arrive as waves and are internally processed in batches, I'm afraid that the effect of parallelism will be limited.

Given the internal complexity of FLP, I'd say that the effort of reworking some internals to allow this parallelization won't worth the limited improvements we could achieve.

pkg/pipeline/transform/kubernetes/kubernetes.go
```diff
@@ -54,180 +55,254 @@ type kubeDataInterface interface {
...
 }
 
 func (k *KubeData) getOwner(info *Info) Owner {
-	if info.OwnerReferences != nil && len(info.OwnerReferences) > 0 {
+	if len(info.OwnerReferences) != 0 {
 		ownerReference := info.OwnerReferences[0]
-		if ownerReference.Kind == "ReplicaSet" {
-			item, ok, err := k.replicaSetInformer.GetIndexer().GetByKey(info.Namespace + "/" + ownerReference.Name)
-			if err != nil {
-				panic(err)
```
jotak, line 127: never noticed this `panic` .. that was abrupt!

jotak (commented):
Code lgtm, smart rewrite! thanks @mariomac 
I'll do a couple of more tests on my cluster before approving

jotak (approved):
lgtm
I smoke-tested the enrichment end to end, everything seems fine

PR #318, opened 2022-09-30
disable CGO

mariomac:
CGO it's not required and usually makes executables more resource-consuming. It also adds some extra requirements to the build/test pipeline (e.g. having to have GCC + libraries, etc...)

Reference read: https://dave.cheney.net/2016/01/18/cgo-is-not-go

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,6 @@ export GO111MODULE=on
 export CGO_ENABLED=0
 export GOOS=linux
 export GOARCH=amd64
-export CGO_ENABLED=1
 
 SHELL := /bin/bash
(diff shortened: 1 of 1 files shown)


KalmanMeth (approved):

KalmanMeth:
@ronensc Probably as @mariomac wrote, it is needed for -race.

mariomac:
Yes, it was needed for `-race` (didn't know that) so I put it directly in the tests.

However I'm doing few more tests before merging. Despite everybody saying CGo is slower than Go in the internet, I noticed a minor increase in the CPU&Memory performance 🤷🏻‍♂️

PR #319, opened 2022-10-03
NETOBSERV-578: removing arrays from pipeline interfaces

mariomac:
This PR removes the batching logic in the Kafka ingester and processes flows 1 by 1 on each pipeline stage.

While the CPU usage doesn't seem to noticeably decrease:

image

It greatly improves memory consumption:

image

(as well as simplifies some parts of the code)


pkg/pipeline/transform/transform_network.go
```diff
@@ -38,23 +38,14 @@ type Network struct {
...
-
-func (n *Network) TransformEntry(inputEntry config.GenericMap) config.GenericMap {
+func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bool) {
 	// copy input entry before transform to avoid alteration on parallel stages
 	outputEntry := inputEntry.Copy()
 
 	// TODO: for efficiency and maintainability, maybe each case in the switch below should be an individual implementation of Transformer
 	for _, rule := range n.Rules {
 		switch rule.Type {
-		case api.TransformNetworkOperationName("AddRegExIf"):
```
mariomac, line 57: Logic behind this change is that the result of this invocation is known at runtime while the value of `api.OpAddRegexIf` is known at compile time.

That allows a much more efficient implementation of `switch`, especially important when this is invoked for each flow, for each rule.

pkg/pipeline/ingest/metrics.go
```diff
@@ -13,16 +13,16 @@ var (
...
-		"Provides number of flows processed, batches processed, and batch size stats (in number of flows)",
-		operational.TypeSummary,
+		"Provides number of flows processed",
+		operational.TypeCounter,
 		"stage",
 	)
-	batchSizeBytesSummary = operational.DefineMetric(
-		"ingest_batch_size_bytes",
-		"Ingested batch size distribution, in bytes",
-		operational.TypeSummary,
```
mariomac, line 25: @jotak This PR reverts part of your previous work. Since we don't batch anymore at ingest time.

Another option would be keep this batchSize summary and apply it at the GRPC interceptor, so we know the batch sizes from the client. For the kafka use case, it would happen that the summary reports that every batch is size 1.

mariomac: (other option, just keep different metrics: `flowsProcessed`, `ingestBytes` for kafka ingest and `batchSizeSummary`, `batchSizeBytesSummary` for GRPC ingest.

jotak: The idea behind these metrics was to get stats like knowing the effective size of agent's batches. As you said, it doesn't work with kafka because `kafkago` hides this detail, but that's still interesting for GRPC.
Since it's measuring ingesters input and not their output, I think removing the batches on output shouldn't affect them..

About using different metric names with or without kafka, the drawback is that it's harder to consume: we need to create more prom queries / grafana dashboards. So we need anyway to do a compromise somewhere. Personally I'd like to stick with the previous metrics.

mariomac: Sounds reasonable. Bringing them back.

mariomac:
It seems this PR is not possible unless we reimplement the connection tracking stage.

@ronensc do you have an estimation about how difficult would be to reimplement the function:
```go
func (ct *conntrackImpl) Extract(flowLogs []config.GenericMap) []config.GenericMap
```
To:
```go
func (ct *conntrackImpl) Extract(flowLogs config.GenericMap) config.GenericMap
```
Currently it seems it's not able to detect endConnections

KalmanMeth:
Not only the connection tracking, but also the aggregation and timebased-topk would need changes. In each of these we save some state and do some computation after batches of flow logs. If now the flow logs come one a time, we have to check that our implementation only runs every so often and not after each call to Extract().

mariomac:
Ok. I'll close this PR at the moment. I noticed similar improvements in memory just by setting the PullQueueCapacity to 50-100.

mariomac:
@jotak @ronensc @KalmanMeth @eranra reopening this PR as I found a compromise solution. I added an intermediate batcher for these parts of the API that really need to work in batches, like the connection tracking and the timebased-topk.

This would improve overall performance and memory but specially in those FLP configurations that do not use extractors. At the same time, this gives us more room for a later reimplementation of extractors to avoid batching.

pkg/pipeline/pipeline_builder.go
```diff
@@ -260,24 +281,31 @@ func (b *builder) getStageNode(pe *pipelineEntry, stageID string) (interface{},
...
 			b.opMetrics.CreateInQueueSizeGauge(stageID, func() int { return len(in) })
 			b.opMetrics.CreateOutQueueSizeGauge(stageID, func() int { return len(out) })
-			for i := range in {
-				b.runMeasured(stageID, func() {
-					out <- pe.Extractor.Extract(i)
-				})
-			}
+			// TODO: replace batcher by rewriting the different extractor implementations
+			// to keep the status while processing flows one by one
+			utils.Batcher(utils.ExitChannel(), b.batchMaxLen, b.batchTimeout, in,
```
mariomac, line 301: Didn't provide unit tests for Batcher as it is already tested as an internal component in the conntrack integration tests.

jpinsonneau (approved):
LGTM, currently deployed and working as expected on LLC cluster

I'll let it run to compare after some hours :+1:  
Thanks @mariomac

pkg/pipeline/pipeline_builder.go
KalmanMeth: Nice idea to build the Batcher in this way for all the stages that need it!

PR #320, opened 2022-10-04
Optimize Network Transformer switch

mariomac:
The api.TransformNetworkOperationName function is not known at compile time while the `api.Op...` constants are.

This gives room to the Go compiler for implementing the `switch` more efficiently (e.g. lookup table vs if-elses...). This is specially important in the network_transformer, which lookups multiple rules for each flow.


mariomac:
Closing, at this change is already done in https://github.com/netobserv/flowlogs-pipeline/pull/319

PR #321, opened 2022-10-05
Remove namespace for nodes

jotak:
This is a minor change that maybe will have 0 impact.. But I've found a weird case where node's flows are sometimes assigned an empty string namespace, and other times namespace is undefined. This is very weird and I don't know why there is this discrepancy, maybe something wrong on the loki side? (since this field is used for indexing).

It ends up duplicating metrics in the console, and make it harder to work with in the console.

### DIFF
--- a/pkg/pipeline/transform/kubernetes/kubernetes.go
+++ b/pkg/pipeline/transform/kubernetes/kubernetes.go
@@ -194,7 +194,7 @@ func (k *KubeData) initNodeInformer(informerFactory informers.SharedInformerFact
 		return &Info{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      node.Name,
-				Namespace: node.Namespace,
+				Namespace: "",
 				Labels:    node.Labels,
 			},
 			ips:  ips,


pkg/pipeline/transform/kubernetes/kubernetes.go
```diff
@@ -194,7 +194,7 @@ func (k *KubeData) initNodeInformer(informerFactory informers.SharedInformerFact
 		return &Info{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      node.Name,
-				Namespace: node.Namespace,
+				Namespace: "",
```
mariomac, line 197: Nitpick: just remove the line and Go by default assigns the empty string.
```suggestion
```

mariomac (approved):

pkg/pipeline/transform/kubernetes/kubernetes.go
jotak: I'm not 100% sure of that, when it comes to json serialization, empty values != zero-values , right?

jotak: (I've been tricked by this oddity quite a few times in the past!)

mariomac: In Go, there is no difference between zero value and empty value. In the case of string, the empty string is the zero value.

Maybe you got this issue when serializing nil pointers vs pointers to empty struct.

jotak: It's this kind of things:

```go
	var slice []string
	js, _ := json.Marshal(slice)
	fmt.Println(string(js)) // prints null

	slice = []string{}
	js, _ = json.Marshal(slice)
	fmt.Println(string(js)) // prints []
```

Maybe this is not about zero-value, but about golang being very permissive on nil slices, acting as if it was an empty slice most of the time.

But apparently you're right, for string there's no such issue

PR #322, opened 2022-10-07
NETOBSERV-633 Restore ingest_flows_processed as a simple counter

jotak:
This removes it as a summary of batch sizes
Batch sizes can however still be figured out, via "Flows processed / Batch size in bytes count"

@memodi this PR restores `netobserv_ingest_flows_processed` as it was before, so it should fix your grafana dashboard without you having anything to do


mariomac (approved):

memodi:
thanks @jotak for restoring this.

PR #323, opened 2022-10-07
Allow defining buffer length of internal nodes

mariomac:
Updated to version 0.2.0 of go-pipes and allow defining the length of the internal buffers between pipeline stages.


OlivierCazade (approved):
LGTM

PR #324, opened 2022-10-10
Speedup Loki JSON serialization

mariomac:
Changing the JSON encoder of Loki:

* Do not sort fields alphabetically
* Do not escape HTML (it still escapes JSON strings)

The improvements in terms of CPU are sensible:

image

### DIFF
--- a/pkg/pipeline/write/write_loki.go
+++ b/pkg/pipeline/write/write_loki.go
@@ -37,6 +37,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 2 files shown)


jotak (approved):
lgtm

ronensc:
Good catch!
@mariomac regarding the CPU plot, how was it created?
What does the blue and yellow lines mean? I assume the blue is CPU before the change and yellow is after (just wanted to make sure)

mariomac:
Hi @ronensc the plot was just the pod CPU consumption as presented by the Openshift UI dashboards' section. Your assumption is correct: blue is before and yellow is after.

PR #327, opened 2022-10-17
Remove unnecessary goroutine of loki writer

OlivierCazade:
Loki write use a goroutine to do async send to loki.

This is unnecessary because loki client is already doing async send.

I tested this change under some load and did not see any impact.


mariomac (approved):

ronensc (approved):

jotak (changes requested):
L256 there's this comment that I wrote: `// TODO / FIXME / FIGUREOUT bla bla` => it should be removed, as well as the channel creation and the associated gauge

jotak (approved):
thanks @OlivierCazade 
lgtm

Issue #298, opened 2022-09-07
Reduce connection tracking records


ronensc:
Assuming the connection tracking is configured to output these record types:
- `newConnection`
- `updateConnection`
- `endConnection`

It means that for each connection, we'll have at least 2 records (`newConnection` and `endConnection`).
One option to resolve this, is to remove `newConnection` from the output record type list.
The problem with that is that we don't have an easy way to follow new connections. I mean, for long connections, the first record will be `updateConnection`. While for short connections, the first and only record will be `endConnection`.
Looking only at `endConnection` records isn't great because if the cluster has many long connections that haven't ended yet, we won't count them.

A possible solution for this new problem could be adding a new boolean field to the connection record: `_IsFirst`. The field indicates whether the connection record (either `updateConnection` or `endConnection`) is the first report of the connection or not.

This issue follows this comment: https://github.com/netobserv/flowlogs-pipeline/pull/287#issuecomment-1238985305

cc @jotak @eranra

KalmanMeth:
completed

PR #328, opened 2022-10-18
conntrack: Add `_IsFirst` field

ronensc:
This PR adds the `_IsFirst` field to connection records (`newConnection`, `updateConnection` and `endConnection`).
It's not added to `flowLog` records.
The field is set to true only on the first record of the connection.
It is useful in cases where `newConnection` records are not outputted (to reduce the number output records)
and there is a need to count the total number of connections: simply counting `_IsFirst=true`

This solves #298

### DIFF
--- a/pkg/pipeline/extract/conntrack/conntrack_test.go
+++ b/pkg/pipeline/extract/conntrack/conntrack_test.go
@@ -33,7 +33,7 @@ import (
 
 var opMetrics = operational.NewMetrics(&config.MetricsSettings{})
 
-func buildMockConnTrackConfigFull(isBidirectional bool, outputRecordType []string,
+func buildMockConnTrackConfig(isBidirectional bool, outputRecordType []string,
 	updateConnectionInterval, endConnectionTimeout time.Duration) *config.StageParam {
 	splitAB := isBidirectional
 	var hash api.ConnTrackHash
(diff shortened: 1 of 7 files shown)


KalmanMeth:
I don't understand how this helps. By counting the number of  `_IsFirst=true` we get the total number of newConnections from the beginning of measurement. So this is essentially equivalent to counting the times we get newConnection (if we get each one). Can't we continue to count newConnection, and allow to have a newConnection reported on the first output record (even if we also have endConnection and/or updateConnection reported on the same record)?

ronensc:
@kalman When the `outputRecordTypes` is configured with `[newConnection, updateConnection, endConnection]`, i.e. to output `newConnection` records, then the `_IsFirst` flag will be set to true only on them. So counting `_IsFirst="true"` is equivalent to counting `_RecordType="newConnection"` as you mentioned.
But, when the `outputRecordTypes` is configured with `[updateConnection, endConnection]`, i.e. to not output `newConnection` records, then, counting `_IsFirst="true"` is my suggest way to count the number of connections. Counting `_RecordType="newConnection"` will return zero.

pkg/pipeline/extract/conntrack/conntrack_test.go
```diff
@@ -482,14 +491,161 @@ func TestUpdateConn_Unidirectional(t *testing.T) {
...
+			"11s: update report AB (with isFirst=true)",
+			startTime.Add(11 * time.Second),
+			nil,
+			[]config.GenericMap{
+				newMockRecordUpdateConn(ipA, portA, ipB, portB, protocol, 111, 11, 1).withHash(hashIdAB).markFirst().get(),
+			},
+		},
+		{
+			"20s: no update report",
+			startTime.Add(20 * time.Second),
```
KalmanMeth, line 553: Perhaps this should be 19 or less. 20 seems too borderline and it might just get an additional update report.

KalmanMeth (approved):

pkg/pipeline/extract/conntrack/conntrack_test.go
ronensc: Thanks for taking a close look at the tests!
Since the update report was at time `11s`, the next update should be at time `21s`. So `21s` is the borderline. That's why I check a second before (at `20s`) that there are no updates and a second after (at `22s`) that there is an update.

jotak (approved):
lgtm!

PR #330, opened 2022-10-28
(wip) improve efficiency of network transformer

mariomac:


jotak:
@mariomac can you describe the changes that you did?
the diff isn't easy to read, so it would help

mariomac:
@jotak ah I didn't realize this PR was still open. It was just a PoC for a pretended performance improvement but I wasn't able to measure any actual improvement so I forgot about it (that's why there is no description in the PR neither the code is documented).

Basically, in `master` branch each transformation rule has a "Type" string field that is [checked inside a switch](https://github.com/netobserv/flowlogs-pipeline/pull/330/files#diff-77b720060ce5b2e61d001e832eac1bf4d94ffdeb9734bbd4ee678d9398ad3892L47) to choose the code that should be actually executed.

I expected that using polimorphism would be faster than comparing strings on each operation, so I created the [Transformer interface](https://github.com/netobserv/flowlogs-pipeline/pull/330/files#diff-0a4b6dbd4cd061a2b5983688106b248eb76eb4f715ad0e47a8a83a4f4bbb9dc4R17-R19) and split the code inside the switch into their own types that implement the interface (e.g. `type AddRegexpIf api.NetworkTransformRule`).

Then, each transformer implementation is added to the rules array (now renamed as transformers) and [you only have to invoke each transform method in the main loop](https://github.com/netobserv/flowlogs-pipeline/pull/330/files#diff-77b720060ce5b2e61d001e832eac1bf4d94ffdeb9734bbd4ee678d9398ad3892R44-R46).

Despite it didn't provide any noticeable performance improvement, maybe is worth documenting and merging this if you think it might improve readability or extensibility.

PR #335, opened 2022-11-25
NETOBSERV-666: K8s decoration not adding namespace if empty

mariomac:
This caused that empty-string and nil namespaces were treated differently in some metrics aggregations:
image

Now there are no duplicities:
image


jpinsonneau (approved):
Should we also do this to all enriched fields ? :thinking: 
I don't think I saw that on other ones but it could be a safety belt for us 

LGTM, thanks @mariomac

PR #336, opened 2022-11-28
S3 encode

KalmanMeth:
This PR adds an encode stage to write flows to an S3 object store.
This is a PR with the initial encode-s3 implementation.
In future PRs we will enhance the functionality, as needed.
We also need to update the documentation on how to use this new stage.


pkg/pipeline/encode/encode_s3.go
```diff
@@ -0,0 +1,219 @@
...
+		log.Errorf("error encoding object: %v", err)
+		return err
+	}
+	log.Debugf("encoded object = %v", b)
+	// TBD: add necessary headers such as authorization (token), gzip, md5, etc
+	uploadInfo, err := e.s3Client.PutObject(context.Background(), bucket, objectName, b, int64(b.Len()), minio.PutObjectOptions{ContentType: "application/octet-stream"})
+	log.Debugf("uploadInfo = %v", uploadInfo)
+	if err != nil {
+		log.Errorf("error writing to object store: %v", err)
+	}
```
ronensc, line 217: I don't there is need to log here since the caller is also logging this error

pkg/pipeline/encode/encode_s3.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// Encode queues entries to be sent to object store
+func (s *encodeS3) Encode(entry config.GenericMap) {
+	log.Debugf("Encode S3, entry = %v", entry)
+	s.mutex.Lock()
+	s.pendingEntries = append(s.pendingEntries, entry)
+	s.mutex.Unlock()
+	s.recordsWritten.Inc()
+	if len(s.pendingEntries) >= s.s3Params.BatchSize {
```
ronensc, line 136: This access to `s.pendingEntries` is not guarded by the mutex.
If there is a context switch to the TimeoutLoop right after a truthy evaluation of this condition but before the call to writeObject(), then we might end up with writing empty objects.

ronensc (approved):
Most of my comments are questions and suggestions.
I think the most important comment is about the mutex.

pkg/pipeline/encode/encode_s3.go
```diff
@@ -0,0 +1,219 @@
...
+		case <-s.exitChan:
+			log.Debugf("exiting createObjectTimeoutLoop because of signal")
+			return
+		case <-ticker.C:
+			log.Debugf("createObjectTimeoutLoop timer expire\n")
+			now := time.Now().Unix()
+			log.Debugf("time now = %d, expiryTime = %d \n", now, s.expiryTime)
+			if now >= s.expiryTime {
+				_ = s.writeObject()
+			}
```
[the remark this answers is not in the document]
KalmanMeth: OK

PR #340, opened 2022-12-07
Add S3 encode tests

KalmanMeth:


pkg/pipeline/encode/encode_s3.go
```diff
@@ -119,12 +118,11 @@ func (s *encodeS3) createObjectTimeoutLoop() {
 			log.Debugf("exiting createObjectTimeoutLoop because of signal")
 			return
 		case <-ticker.C:
-			log.Debugf("createObjectTimeoutLoop timer expire")
 			now := time.Now()
 			log.Debugf("time now = %v, expiryTime = %v", now, s.expiryTime)
-			if now.After(s.expiryTime) {
-				_ = s.writeObject()
-			}
```
ronensc, line 127: IIUC, this is the only place where `expiryTime` is read. If it is removed from here, maybe the variable can be removed altogether.

pkg/pipeline/encode/encode_s3_test.go
```diff
@@ -98,19 +107,85 @@ func Test_EncodeS3(t *testing.T) {
...
+	expectedSubstring = "hour="
+	require.True(t, strings.Contains(objectName0, expectedSubstring))
+	expectedSubstring = "stream-id="
+	require.True(t, strings.Contains(objectName0, expectedSubstring))
+	expectedSubstring = "00000000"
+	require.True(t, strings.Contains(objectName0, expectedSubstring))
+	expectedSubstring = "00000001"
+	require.True(t, strings.Contains(fakeWriter.objectNames[1], expectedSubstring))
+	expectedSubstring = "00000002"
+	require.True(t, strings.Contains(fakeWriter.objectNames[2], expectedSubstring))
```
ronensc, line 144: Using `require.Contains()` is preferred over `require.True()`.
The advantage of  `require.Contains()` is that when the condition is not met, the error message is more verbose.

ronensc (approved):

pkg/pipeline/encode/encode_s3.go
KalmanMeth: This is exactly for what the parameter is needed, to define the user's desired timeout period when to create a new object. So I don't think we can remove this.

pkg/pipeline/encode/encode_s3_test.go
KalmanMeth: updated accordingly

PR #343, opened 2022-12-08
Added arch detection during build

OlivierCazade:


mariomac:
Mi main concern here is: can we make sure that the machines in the CI pipeline are x86_64?

Makefile
```diff
@@ -8,7 +8,7 @@ export GOFLAGS=-mod=vendor
 export GO111MODULE=on
 export CGO_ENABLED=0
 export GOOS=linux
-export GOARCH=amd64
+export GOARCH=$(uname -m)
```
mariomac, line 11: Maybe I'd just allow overriding it so we can make sure that if e.g. GitHub Actions are run in an ARM machine, they won't break the containers we deploy to Quay.
```Makefile
GOARCH ?= amd64
```

jotak:
@OlivierCazade if I remember correctly, in your PoC you did different changes than this one, right? Is this PR still relevant?

OlivierCazade:
No, I will submit a PR with multi stage for cross compilation, we can close this one.

PR #344, opened 2022-12-11
Add sampling factor multiplier

KalmanMeth:
This PR allows to add in the configuration file a multiplier to apply to particular fields in a flow log.
For example, if we know that we only get a sampling of the flow logs (e.g. 1/20), then the administrator may want us to multiply the measured number of bytes and/or packets in the flow log entries in order to get a closer estimate of the actual number of bytes and/or packets.
Each field that requires a multiplier must be individually marked.
The implementation in this PR adds the functionality in the transform_network stage, where we already have fields that can be used to represent this operation.
We should consider whether it makes more sense to place the operation in the transform_generic stage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,7 @@ Using `remove_entry_if_not_equal` will remove the entry if the specified field e
 1. Compute geo-location from IP addresses
 1. Resolve kubernetes information from IP addresses
 1. Perform regex operations on field values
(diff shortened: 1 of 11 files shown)


KalmanMeth:
I think the implementation of this functionality in the transform-generic stage is more natural and fits better the use of the various stages. 
See #345 .

PR #345, opened 2022-12-12
implemented multiplier in transform-generic

KalmanMeth:
This PR allows to add in the configuration file a multiplier to apply to particular fields in a flow log.
For example, if we know that we only get a sampling of the flow logs (e.g. 1/20), then the administrator may want us to multiply the measured number of bytes and/or packets in the flow log entries in order to get a closer estimate of the actual number of bytes and/or packets.
Each field that requires a multiplier must be individually marked.
This PR implements this functionality in the transform-generic stage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,7 @@ The rule `multiplier` takes the input field, multiplies it by the provided value
 places the result in the output field.
 This is useful to use when provided with only a sample of the flow logs (e.g. 1 our of 20),
 and some of the variables need to be adjusted accordingly.
+If `multipier` is not set or if it is set to 0, then the input field is simply copied to the output field.
 
 For example, suppose we have a flow log with the following syntax:
 ```
--- a/README.md
+++ b/README.md
@@ -254,6 +254,11 @@ Specify `policy: replace_keys` to use only the newly specified keys.
 To include the original keys and values in addition to those specified in the `rules`,
 specify `policy: preserve_original_keys`.
 
+The rule `multiplier` takes the input field, multiplies it by the provided value, and
+places the result in the output field.
+This is useful to use when provided with only a sample of the flow logs (e.g. 1 our of 20),
+and some of the variables need to be adjusted accordingly.
+
 For example, suppose we have a flow log with the following syntax:
 ```
 {"Bytes":20800,"DstAddr":"10.130.2.2","DstPort":36936,"Packets":400,"Proto":6,"SequenceNum":1919,"SrcAddr":"10.130.2.13","SrcHostIP":"10.0.197.206","SrcPort":3100,"TCPFlags":0,"TimeFlowStart":0,"TimeReceived":1637501832}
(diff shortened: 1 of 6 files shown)


pkg/pipeline/transform/transform_generic.go
```diff
@@ -33,17 +33,50 @@ type Generic struct {
...
+			case uint64:
+				outputEntry[transformRule.Output] = uint64(transformRule.Multiplier) * outputEntry[transformRule.Input].(uint64)
+			case float32:
+				outputEntry[transformRule.Output] = float32(transformRule.Multiplier) * entry[transformRule.Input].(float32)
+			case float64:
+				outputEntry[transformRule.Output] = float64(transformRule.Multiplier) * entry[transformRule.Input].(float64)
+			default:
+				ok = false
+				log.Errorf("%s not of numerical type; cannot perform multiplication", transformRule.Output)
+			}
```
ronensc, line 73: I suggest to extract the switch to a function and keep the Transform() method clean and short.

pkg/pipeline/transform/transform_generic.go
```diff
@@ -33,17 +33,50 @@ type Generic struct {
...
+				outputEntry[transformRule.Output] = int64(transformRule.Multiplier) * outputEntry[transformRule.Input].(int64)
+			case uint64:
+				outputEntry[transformRule.Output] = uint64(transformRule.Multiplier) * outputEntry[transformRule.Input].(uint64)
+			case float32:
+				outputEntry[transformRule.Output] = float32(transformRule.Multiplier) * entry[transformRule.Input].(float32)
+			case float64:
+				outputEntry[transformRule.Output] = float64(transformRule.Multiplier) * entry[transformRule.Input].(float64)
+			default:
+				ok = false
+				log.Errorf("%s not of numerical type; cannot perform multiplication", transformRule.Output)
```
ronensc, line 72: If we return false, then this will omit the flow log. Is this the desired behavior?

ronensc, line 72: Should we add an operational metric to track the number of errors of this type?

KalmanMeth: Yes. We cannot produce the proper results because of the type error, so we skip the entry.

pkg/pipeline/transform/transform_generic.go
KalmanMeth: done

pkg/pipeline/transform/transform_generic.go
KalmanMeth: The other transforms do not currently have operational metrics. We should probably make a project-wide decision on whether to output such metrics from the transform stage.

README.md
```diff
@@ -254,6 +254,11 @@ Specify `policy: replace_keys` to use only the newly specified keys.
 To include the original keys and values in addition to those specified in the `rules`,
 specify `policy: preserve_original_keys`.
 
+The rule `multiplier` takes the input field, multiplies it by the provided value, and
+places the result in the output field.
+This is useful to use when provided with only a sample of the flow logs (e.g. 1 our of 20),
+and some of the variables need to be adjusted accordingly.
```
ronensc, line 260: Should we mention that for `multiplier: 0` the feature is disabled?

ronensc (approved):

README.md
KalmanMeth: done

PR #342, opened 2022-12-07
NETOBSERV-743: update build image & CI tests to Go 1.18 (4.12 backport)

mariomac:
backport of https://github.com/netobserv/flowlogs-pipeline/pull/341

PR #347, opened 2022-12-12
fix linting in Go 1.19 (for release-4.12)

mariomac:


eranra (approved):
THX @mario --- I checked with go 1.19, and it builds as expected :-)

eranra (commented):
@mariomac I see that the CI tests fails --- you probably need to fix some of that process from 1.18 to 1.19

mariomac:
@eranra yes, this will pass in release-4.12 when this PR is also merged: https://github.com/netobserv/flowlogs-pipeline/pull/342

PR #348, opened 2022-12-12
change error to warning

KalmanMeth:
When we have a location enrichment failure, we currently isue an error, which fills the log with error messages on each faulty flow log. We were requested to change this to a warning, so that it doesn't flood the error log.


mariomac (approved):
Better warning than error, but I'd even set this to Debug level.

PR #350, opened 2022-12-16
NETOBSERV-696 Reinterpret flow direction at a higher level

jotak:
Flow direction, as provided by the agent, relates to the underlying interface, ie. the flow can be said as "ingress" when going from pod IP to host interface, although at a higher level of abstraction it should be considered egress.

This commit reinterpret flow directions based on source and dest IPs and reporter IP.

The initial flow direction is preserved, copied into a new field such as IfDirection (for interface direction)


jotak:
Opening as draft, as it must be tested fully with NETOBSERV-759

pkg/pipeline/transform/transform_network_direction.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+)
+
+const (
+	ingress = 0
+	egress  = 1
+)
+
+func reinterpretDirection(output config.GenericMap, info *api.DirectionInfo) {
+	output[info.IfDirectionField] = output[info.FlowDirectionField]
```
KalmanMeth, line 14: We should check that the field specified by `info.FlowDirectionField` actually exists.

pkg/pipeline/transform/transform_network_direction.go
```diff
@@ -0,0 +1,38 @@
...
+		}
+	}
+	if srcNode != dstNode {
+		if srcNode == reporter {
+			output[info.FlowDirectionField] = egress
+		} else if dstNode == reporter {
+			output[info.FlowDirectionField] = ingress
+		}
+	}
+}
```
KalmanMeth, line 38: Does this give the correct result even if one or more of these values is nil? What if all of `srcNode`, `dstNode`, and `reporter` are nil?

jotak: good catches, thanks!
I've adding more missing fields coverage in last commit
Basically:
- When no FlowDirection set, do not set IfDirection so it's consistent 
- When no reporter set, then the whole reinterpretation cannot be done so we just set `IfDirection` (for consistency) but no change in `FlowDirection`.
- When src and/or dst aren't set they just won't match reporter so FlowDirection will be unchanged

That's also added to unit tests

pkg/pipeline/transform/transform_network_test.go
```diff
@@ -366,3 +366,147 @@ func (*fakeKubeData) GetInfo(n string) (*kubernetes.Info, error) {
...
+				},
+			},
+		},
+	}
+
+	tr, err := NewTransformNetwork(cfg)
+	require.NoError(t, err)
+
+	// Source reporter => egress (1)
+	output, ok := tr.Transform(entryReporterSource)
```
mariomac, line 433: I would either:
- embed here the struct instantiation so we can see the input and the expected output in the same page, without having to stroll up and down to understand the tests.
or,
- use an array of testcases where you can see them all together, like in https://github.com/netobserv/netobserv-ebpf-agent/pull/78/files#diff-029307d4e44046bf27e28c7ae51e1673c020ff459796821bf11d33d382082e77R34-R72

mariomac (commented):
Mostly OK, just minor styling comments.

First, we should merge this PR of the agent: https://github.com/netobserv/netobserv-ebpf-agent/pull/78

Then, we could update the vendors and merge this another PR of FLP (either in main or in the branch of this PR): https://github.com/netobserv/flowlogs-pipeline/pull/353

pkg/pipeline/transform/transform_network_test.go
jotak: ah, I must say I prefer the flat / straightforward approach rather than the array of test cases, but I get your point of seeing the input and expectations side by side ; i'll update that

jotak: PS: my rationale for flat test cases is just that I find it easier to quickly jump to a faulty assertion , but that's really a matter of preference

pkg/api/transform_network.go
```diff
@@ -66,4 +69,12 @@ type NetworkTransformRule struct {
 	Assignee   string `yaml:"assignee,omitempty" json:"assignee,omitempty" doc:"value needs to assign to output field"`
 }
 
+type DirectionInfo struct {
+	ReporterIPField    string `yaml:"reporterIPField,omitempty" json:"reporterIPField,omitempty" doc:"field providing the reporter (agent) host IP"`
+	SrcHostField       string `yaml:"srcHostField,omitempty" json:"srcHostField,omitempty" doc:"source host field"`
+	DstHostField       string `yaml:"dstHostField,omitempty" json:"dstHostField,omitempty" doc:"destination host field"`
+	FlowDirectionField string `yaml:"flowDirectionField,omitempty" json:"flowDirectionField,omitempty" doc:"field providing the flow direction in the input entries; it will be rewritten"`
+	IfDirectionField   string `yaml:"ifDirectionField,omitempty" json:"ifDirectionField,omitempty" doc:"interface-level field for flow direction, to create in output"`
+}
```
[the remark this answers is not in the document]
jotak: I believe the differences with the existing structure are big enough to not try reusing it (i don't want to force a square peg into a round hole)
I've added the config validation as you suggest in last commit

pkg/pipeline/transform/transform_network_direction.go
```diff
@@ -0,0 +1,43 @@
...
+		if str, ok := gen.(string); ok {
+			dstNode = str
+		}
+	}
+	if srcNode != dstNode {
+		if srcNode == reporter {
+			output[info.FlowDirectionField] = egress
+		} else if dstNode == reporter {
+			output[info.FlowDirectionField] = ingress
+		}
```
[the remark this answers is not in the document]
jotak: yes, although it shouldn't happen (if it happens this whole feature becomes irrelevant)

mariomac (approved):

PR #351, opened 2022-12-19
Made prom and health server url configurable

### LINKED ISSUE #349: Make prom server and health server address url configurable
Currently prom server and health server by default runs on url `0.0.0.0` but it should be configurable.

tkuldeep:
* Make `address` field as configurable for prom server.
* Make health address as configurable similar to port.

Fixes #349

Signed-off-by: Kuldeep Singh Pal <tkuldeep.allen@gmail.com>

### DIFF
--- a/pkg/confgen/flowlogs2metrics_config.go
+++ b/pkg/confgen/flowlogs2metrics_config.go
@@ -56,6 +56,7 @@ func (cg *ConfGen) GenerateFlowlogs2PipelineConfig() *config.ConfigFileStruct {
 	}
 	if len(cg.promMetrics) > 0 {
 		metricsNode.EncodePrometheus("encode_prom", api.PromEncode{
+			Address: cg.config.Encode.Prom.Address,
 			Port:    cg.config.Encode.Prom.Port,
 			Prefix:  cg.config.Encode.Prom.Prefix,
 			Metrics: cg.promMetrics,
--- a/pkg/operational/health.go
+++ b/pkg/operational/health.go
@@ -27,8 +27,6 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const defaultServerHost = "0.0.0.0"
-
 type Server struct {
 	handler healthcheck.Handler
 	Address string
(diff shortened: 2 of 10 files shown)


pkg/pipeline/encode/encode_prom.go
```diff
@@ -348,7 +348,7 @@ func NewEncodeProm(opMetrics *operational.Metrics, params config.StageParam) (En
 	log.Debugf("histos = %v", histos)
 	log.Debugf("aggHistos = %v", aggHistos)
 
-	addr := fmt.Sprintf(":%v", cfg.Port)
+	addr := fmt.Sprintf("%s:%v", cfg.Address, cfg.Port)
```
KalmanMeth, line 351: The prom stage is used also by others in existing deployments, so we need to maintain that 0.0.0.0 is the default, even if the user did not specify it.

pkg/operational/health.go
```diff
@@ -45,8 +43,7 @@ func (hs *Server) Serve() {
 func NewHealthServer(opts *config.Options, isAlive healthcheck.Check, isReady healthcheck.Check) *Server {
 
 	handler := healthcheck.NewHandler()
-	address := net.JoinHostPort(defaultServerHost, opts.Health.Port)
-
+	address := net.JoinHostPort(opts.Health.Address, opts.Health.Port)
```
KalmanMeth, line 46: The health metrics are used also by others in existing deployments, so we need to maintain that 0.0.0.0 is the default, even if the user did not specify it.

KalmanMeth: Verify that the health_test.go runs successfully. We also need to (eventually) add a test to verify that the new functionality (address) is handled properly.

pkg/confgen/flowlogs2metrics_config.go
```diff
@@ -56,6 +56,7 @@ func (cg *ConfGen) GenerateFlowlogs2PipelineConfig() *config.ConfigFileStruct {
 	}
 	if len(cg.promMetrics) > 0 {
 		metricsNode.EncodePrometheus("encode_prom", api.PromEncode{
+			Address: cg.config.Encode.Prom.Address,
```
KalmanMeth, line 59: Check here that cg.config.Encode.Prom.Address is actually defined (is not nil). If not, use 0.0.0.0.

pkg/pipeline/encode/encode_prom.go
KalmanMeth: If address is not specified, then we get ":<port>" and not "0.0.0.0:<port>", but that is what we had before, and it works OK.

pkg/confgen/flowlogs2metrics_config.go
KalmanMeth: This is not necessary. If address is not specified, then we end up with ":<port>", just like we used to have earlier, and it works OK.

KalmanMeth (approved):

PR #353, opened 2022-12-19
NETOBSERV-759: parse AgentIP from protobuf encoding

mariomac:
First, the following PR should be merged: https://github.com/netobserv/netobserv-ebpf-agent/pull/78

Then we can update te vendors with the `main` branch sources.


go.mod
```diff
@@ -113,3 +113,5 @@ require (
 )
 
 replace github.com/prometheus/common v0.32.1 => github.com/netobserv/prometheus-common v0.31.2-0.20220720134304-43e74fd22881
+
+replace github.com/netobserv/netobserv-ebpf-agent v0.2.2-0.20221031160841-ccb7addde22f => ../netobserv-ebpf-agent
```
mariomac, line 117: Just a local fix until we merge the last netobserv-ebpf-agent fix into master. Will crash CI but should work locally if both flp and the agent folders are in the same parent folder.

KalmanMeth (approved):

jotak (approved):

PR #357, opened 2022-12-26
conntrack: Configurable timeouts

ronensc:
This PR allows configuring different timeouts for connections based on their key fields.

```yaml
conntrack:
  keyDefinition:
    fieldGroups:
    - name: protocol
      fields:
      - Proto
    # ...
  scheduling:
  - selector: # UDP connections
      Proto: 17
    endConnectionTimeout: 5s
    updateConnectionInterval: 40s
  - selector: {} # Default group
    endConnectionTimeout: 10s
    updateConnectionInterval: 30s
```

Implementation: Instead of using a single Multi-Order Map in the connection store for all the connections, the connection store will have a Multi-Order Map for each scheduling group. In addition, the store has a new `hashId2groupIdx` map that given a connection hash ID, finds the Multi-Order Map of the connection.

Other changes:
- Add a scheduling group label to the `conntrack_memory_connections` operational metric. This allows monitoring the number of active connections per scheduling group.


pkg/api/conntrack.go
```diff
@@ -83,6 +82,12 @@ type ConnTrackOperationEnum struct {
 	Max   string `yaml:"max" doc:"max"`
 }
 
+type ConnTrackSchedulingGroup struct {
+	Selector                 map[string]string `yaml:"selector,omitempty" doc:"key-value map to match against connection fields to apply this scheduling"`
```
KalmanMeth, line 86: Selector matches only string: string. Is this the intended usage? So we cannot, for instance, select on integer values?

ronensc: We can select integer values. The integer values are [converted](https://github.com/netobserv/flowlogs-pipeline/pull/357/files#diff-bd390b4fc85fff1d1e1532f7a0077a03f1e34265d04cbe4ace449fb22e6a2d6fR113) to string before the comparison.

At first, I used `map[string]interface{}`. But then I noticed that setting the selector to `Proto: 6` is parsed as `int`. The problem is that the grpc ingester [parses](https://github.com/netobserv/flowlogs-pipeline/blob/30d42fd09a981391ee6ea52ef6dc18786474509c/vendor/github.com/netobserv/netobserv-ebpf-agent/pkg/pbflow/flow.pb.go#L547) `Proto` as `uint32`. Comparing `int` and `uint32` is evaluated to false regardless of the values. That's why I convert and compare strings.

pkg/api/conntrack_test.go
```diff
@@ -145,6 +145,58 @@ func TestConnTrackValidate(t *testing.T) {
...
+						Selector: map[string]string{
+							"srcIP": "value",
+						},
+					},
+				},
+			},
+			conntrackInvalidError{defaultGroupAndNotLast: true},
+		},
+		{
+			"Exactly 1 default selector",
```
KalmanMeth, line 189: Do we try to test if 2 defaults are defined?

ronensc: I'll add a test for this

pkg/pipeline/extract/conntrack/conntrack.go
```diff
@@ -116,65 +116,50 @@ func (ct *conntrackImpl) Extract(flowLogs []config.GenericMap) []config.GenericM
...
 		}
-		return
-	})
+		addIsFirstField(record, isFirst)
+		outputRecords = append(outputRecords, record)
+	}
 	return outputRecords
 }
 
+// TBD: think of how to avoid confusion with the word "update". It is used with periodic connections updates and for update a connection with incoming flow logs.
```
jotak, line 156: Just a suggestion, you could use the terminology of "heartbeats" to refer to the periodic updates on long connections?

ronensc: I like this idea! thanks @jotak :)

jotak (approved):
lgtm, nice PR!

ronensc:
I tired to improve `isMatchSelector()`. Instead of converting both values to string and compare strings,
I added a check for the type of the value and convert the selector to that type. So at the end we have only one conversion and the comparison is in the original type.
Currently, I implemented checks only for `uint32`, `uint64`, `int64`, `int` and `bool` since these types are in use by the [protobuf decoder](https://github.com/netobserv/flowlogs-pipeline/blob/fc3940de8d10c48b351763a47e4fce16952bf61e/pkg/pipeline/decode/decode_protobuf.go#L37-L54). Other types fallback to the string conversion-comparison.
I added a benchmark test to compare the performance of both approaches.
It seems that the new approach is faster. But, I'm not pleased with the code. It feels too verbose to me.
@jotak @KalmanMeth wdyt?, do you have ideas for improving?
```
goos: linux
goarch: amd64
pkg: github.com/netobserv/flowlogs-pipeline/pkg/pipeline/extract/conntrack
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkIsMatchSelectorStrings
BenchmarkIsMatchSelectorStrings/Proto_1
BenchmarkIsMatchSelectorStrings/Proto_1-4         	10560730	       133.7 ns/op
BenchmarkIsMatchSelectorStrings/Proto_2
BenchmarkIsMatchSelectorStrings/Proto_2-4         	10563658	       100.8 ns/op
BenchmarkIsMatchSelectorStrings/Multiple_fields_1
BenchmarkIsMatchSelectorStrings/Multiple_fields_1-4         	 2900412	       413.0 ns/op
BenchmarkIsMatchSelectorStrings/Multiple_fields_2
BenchmarkIsMatchSelectorStrings/Multiple_fields_2-4         	 5226672	       203.3 ns/op
BenchmarkIsMatchSelectorGeneric
BenchmarkIsMatchSelectorGeneric/Proto_1
BenchmarkIsMatchSelectorGeneric/Proto_1-4                   	24312217	        50.79 ns/op
BenchmarkIsMatchSelectorGeneric/Proto_2
BenchmarkIsMatchSelectorGeneric/Proto_2-4                   	28917752	        41.54 ns/op
BenchmarkIsMatchSelectorGeneric/Multiple_fields_1
BenchmarkIsMatchSelectorGeneric/Multiple_fields_1-4         	10039882	       119.5 ns/op
BenchmarkIsMatchSelectorGeneric/Multiple_fields_2
BenchmarkIsMatchSelectorGeneric/Multiple_fields_2-4         	12100579	        99.83 ns/op
PASS
```

pkg/pipeline/extract/conntrack/conn.go
```diff
@@ -106,6 +109,64 @@ func (c *connType) markReported() bool {
...
+			selectorValue, err := utils.ConvertToBool(v)
+			if err != nil || connValue != selectorValue {
+				return false
+			}
+		case string:
+		default:
+			return fmt.Sprintf("%v", c.keys[k]) != connValue
+		}
+	}
+	return true
```
KalmanMeth, line 167: For the default case, if it comes out `true`, don't we have to continue the loop to check the remaining conditions?

pkg/pipeline/extract/conntrack/conntrack.go
```diff
@@ -228,7 +213,7 @@ func NewConnectionTrack(opMetrics *operational.Metrics, params config.StageParam
 	metrics := newMetrics(opMetrics)
 	conntrack := &conntrackImpl{
 		clock:                        clock,
-		connStore:                    newConnectionStore(metrics),
+		connStore:                    newConnectionStore(cfg.Scheduling, metrics, clock.Now),
```
KalmanMeth, line 216: Why is clock passed as a parameter to NewConnectionTrack() rather than defining it here?

pkg/pipeline/extract/conntrack/conn.go
ronensc: Good catch! My bad.
In addition, the `case string:` is also wrong, in Go there is no automatic fallthrough and in type assertion switch it's not allowed anyway. So, in case of strings, it simply ignores them. I'll add a fix and another test cases.

pkg/pipeline/extract/conntrack/conntrack.go
ronensc: `NewConnectionTrack()` accepts the clock as an argument to allow mocking it in the unit tests.
Mocking example:
https://github.com/netobserv/flowlogs-pipeline/blob/67fb882dab394b808ab09716974ba55145497340/pkg/pipeline/extract/conntrack/conntrack_test.go#L608
https://github.com/netobserv/flowlogs-pipeline/blob/67fb882dab394b808ab09716974ba55145497340/pkg/pipeline/extract/conntrack/conntrack_test.go#L613
https://github.com/netobserv/flowlogs-pipeline/blob/67fb882dab394b808ab09716974ba55145497340/pkg/pipeline/extract/conntrack/conntrack_test.go#L657

jotak:
(and sorry for the delay!)
/lgtm

jotak:
@ronensc iirc we have the same kind of "unelegant but better" conversion somewhere in the code for loki... FLP is in a position where we must not neglect optimizations, so faster is better IMO

pkg/pipeline/extract/conntrack/conntrack.go
```diff
@@ -72,9 +72,9 @@ func (ct *conntrackImpl) Extract(flowLogs []config.GenericMap) []config.GenericM
 				Hash(computedHash).
 				KeysFrom(fl, ct.config.KeyDefinition).
 				Aggregators(ct.aggregators).
```
KalmanMeth, line 74: KeysFrom(fl, ...) This seems to be dependent on the flow log. But this is used to define fields in the connType structure. Might the values not be different for the same connection with a different flow log?

KalmanMeth (approved):

PR #359, opened 2023-01-03
NETOBSERV-386 IP categorization

jotak:
JIRA: https://issues.redhat.com/browse/NETOBSERV-386

Allows to configure IP ranges and assign them any name
IPs falling in those ranges are flagged with that name

### DIFF
--- a/pkg/pipeline/transform/transform_network.go
+++ b/pkg/pipeline/transform/transform_network.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"regexp"
 	"strconv"
+	"time"
 
 	"github.com/Knetic/govaluate"
 	"github.com/netobserv/flowlogs-pipeline/pkg/api"
@@ -30,6 +31,7 @@ import (
 	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/transform/kubernetes"
 	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/transform/location"
 	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/transform/netdb"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/utils"
 	"github.com/sirupsen/logrus"
 )
 
@@ -39,6 +41,7 @@ type Network struct {
 	api.TransformNetwork
 	svcNames   *netdb.ServiceNames
 	categories []subnetCategory
+	ipCatCache *utils.TimedCache
 }
 
 type subnetCategory struct {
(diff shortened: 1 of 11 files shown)


pkg/pipeline/transform/transform_network.go
```diff
@@ -152,6 +165,17 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
...
+	for _, subnetCat := range n.categories {
+		for _, cidr := range subnetCat.cidrs {
+			if cidr.Contains(ip) {
+				return subnetCat.name
+			}
+		}
+	}
+	return ""
+}
+
```
mariomac, line 178: I'm afraid this exhaustive search could increase noticeably the CPU usage. I'd add an lru cache implementation (like [this](https://github.com/netobserv/flowlogs-pipeline/blob/main/pkg/pipeline/utils/timed_cache.go) or [this](https://github.com/mariomac/guara/blob/main/pkg/cache/lru.go)) and check there the IP first.
In pseudo-golang:
```go
func (n *Network) categorizeIP(ip net.IP) string {
	if category, ok := n.cache.Get(ip); ok {
		return category
	}
	for _, subnetCat := range n.categories {
		for _, cidr := range subnetCat.cidrs {
			if cidr.Contains(ip) {
				n.cache.Put(ip, subnetCat.name)
				return subnetCat.name
			}
		}
	}
	n.cache.Put(ip, "")
	return ""
}
```

jotak: Yes I have the same concern and wanted to point out in the documentation (when this is implemented on operator side) that this could be disabled to save processor's resource.
But +1 about implementing a cache right now

jotak: done
I took the opportunity to update a little bit `TimedCache` in order to use go's time API + a couple of small stuffs

mariomac (approved):

pkg/pipeline/transform/transform_network.go
```diff
@@ -211,11 +246,28 @@ func NewTransformNetwork(params config.StageParam) (Transformer, error) {
...
 	return &Network{
 		TransformNetwork: api.TransformNetwork{
 			Rules:         jsonNetworkTransform.Rules,
 			DirectionInfo: jsonNetworkTransform.DirectionInfo,
 		},
-		svcNames: servicesDB,
+		svcNames:   servicesDB,
+		categories: subnetCats,
+		ipCatCache: utils.NewQuietExpiringTimedCache(2 * time.Minute),
```
KalmanMeth, line 271: Should this timeout be configurable?

jotak: I don't think it's important to make it configurable, it won't have any impact on the results (unlike conn-tracking timeout, for instance, where timeout impacts updates granularity) - this one is really an implementation detail

jpinsonneau:
`NetworkTransformDirectionInfo` is a breaking change

PR #363, opened 2023-01-20
Fix release multiarch

OlivierCazade:
Removed basename command in makefile and switched to docker because of podman version in github runner


jotak:
@OlivierCazade for information you can use the ok-to-test label to test the build & push (its action needs to be updated as well, I guess)

PR #369, opened 2023-01-26
added jsonnet and grafanet into confgenerator code

KalmanMeth:
As a step towards creating grafana dashboards in the Network Observability Operator, this PR adds into the confgenerator the steps of running jsonnet using the grafannet defined constructs to generate the grafana panels, accessible via a function call (instead of performing these operations externally and then reading additional files).
The grafannet files are embedded in the confgenerator code so that they will be accessible when we run this code from the operator.


pkg/confgen/grafana_jsonnet.go
```diff
@@ -165,32 +168,39 @@ type Dashboard struct {
 	Panels []byte
 }
 
-func (cg *ConfGen) generateGrafanaJsonnet(folderName string) error {
+func (dashboard *Dashboard) generateDashboardJson() []byte {
```
ronensc, line 171: Please consider renaming to `generateDashboardJsonnet()` because the output is jsonnet rather than json.

pkg/confgen/grafana_jsonnet.go
```diff
@@ -292,3 +327,94 @@ func (cg *ConfGen) addPanelsToDashboards(dashboards Dashboards) (Dashboards, err
...
+	fsBase  embed.FS
+	fsCache map[string]*fsCacheEntry
+}
+
+type fsCacheEntry struct {
+	contents *[]byte
+	exists   bool
+}
+
+func (cg *ConfGen) GenerateGrafanaJson() (string, error) {
```
ronensc, line 346: It seems that `GenerateGrafanaJson()` is called only in `confgen_test.go` and not in production code. If it is still needed, maybe it can be moved to `confgen_test.go`?

ronensc (commented):
We have the [hack/generate-dashboards.sh](https://github.com/netobserv/flowlogs-pipeline/blob/bb058e1dec0ddc960040ca0efab291b3cc8dc0cf/hack/generate-dashboards.sh) script which is triggered by `make dashboards`.
This script downloads the `jsonnet` CLI and `grafonnet-lib` to `./bin`
So we have 2 instances of `grafonnet-lib`. Can we consolidate the two?

pkg/confgen/grafana_jsonnet.go
KalmanMeth: GenerateGrafanaJson() is the function that will be called from the Network Observability Operator to create the dashboards for Openshift.

KalmanMeth:
> So we have 2 instances of `grafonnet-lib`. Can we consolidate the two?
Yes, we should remove the script from hack once this PR is merged.

pkg/confgen/grafana/grafonnet-lib/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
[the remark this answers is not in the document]
KalmanMeth: Added a sentence in confGenerator.md. Note that the License is already included in the grafonnet-lib directory.

ronensc (approved):

PR #378, opened 2023-02-08
corrected field to perform multipler

KalmanMeth:


ronensc (approved):
Can't believe I've missed this in the review...

PR #387, opened 2023-02-14
NETOBSERV-838 Group ingress+egress flows in same connection

jpinsonneau:
This PR skip duplicates in connection tracking aggregations.

It rely on `Duplicate` field [provided by eBPF agent](https://github.com/netobserv/netobserv-ebpf-agent/pull/66)
image
image

### DIFF
--- a/pkg/config/generic_map.go
+++ b/pkg/config/generic_map.go
@@ -29,3 +29,12 @@ func (m GenericMap) Copy() GenericMap {
 
 	return result
 }
+
+func (m GenericMap) IsDuplicate() bool {
+	if duplicate, hasKey := m["Duplicate"]; hasKey {
+		if isDuplicate, ok := duplicate.(bool); ok && isDuplicate {
+			return true
+		}
+	}
+	return false
+}
--- a/pkg/config/generic_map.go
+++ b/pkg/config/generic_map.go
@@ -17,6 +17,8 @@
 
 package config
 
+import "github.com/netobserv/flowlogs-pipeline/pkg/utils"
+
 type GenericMap map[string]interface{}
 
 // Copy will create a flat copy of GenericMap
@@ -32,8 +34,8 @@ func (m GenericMap) Copy() GenericMap {
 
 func (m GenericMap) IsDuplicate() bool {
 	if duplicate, hasKey := m["Duplicate"]; hasKey {
-		if isDuplicate, ok := duplicate.(bool); ok && isDuplicate {
-			return true
+		if isDuplicate, err := utils.ConvertToBool(duplicate); err != nil {
+			return isDuplicate
 		}
 	}
 	return false
--- a/pkg/pipeline/extract/conntrack/conntrack.go
+++ b/pkg/pipeline/extract/conntrack/conntrack.go
@@ -160,8 +160,10 @@ func (ct *conntrackImpl) prepareHeartbeatRecords() []config.GenericMap {
 
 func (ct *conntrackImpl) updateConnection(conn connection, flowLog config.GenericMap, flowLogHash totalHashType) {
 	d := ct.getFlowLogDirection(conn, flowLogHash)
-	for _, agg := range ct.aggregators {
-		agg.update(conn, flowLog, d)
+	if !flowLog.IsDuplicate() {
+		for _, agg := range ct.aggregators {
+			agg.update(conn, flowLog, d)
+		}
 	}
 	ct.connStore.updateConnectionExpiryTime(flowLogHash.hashTotal)
 }
--- a/pkg/pipeline/extract/conntrack/conntrack.go
+++ b/pkg/pipeline/extract/conntrack/conntrack.go
@@ -159,8 +159,8 @@ func (ct *conntrackImpl) prepareHeartbeatRecords() []config.GenericMap {
 }
 
 func (ct *conntrackImpl) updateConnection(conn connection, flowLog config.GenericMap, flowLogHash totalHashType) {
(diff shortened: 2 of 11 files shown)


jpinsonneau:
@ronensc please let me know if you are fine with that approach. The [initial ticket](https://issues.redhat.com/browse/NETOBSERV-838) suggested to rely on 2 hash ids but I feel it's less performant as we already have the duplicate field on flows.

pkg/pipeline/extract/conntrack/conntrack.go
```diff
@@ -160,8 +160,10 @@ func (ct *conntrackImpl) prepareHeartbeatRecords() []config.GenericMap {
 
 func (ct *conntrackImpl) updateConnection(conn connection, flowLog config.GenericMap, flowLogHash totalHashType) {
 	d := ct.getFlowLogDirection(conn, flowLogHash)
```
ronensc, line 162: I think the call to `getFlowLogDirection()` could be moved under the new "if isn't duplicated"

pkg/config/generic_map.go
```diff
@@ -29,3 +29,12 @@ func (m GenericMap) Copy() GenericMap {
 
 	return result
 }
+
+func (m GenericMap) IsDuplicate() bool {
+	if duplicate, hasKey := m["Duplicate"]; hasKey {
+		if isDuplicate, ok := duplicate.(bool); ok && isDuplicate {
```
ronensc, line 35: We have `ConvertToBool()` that can handle cases where the type of `m["Duplicate"]` is not bool but is convertible to bool.
https://github.com/netobserv/flowlogs-pipeline/blob/3f2481ef3474ba0b07b595a02b50f813c9938916/pkg/pipeline/utils/convert.go#L220

I just noticed that `ConvertToBool()` is missing the trivial `case bool:`. If you think it's worth using `ConvertToBool()` here, the fix of adding the trivial case could be done in this PR as well.

ronensc (approved):
@jpinsonneau  I agree with you that this is much simpler and elegant than the approach suggested in the ticket.
@jotak , do you remember if there was a reason why we didn't follow this path the first place?

pkg/config/generic_map.go
jpinsonneau: I'm getting some conflicts when I try to import `pipeline/utils`:
```
$ make build
/home/julien/dev/flowlogs-pipeline/bin/golangci-lint-v1.50.1 run --enable goimports --enable gofmt --enable ineffassign --timeout 5m
ERRO Running error: context loading failed: failed to load packages: failed to load with go/packages: internal error: go list gives conflicting information for package github.com/netobserv/flowlogs-pipeline/pkg/config [github.com/netobserv/flowlogs-pipeline/pkg/config.test] 
make: *** [Makefile:70: lint] Error 3
```

Any idea ? Should we move `utils` outside of the `pipeline` package ?

pkg/pipeline/extract/conntrack/conntrack.go
jpinsonneau: flaky tests on data race => https://github.com/netobserv/flowlogs-pipeline/actions/runs/4185399722/jobs/7252391870

pkg/config/generic_map.go
ronensc: I think this errors implies a cyclic reference: `config` imports `pipeline/utils` and `pipeline/utils` imports `config`.
Looking at `pipeline/utils` files, it seems that only `batcher.go` has an import to `config`.
https://github.com/netobserv/flowlogs-pipeline/blob/3f2481ef3474ba0b07b595a02b50f813c9938916/pkg/pipeline/utils/batcher.go#L6

Hm... I'm not sure what would be the most proper solution in this case.

jpinsonneau: Yeah so we should either move away `Convert*` functions or `batcher.go` from `pipeline/utils`...
Or keep as is.

ronensc: I think I would go with the first option of moving `convert.go` to an "FLP-independent" utils package.

jpinsonneau: https://github.com/netobserv/flowlogs-pipeline/pull/387/commits/70514bf5353307c03d92be333bf72a7299f369d8 here we go

jpinsonneau:
rebased to pass the go `1.19` e2e checks

ronensc (approved):

jotak:
@ronensc @jpinsonneau I haven't reviewed the code, but what does it mean "This PR skip duplicates in connection tracking aggregations." ? iirc we wanted to keep duplicated on purpose because it gives information about the network interfaces. But probably I don't have all the context, I will do a better review hopefully tomorrow

jpinsonneau:
> @ronensc @jpinsonneau I haven't reviewed the code, but what does it mean "This PR skip duplicates in connection tracking aggregations." ? iirc we wanted to keep duplicated on purpose because it gives information about the network interfaces. But probably I don't have all the context, I will do a better review hopefully tomorrow

Connection tracking implements aggregators (`sum`, `count`, `min`, `max`) based on its flows. However, duplicates were initially counted and the aggregators were wrong.

This PR simply skip flows containing `"Duplicate": true` in aggregators calculations. The duplicated flows are still part of the connection since they share the same `hashId`.

For performance reasons, we reuse `Duplicate` field from https://github.com/netobserv/netobserv-ebpf-agent/pull/66

jotak:
Hmm ok this change looks good, but it doesn't address the "group ingress+egress" req, as the PR title shows, does it? Is there another PR for that?

PR #388, opened 2023-02-14
Add Count support to timebased TopK

KalmanMeth:


ronensc (approved):
+1 for using more meaningful names for the configs and the unit tests

Issue #299, opened 2022-09-07
Detect end connections by TCP flags


ronensc:
In the current implementation of the connection tracking, a connection is considered ended if a configurable timeout since the last flow-log of the connection has expired.
For protocols like TCP, we can determine that a connection has ended by inspecting its TCP flags (`FIN` flag) and cut the timeout.

This is also related to:
https://issues.redhat.com/browse/NETOBSERV-569

cc @jotak @eranra

eranra:
@mariomac @praveingk do you need to refer to this issue from a new one under the agent repo (or Jira?)

praveingk:
@ronensc @jotak @eranra @mariomac : Apart from FIN/RST, Are there other use-cases planned for connection tracking?
I initially had the logic to evict on FIN/RST, but we wanted to keep the v2 simple and introduce that as part of a bigger connection tracking feature. 
Some questions : 
1) From ebpf agent, do you need only the TCP Flags? are there more info needed?
2) If only TCP Flags? Do we want to track events like "who initiated the connection", "who closed it", etc?
3) Do we need other states like window size (16 bit), which are useful for performance analysis?

jotak:
> If only TCP Flags? Do we want to track events like "who initiated the connection", "who closed it", etc?
"Who initiated the connection" can be deduced from the flow w/ flags, right? Source IP on SYN == initiator?

praveingk:
Yes, I meant TCP flags could be used to answer those questions in userspace

PR #391, opened 2023-02-16
conntrack: handle TCP flags

ronensc:
This PR adds the ability to read the TCP flags field and perform the following actions:
- Ending a connection when the `FIN_ACK` flag is set and avoid waiting the `EndConnectionTimeout`.
- Swapping source and destination when `SYN_ACK` is set on the first flowlog.

Each action can be enabled/disabled by a feature flag in the conntrack config:
```
conntrack:
  # ...
  tcpFlags:
    fieldName: "TCPFlags"
    detectEndConnection: true
    swapAB: true
```

An operational metric was added to track how many of these actions were performed:
```
conntrack_tcp_flags{action="detectEndConnection"} 1
conntrack_tcp_flags{action="swapAB"} 1
```

Additional changes:
- Updated `go.mod` to 1.18 to be able to update `netobserv-ebpf-agent`
- Updated `netobserv-ebpf-agent` dependency to get the `Flags` field  in the flow protobuf
- Added `json` tag to the conntrack API

Solves #299


docs/api.md
```diff
@@ -237,6 +237,10 @@ Following is the supported API format for specifying connection tracking:
                  endConnectionTimeout: duration of time to wait from the last flow log to end a connection
                  heartbeatInterval: duration of time to wait between heartbeat reports of a connection
          maxConnectionsTracked: maximum number of connections we keep in our cache (0 means no limit)
+         tcpFlags: settings for handling TCP flags
+             fieldName: name of the field containing TCP flags
+             detectEndConnection: detect end connections by FIN_ACK flag
+             swapAB: swap source and destination when the first flowlog contains the SYN_ACK flag
```
jpinsonneau, line 243: I wonder if these two should be the default behavior as it's more reliable & convenient than the timeouts

KalmanMeth (approved):

docs/api.md
ronensc: The `swapAB` feature isn't related to the timeouts.
The `detectEndConnection` feature is in addition to the timeouts. It can't replace the timeouts because it's not guaranteed that will receive a flowlog with `FIN_ACK` flag for every TCP connection (either because of sampling or because of SYN attack). But, it may allow us to increase the `endConnectionTimeout` for TCP.

pkg/api/conntrack.go
```diff
@@ -28,33 +28,33 @@ const (
...
-	MaxConnectionsTracked int                        `yaml:"maxConnectionsTracked,omitempty" doc:"maximum number of connections we keep in our cache (0 means no limit)"`
+	KeyDefinition         KeyDefinition              `yaml:"keyDefinition,omitempty" json:"keyDefinition,omitempty" doc:"fields that are used to identify the connection"`
+	OutputRecordTypes     []string                   `yaml:"outputRecordTypes,omitempty" json:"outputRecordTypes,omitempty" enum:"ConnTrackOutputRecordTypeEnum" doc:"output record types to emit"`
+	OutputFields          []OutputField              `yaml:"outputFields,omitempty" json:"outputFields,omitempty" doc:"list of output fields"`
+	Scheduling            []ConnTrackSchedulingGroup `yaml:"scheduling,omitempty" json:"scheduling,omitempty" doc:"list of timeouts and intervals to apply per selector"`
+	MaxConnectionsTracked int                        `yaml:"maxConnectionsTracked,omitempty" json:"maxConnectionsTracked,omitempty" doc:"maximum number of connections we keep in our cache (0 means no limit)"`
+	TCPFlags              ConnTrackTCPFlags          `yaml:"tcpFlags,omitempty" json:"tcpFlags" doc:"settings for handling TCP flags"`
```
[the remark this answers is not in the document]
ronensc: Thanks! good catch

KalmanMeth (approved):

PR #394, opened 2023-02-20
add config parameter to supress go metrics

KalmanMeth:


pkg/config/config.go
```diff
@@ -38,7 +38,7 @@ type ConfigFileStruct struct {
 	LogLevel        string          `yaml:"log-level,omitempty" json:"log-level,omitempty"`
 	Pipeline        []Stage         `yaml:"pipeline,omitempty" json:"pipeline,omitempty"`
 	Parameters      []StageParam    `yaml:"parameters,omitempty" json:"parameters,omitempty"`
-	MetricsSettings MetricsSettings `yaml:"metricsSettings,omitempty" json:"metricsSettings,omitempty"`
+	MetricsSettings MetricsSettings `yaml:"metrics-settings,omitempty" json:"metrics-settings,omitempty"`
```
ronensc, line 41: Why the `camelCase` was replaced by `kebab-case`?

KalmanMeth: Changed back. We should consider changing also log-level to be consistent.

jotak: I don't get it ... it was previously `kebab-case` and not `camelCase`, no??
On the operator side, it's using `metrics-settings` to configure FLP (so, changing this would be a breaking change are require operator PR)

KalmanMeth: I'm fine either way. Let's just make a decision and try to be consistent.

ronensc: IIUC, the conflict rises because `MetricsSettings` can be set both in the [command-line](https://github.com/netobserv/flowlogs-pipeline/blob/45ce377bc23f92307ff7d2887cd7a2067b8c98b6/cmd/flowlogs-pipeline/main.go#L146) and in the [config file](https://github.com/netobserv/flowlogs-pipeline/blob/3f1a39fc472cde350a475b43db5a07e998a0774a/pkg/config/config.go#L41).
While the convention of the command-line arguments seems to be `kebab-case` (`log-level` is the only other example), the convention of the config file is `camelCase`.
If we want to keep both conventions and avoid changing the operator's code, I think the easiest solution would be to keep the tags of the fields of `ConfigFileStruct` that are shared with the command-line arguments formatted as `kebab-case`.
And the only thing left to do is to revert the change of the command-line argument in 
https://github.com/netobserv/flowlogs-pipeline/pull/394/files#diff-704c1b33b10fb781ff935cf83d44f3e6c9a48efdcb46c008463c2223ab43717eR146

@jotak @KalmanMeth WDYT?

KalmanMeth: done

ronensc (approved):

PR #395, opened 2023-02-27
Add ingest stage to provide synthetic workload for benchmarks

KalmanMeth:
This PR introduces ingest_synthetic stage to provide a steady stream of flow logs for a configured number of simulated connections. This may be used to control the input flows to be able to perform benchmarking and memory usages of stages.
To configure the stage, add in the config file the following fields:

`
parameters:
- name: ingest_syn
  ingest:
    type: synthetic
    synthetic:
      connections: <nnn>
      batchMaxLen: <nnn>
      flowLogsPerMin: <nnn>
`


pkg/pipeline/ingest/ingest_synthetic_test.go
```diff
@@ -0,0 +1,91 @@
...
+		srcAddr  string
+		dstAddr  string
+		srcPort  int
+		dstPort  int
+		protocol int
+	}
+
+	// Start collecting flows from the ingester and ensure we have the specified number of distinct connections
+	connectionMap := make(map[connection]int)
+	for i := 0; i < (3 * connections); i++ {
```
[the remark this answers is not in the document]
KalmanMeth: It's just to have (many) more flow logs than connections, and to verify that we accumulate the proper number of connections with multiple flow logs per connection.

pkg/pipeline/ingest/ingest_synthetic.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/netobserv/flowlogs-pipeline/pkg/operational"
+	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/utils"
+	"github.com/prometheus/client_golang/prometheus"
+	log "github.com/sirupsen/logrus"
+)
+
+type IngestSynthetic struct {
+	params           api.IngestSynthetic
+	exitChan         <-chan struct{}
+	metricsProcessed prometheus.Counter
```
ronensc, line 34: Please rename `metricsProcessed` here as well

pkg/pipeline/ingest/ingest_synthetic.go
```diff
@@ -0,0 +1,128 @@
...
+				subBatch := flowLogs[next : next+subBatchLen]
+				ingestS.sendBatch(subBatch, out)
+				ingestS.metricsProcessed.Add(float64(subBatchLen))
+				flowsLeft -= subBatchLen
+				next += subBatchLen
+				if subBatchLen == remainder {
+					next = 0
+					subBatchLen = flowsLeft
+				}
+			}
```
ronensc, line 94: I explored a different approach to achieve the goal of this loop:
https://github.com/ronensc/flowlogs-pipeline/blob/4266dc2e507e89c007a521d6eaeb7bc34de64abd/pkg/pipeline/ingest/ingest_synthetic.go#L73-L81

Please let me know if it is indeed equivalent and if you think it is clearer.
Since we loop over the sub-batch anyway in `sendBatch()` I incorporated it into the loop and removed `sendBatch()`.

ronensc (commented):
Thanks for addressing my feedback and adding the comments

pkg/pipeline/ingest/ingest_synthetic.go
KalmanMeth: your suggestion is much simpler and cleaner. Done.

ronensc (approved):

PR #399, opened 2023-03-02
Move prometheus server to main

KalmanMeth:
Even if we have no encode_prom, define the prometheus endpoint so that operational metrics can be reported.
Define prefix for operational metrics.
Remove addr:port from encode_prom definition.
Allow multiple instantiations of encode_prom with different parameters.
All encode_prom instances use the same global prometheus endpoint.


pkg/pipeline/utils/prom_server.go
```diff
@@ -0,0 +1,47 @@
...
+	"os"
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/api"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/sirupsen/logrus"
+)
+
+// StartServer listens for prometheus resource usage requests
+func StartPromServer(tlsConfig *api.PromTLSConf, server *http.Server) {
+	logrus.Debugf("entering startServer")
```
jotak, line 31: nit: doc mismatch `StartServer` => `StartPromServer`, and debug log `startServer` => `StartPromServer`

pkg/pipeline/utils/prom_server.go
```diff
@@ -0,0 +1,47 @@
...
+
+	var err error
+	if tlsConfig != nil {
+		err = server.ListenAndServeTLS(tlsConfig.CertPath, tlsConfig.KeyPath)
+	} else {
+		err = server.ListenAndServe()
+	}
+	if err != nil && err != http.ErrServerClosed {
+		logrus.Errorf("error in http.ListenAndServe: %v", err)
+		os.Exit(1)
```
jotak, line 45: I saw that you kept the same behaviour, but surely it would make more sense to not exit when `metricsSettings.NoPanic` is true?

jotak:
Thanks @KalmanMeth for this! :)

jotak:
@KalmanMeth the PR looks good to me, just a few minor remarks above.
Since it's a breaking change we must put it on hold because changes are necessary on the operator side as well
/hold

pkg/pipeline/utils/prom_server.go
KalmanMeth: done

pkg/pipeline/utils/prom_server.go
KalmanMeth: done

jotak (approved):
lgtm
restarted the e2e test that failed with a timeout

jotak:
@KalmanMeth I mentioned this shouldn't be merged before the operator PR is ready to address the breaking changes; currently operator deployment is broken, so we need to do either a quick fix on the operator side, or rollback this one

KalmanMeth:
@jotak Sorry for the misunderstanding. I saw that you had removed the the no-merge label and added the label lgtm.

jotak:
the do-not-merge label is still there :)
no worries .. should be easy to fix in the operator

PR #401, opened 2023-03-07
Increasing test coverage

jotak:
- k8s enrichment
- convert utils


pkg/pipeline/transform/kubernetes/kubernetes_test.go
```diff
@@ -47,50 +47,161 @@ func (indexMock *IndexerMock) ByIndex(indexName, indexedValue string) ([]interfa
...
 	args := informerMock.Called()
 	return args.Get(0).(cache.Indexer)
 }
 
-func TestGetInfoPods(t *testing.T) {
-	kubeData := KubeData{}
-	// pods informer
-	pidx := IndexerMock{}
-	pidx.On("ByIndex", IndexIP, "1.2.3.4").Return([]interface{}{&Info{
+func (m *IndexerMock) mockPod(ip, name, namespace, nodeIP string, owner *Owner) {
```
ronensc, line 60: There is inconsistency in the receiver variable name of `*IndexerMock`. Sometimes it's `indexMock` and sometimes it's `m`. I think it'll make sense to rename the instances of `indexMock` to `m` (or vice versa).

pkg/pipeline/transform/kubernetes/kubernetes_test.go
```diff
@@ -47,50 +47,161 @@ func (indexMock *IndexerMock) ByIndex(indexName, indexedValue string) ([]interfa
 	return args.Get(0).([]interface{}), args.Error(1)
 }
```
ronensc, line 48: I see there is a `kubernetes-mock.go` file. Would it make sense to move the mocks defined in this file to that file?

pkg/utils/convert_test.go
```diff
@@ -0,0 +1,115 @@
...
+		input:   time.Duration(42),
+		wantf64: 42.0,
+		wantu32: 42,
+		wantu64: 42,
+		wanti64: 42,
+		wanti:   42,
+	}}
+	for _, tc := range cases {
+		tc := tc
+		t.Run(fmt.Sprintf("%t", tc.input), func(t *testing.T) {
```
ronensc, line 93: `%t` prints the word true or false. Perhaps `%T` was meant here? (`%T` prints a Go-syntax representation of the type of the value)
Reference:
https://pkg.go.dev/fmt#hdr-Printing

pkg/utils/convert_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestConvertToFloat64(t *testing.T) {
```
ronensc, line 11: Since this tests other functions besides `ConvertToFloat64()`, I would rename it to something like `TestConvertToNumericTypes()`

ronensc (commented):
Thanks for this PR!
And sorry for not adding these tests myself... 😳

pkg/utils/convert_test.go
```diff
@@ -0,0 +1,115 @@
...
+	}, {
+		input:   float32(1.1),
+		wantf64: 1.1,
+		wantu32: 1,
+		wantu64: 1,
+		wanti64: 1,
+		wanti:   1,
+	}, {
+		input:   "1",
+		wantf64: 1,
```
KalmanMeth, line 36: Why is wantf64 equal to 1 and not equal to 1.0?

pkg/utils/convert_test.go
```diff
@@ -0,0 +1,115 @@
...
+	}, {
+		input:   time.Duration(42),
+		wantf64: 42.0,
+		wantu32: 42,
+		wantu64: 42,
+		wanti64: 42,
+		wanti:   42,
+	}}
+	for _, tc := range cases {
+		tc := tc
```
KalmanMeth, line 92: What does this line contribute?

pkg/utils/convert_test.go
jotak: oops good catch ... for some reason it wasn't writing true/false, but the some garbage AND the type so I didn't pay much attention

pkg/utils/convert_test.go
jotak: it's not really useful unless we make tests run in parallel at some point (it captures the variable in the more narrow scope). I can remove it

jotak:
thanks @ronensc - feedback should be addressed

ronensc (approved):

PR #404, opened 2023-03-14
NETOBSERV-924 adapt to new ovn annotation format

jotak:
The format of this annotation has changed between ocp 4.12 and 4.13 This kind of IPs are typically used in host-network traffic

### DIFF
--- a/pkg/pipeline/transform/kubernetes/cni/ovn_kubernetes.go
+++ b/pkg/pipeline/transform/kubernetes/cni/ovn_kubernetes.go
@@ -44,19 +44,9 @@ func AddOvnIPs(ips []string, node *v1.Node) []string {
 }
 
 func unmarshalAnnotation(annot []byte) (string, error) {
-	// Depending on OVN (OCP) version, the annotation might be JSON-encoded as a string, or an array of strings
-	// Try first as string (legacy)
-	var subnetsAsString map[string]string
-	err := json.Unmarshal(annot, &subnetsAsString)
-	if err == nil {
-		if subnet, ok := subnetsAsString["default"]; ok {
-			return subnet, nil
-		}
-		return "", fmt.Errorf("unexpected content for annotation %s: %s", ovnSubnetAnnotation, annot)
-	}
-
+	// Depending on OVN (OCP) version, the annotation might be JSON-encoded as a string (legacy), or an array of strings
 	var subnetsAsArray map[string][]string
-	err = json.Unmarshal(annot, &subnetsAsArray)
+	err := json.Unmarshal(annot, &subnetsAsArray)
 	if err == nil {
 		if subnets, ok := subnetsAsArray["default"]; ok {
 			if len(subnets) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/pipeline/transform/kubernetes/cni/ovn_kubernetes.go
```diff
@@ -43,27 +43,49 @@ func AddOvnIPs(ips []string, node *v1.Node) []string {
...
+		if err != nil {
+			return "", err
+		}
+		ip4 := ip0.To4()
+		if ip4 == nil {
+			// TODO: what's the rule with ipv6?
+			return "", nil
 		}
-		return "", fmt.Errorf("unexpected content for annotation %s: %s", ovnSubnetAnnotation, subnetsJSON)
+		return fmt.Sprintf("%d.%d.%d.2", ip4[0], ip4[1], ip4[2]), nil
```
msherif1234, line 88: what is `.2` ? maybe u can use `fmt.Sprintf("%s",net.IPv4(ip4))`

jotak: The IP here is the first one of the range provided by the subnet that is read above; e.g. we read the subnet `10.128.0.0/23` ; then it's a tacit rule with ovn-k that the `.2` ip from that subnet is used for the mp0 interface, which is the one we want to track here. So we need somehow to convert `10.128.0.0/23` into `10.128.0.2`.

OlivierCazade (dismissed):

pkg/pipeline/transform/kubernetes/cni/ovn_kubernetes.go
```diff
@@ -43,27 +43,49 @@ func AddOvnIPs(ips []string, node *v1.Node) []string {
...
+func unmarshalAnnotation(annot []byte) (string, error) {
+	// Depending on OVN (OCP) version, the annotation might be JSON-encoded as a string, or an array of strings
+	// Try first as string (legacy)
+	var subnetsAsString map[string]string
+	err := json.Unmarshal(annot, &subnetsAsString)
+	if err == nil {
+		if subnet, ok := subnetsAsString["default"]; ok {
+			return subnet, nil
+		}
+		return "", fmt.Errorf("unexpected content for annotation %s: %s", ovnSubnetAnnotation, annot)
```
stleerh, line 55: (Minor) Wouldn't it be better to try the new way first and then if it fails, try legacy?

pkg/pipeline/transform/kubernetes/cni/ovn_kubernetes.go
stleerh: That will work only if you can assume the suffix is never greater than 24.  Otherwise, it should add 2 or | 2.

jotak: From past discussions, this is an assumption that we can make - although there's no guarantee that it won't change.

jpinsonneau (approved):

PR #415, opened 2023-03-29
conntrack: Fix a bug in swapAB

ronensc:
When implementing the swapAB logic I missed the fact that in addition to swapping the values of the connection source and destination, their hashes should be swapped as well. Without this, aggregates such as `Bytes_AB` and `Bytes_BA` would be wrong.
I realized that I had the same bug in the unit test the was supposed to test it.
This PR fixes the problem and the test.

Once merged, this will allow us to spot connections that their source and destination were swapped by inspecting their `newConnection` record. The `Bytes_AB` would be 0 while `Bytes_BA` would be >0.

### DIFF
--- a/pkg/pipeline/extract/conntrack/conn.go
+++ b/pkg/pipeline/extract/conntrack/conn.go
@@ -180,6 +180,9 @@ func NewConnBuilder(metrics *metricsType) *connBuilder {
 }
 
 func (cb *connBuilder) Hash(h totalHashType) *connBuilder {
+	if cb.shouldSwapAB {
+		h.hashA, h.hashB = h.hashB, h.hashA
+	}
 	cb.conn.hash = h
 	return cb
 }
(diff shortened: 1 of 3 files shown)


jpinsonneau (approved):
Good catch @ronensc ! Code LGTM

Would this bug result in duplicated connections coming from the same FLP ? I don't think I've seen such

ronensc:
> Good catch @ronensc ! Code LGTM
> 
> Would this bug result in duplicated connections coming from the same FLP ? I don't think I've seen such

I don't think this bug would result in duplicated connections

jpinsonneau:
@ronensc @jotak we should backport this in [release-1.2](https://github.com/netobserv/flowlogs-pipeline/tree/release-1.2)

memodi:
> - @memodi is that ok with you?

Yes, I am fine. @Amoghrd -  I vaguely remember you mentioned about noticing something similar..

PR #421, opened 2023-04-11
NETOBSERV-973 skip duplicates

jpinsonneau:
This PR skip duplicated flows in connection tracking instead of only skipping aggregates

Backport: https://github.com/netobserv/flowlogs-pipeline/pull/420


jotak:
/lgtm
but can you add more info to the PR description please? (why moving it to a different level is better)

PR #424, opened 2023-04-14
NETOBSERV-974 Implement SASL for Kafka (producer+consumer)

jotak:


pkg/pipeline/utils/sasl.go
```diff
@@ -0,0 +1,40 @@
...
+	"os"
+	"strings"
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/api"
+	"github.com/segmentio/kafka-go/sasl"
+	"github.com/segmentio/kafka-go/sasl/plain"
+	"github.com/segmentio/kafka-go/sasl/scram"
+)
+
+func SetupSASLMechanism(cfg *api.SASLConfig) (sasl.Mechanism, error) {
```
OlivierCazade, line 14: nitpick: Since the function is tied to the kafka-go type `sasl.Mechanism` may be it should be in the kafka ingest file.

OlivierCazade (approved):
LGTM, only a minor comment.

KalmanMeth (approved):

Issue #423, opened 2023-04-13
Align docker images used in multi-stage builds


jvidalallende:
According to https://github.com/netobserv/flowlogs-pipeline/blob/main/contrib/docker/Dockerfile, this multi-stage docker build is using the [golang](https://hub.docker.com/_/golang/) image to build the binary (which is based on Debian bullseye), and [ubi9-minimal](https://catalog.redhat.com/software/containers/ubi9-minimal/61832888c0d15aff4912fe0d) as runtime.

When building golang binaries, if the `net` or `os/user` (among others) are part of the build, either direcly or indirectly, the resulting binary is dynamically linked. You can verify that by running `ldd <binary>`.

Since the resulting binary depends on the builders's `libc`, this can be an issue if there is a version mismatch on the runtime's `libc`. This has already been seen when using `ubi8-minimal` as runtime for binaries built on `golang:1.19`.

To prevent this from happenning, the suggestion is to use a RHEL9-based distribution as builder. You can easily get the same build environment as in the `golang:1.19` image from a `ubi9-minimal` image:

```
FROM registry.access.redhat.com/ubi9/ubi-minimal:9.1 AS builder

ARG GO_VERSION

RUN microdnf install -y --nodocs --setopt=install_weak_deps=0 --disableplugin=subscription-manager \
        make git gcc tar jq which findutils\
    && microdnf clean all --disableplugin=subscription-manager

RUN curl -fsSL https://go.dev/dl/go${GO_VERSION}.linux-amd64.tar.gz | tar -C /usr/local -xz \
    && ln -s /usr/local/go/bin/go /usr/bin/go \
    && ln -s /usr/local/go/bin/gofmt /usr/bin/gofmt
```

This expects a build argument `GO_VERSION` to be passed to the image (e.g. `docker build -b 'GO_VERSION=1.19.8'...`). This should allow you to easily get updates on newer language versions.

If the builder image does not change frequently, you may consider pushing it to some registry and afterwards pull it on builds to save some build time.

eranra:
also cc: @jotak @jpinsonneau @OlivierCazade

PR #426, opened 2023-04-20
NETOBSERV-994 FLP multiarch consistency

jpinsonneau:
These changes brings consistency between netobserv components

- added shortcuts
- use the same target for multi arch / single arch relying on `MULTIARCH_TARGETS` array
- renamed variables and targets so every repo can run the same targets such as:
  - `IMAGE_ORG=jpinsonn VERSION=multiarch make images` to push all images and manifest
  - `MULTIARCH_TARGETS=amd64 make image-build image-push` to push a single amd64 `image to quay.io/netobserv/flowlogs-pipeline:latest-adm64`


jpinsonneau:
> @jpinsonneau did you see this issue: #423 ? I was wondering if that could also impact what you're doing (I don't think so, but haven't had time to look closely)

@jotak I suggest to address this in a separate PR
We need to create our own docker image, save it to netobserv quay (I can't push as maintainer :crying_cat_face: ) and then update the builder images in all of our components passing `--build-arg GO_VERSION=1.19.8` in `build_target` functions

Makefile
```diff
@@ -119,215 +153,49 @@ benchmarks: $(BENCHSTAT) validate_go ## Benchmark
...
+# note: to build and push custom image tag use: IMAGE_ORG=myuser VERSION=dev make images
+.PHONY: image-build
+image-build: ## Build MULTIARCH_TARGETS images
+	trap 'exit' INT; \
+	$(foreach target,$(MULTIARCH_TARGETS),$(call build_target,$(target)))
+
+.PHONY: image-push
+image-push: ## Push MULTIARCH_TARGETS images
+	trap 'exit' INT; \
+	$(foreach target,$(MULTIARCH_TARGETS),$(call push_target,$(target)))
```
msherif1234, line 167: how long it takes before and after this change to build and push images 
I am curious if that takes much longer time we could have multi-target build option and just leave the current options for the default target ?
depends on how frequently we need to build for all targets do we do that when release new version or on every PR ?

jpinsonneau: It's mainly 3 times longer as `MULTIARCH_TARGETS` contains `amd64 arm64 ppc64le` currently. The manifest creation is super fast and can even be skipped if you point a single arch image.
I'm fine setting `amd64` only as default and letting a comment with example to build the 3 archs.

OlivierCazade (approved):
LGTM

PR #429, opened 2023-04-29
NETOBSERV-1063: Add TCP drop & DNS tracking to flowlogs-pipeline

msherif1234:
add TCPDrop and DNS tracking support
flatten ICMP fields


pkg/pipeline/decode/decode_protobuf.go
```diff
@@ -37,26 +37,35 @@ func PBFlowToMap(flow *pbflow.Record) config.GenericMap {
...
+		"Interface":         flow.Interface,
+		"AgentIP":           ipToStr(flow.AgentIp),
+		"Flags":             flow.Flags,
+		"IcmpType":          flow.Icmp.GetIcmpType(),
+		"IcmpCode":          flow.Icmp.GetIcmpCode(),
+		"TcpDropBytes":      flow.TcpDropBytes,
+		"TcpDropPackets":    flow.TcpDropPackets,
+		"TcpDropFlags":      flow.GetTcpDropFlags(),
+		"TcpDropState":      tcpStateToStr(flow.GetTcpDropState()),
+		"TcpDropCause":      tcpDropCauseToStr(flow.GetTcpDropCause()),
```
jpinsonneau, line 64: :+1: that's perfect to centralize the code ! 

Can we also do that with ICMP Type & Code fields above ?
Should we also do that for Protocol ?

jpinsonneau (approved):

PR #430, opened 2023-05-03
NETOBSERV-915 skip reinterpret direction for conversations

jpinsonneau:
Skip `reinterpretDirection` for conversation events as source and destination can be swapped by TCP flag `SYN_ACK` interpretation.

Related PRs:
- https://github.com/netobserv/network-observability-operator/pull/313
- https://github.com/netobserv/network-observability-console-plugin/pull/317

### DIFF
--- a/pkg/pipeline/transform/transform_network.go
+++ b/pkg/pipeline/transform/transform_network.go
@@ -151,7 +151,10 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
 				}
 			}
 		case api.OpReinterpretDirection:
-			reinterpretDirection(outputEntry, &n.DirectionInfo)
+			// only reinterpret direction on flowlogs
+			if rt, ok := outputEntry["_RecordType"]; !ok || rt == "flowLog" {
+				reinterpretDirection(outputEntry, &n.DirectionInfo)
+			}
 		case api.OpAddIPCategory:
 			if strIP, ok := outputEntry[rule.Input].(string); ok {
 				cat, ok := n.ipCatCache.GetCacheEntry(strIP)


jotak:
Just to clarify: are we safe doing so just because the "Reporter" field is anyway disabled in UI for Conversation views?

I'm asking because, else, it's not very clear to me why we would skip reinterpret for conntrack. It was added because the concept of INGRESS/EGRESS direction, provided be the ebpf agent, is bound to an interface and not to the whole node as expected with our "Reporter" definition (well explained by Mario here: https://issues.redhat.com/browse/NETOBSERV-696?focusedId=21315152&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21315152 )
image

So the "reinterpret" is a way to "normalize" what we consider as ingress or egress, and I don't see why it should be different with conntrack events - but if the "Reporter" setting is disabled anyway for conntrack, then I guess that's ok

jotak:
/lgtm
but still curious about the explanation :)

jpinsonneau:
> Just to clarify: are we safe doing so just because the "Reporter" field is anyway disabled in UI for Conversation views?

The goal is to enable it https://github.com/netobserv/network-observability-operator/pull/313

> I'm asking because, else, it's not very clear to me why we would skip reinterpret for conntrack. It was added because the concept of INGRESS/EGRESS direction, provided be the ebpf agent, is bound to an interface and not to the whole node as expected with our "Reporter" definition (well explained by Mario here: https://issues.redhat.com/browse/NETOBSERV-696?focusedId=21315152&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21315152 ) image
> 
> So the "reinterpret" is a way to "normalize" what we consider as ingress or egress, and I don't see why it should be different with conntrack events - but if the "Reporter" setting is disabled anyway for conntrack, then I guess that's ok

As far as I have seen in my tests, it conflicts with the [swapAB capability](https://github.com/netobserv/flowlogs-pipeline/blob/9740aa3ec29e7b78de2a602c3fbc915e9eb74ed7/pkg/api/conntrack.go#L99) as it just update src / dst but doesn't affect Direction / IfDirection.
Keeping the original values avoid mixing unconsistent check on [srcNode == reporter](https://github.com/netobserv/flowlogs-pipeline/blob/e6e77158b956a31b052459dfa492eb806cb0a9dd/pkg/pipeline/transform/transform_network_direction.go#L55).
Some alternatives could be:
- disable `swapAB`
- keep track of swaps to manage this case in `reinterpretDirection`
- update `Direction` when swapping (may need hardcoded checks or will improve code complexity / config) 

> PS: I wonder if doing the "reinterpret" _before_ processing conntrack wouldn't have also fixed the issues that we had
It would; but today everything is mixed inside the `enrich` stage that must be after connection tracking. Keeping this reinterpret before would force us to add an extra stage at first and the actual enrich. It will impact performances.

Why don't we do that reinterpretation in eBPF agent ?

PR #431, opened 2023-05-04
Remove error logs in processing loop

jotak:
Errors are provided as metrics only to minimize the impact on resources, and logs readability, when an error is thrown repeatedly.

### DIFF
--- a/pkg/pipeline/ingest/ingest_grpc.go
+++ b/pkg/pipeline/ingest/ingest_grpc.go
@@ -105,7 +105,8 @@ func instrumentGRPC(m *metrics) grpc2.UnaryServerInterceptor {
 
 		resp, err = handler(ctx, req)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


jpinsonneau:
> Errors are provided as metrics only to minimize the impact on resources, and logs readability, when an error is thrown repeatedly.

I disagree with this approach especially at current project state. We should give qe and end users all the necessary info to identify bugs and help us to fix these.
If you don't want to impact resources, we can simply update these to `debug` or `trace` and document how to report a bug with a step including this change.

jotak:
@jpinsonneau So we can indeed move the logs as "trace" or "debug" but the value is low since we discourage users from setting such as level, for performance reasons. QE also don't use that in perf tests. Metrics are really the best way to get these errors.

OlivierCazade (approved):

jpinsonneau (approved):

PR #436, opened 2023-05-23
fix github ci image argument

jpinsonneau:
@msherif1234 it seems we missed that one :smile_cat:


jotak:
/lgtm
(ideally I think we should merge the manifest build/push targets inside the image build/push targets?)

PR #443, opened 2023-06-15
changed error to warning

KalmanMeth:
If we use multiple encode_prom stages, each instance tries to register its operational metrics. This results in a prometheus.AlreadyRegisteredError, which is not really an error, so we convert it to a warning.


jpinsonneau (approved):
LGTM thanks @KalmanMeth !

PR #447, opened 2023-06-25
initial support for labels

KalmanMeth:


docs/api.md
```diff
@@ -148,6 +148,7 @@ Following is the supported API format for network transformations:
                  type: (enum) one of the following:
                      add_regex_if: add output field if input field satisfies regex pattern from parameters field
                      add_if: add output field if input field satisfies criteria from parameters field
+                     add_label_if: add output field to list of labels if input field satisfies criteria from parameters field
```
jotak, line 151: It's something we should have raised earlier as it also concerns "add_regex_if" and "add_if",  but this doesn't look like a network-specific transformation, shouldn't that rather go in generic_transform?

jotak:
The code looks good, but is that part of the changes needed for multi-cloud? I don't really see how that will be used, can you elaborate on that?

docs/api.md
KalmanMeth: @jotak Should we do this in a separate PR? We would move add_if, add_regex_if, add_label_if to generic transform.

KalmanMeth:
> Second thing that is more a concern to me, is the performance impact implied by this evaluation library compared to a more specific implementation.

Do you think we should add a cache so we only have to perform each evaluation once, and then future identical comparisons will be obtained from the cache?

docs/api.md
jotak: yeah no problem with me to do that later

jotak:
@KalmanMeth at the moment I still don't see how that's going to be used in the context of k8s metadata labels, so it's hard for me to review/ answer. Especially I don't get what's going to be set in the "If" part of that filter. Can you provide an example?

docs/api.md
eranra: @KalmanMeth if you can open an issue so we will not forget? This will be a "Breaking changes" one

pkg/pipeline/transform/transform_network.go
```diff
@@ -81,6 +83,17 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
 				}
 				outputEntry[rule.Output+"_Evaluate"] = true
 			}
+		case api.OpAddLabelIf:
```
eranra, line 86: @KalmanMeth this is an "expensive" operation and I envision that we will also want to add "static" labels so can you add another operation that is not with "if" but just a simple AddLabel operation?

docs/api.md
eranra: @KalmanMeth I see that the way you set the value in the code is `labels[rule.Output] = rule.c` --- can you extend the text to explain that output =key and Assignee=value

eranra (commented):
@jpinsonneau @KalmanMeth The main question and value proposition for this "labels" approach is if we can search a subset of the key=value labels from the list using some Loki query capabilities. @KalmanMeth did you check that manually somehow? @jpinsonneau @jotak any inputs from your POV on that ( one thing I thought about was the wrap the labels with json curly brachets, but I am not sure if this is needed or will help )

jotak:
@eranra @KalmanMeth : Initially, here's how I would have seen a label mapper:

Something that is part of kube enrichment, configured like that:

```yaml
- name: enrich
  transform:
    type: network
    network:
      rules:
      - input: SrcAddr
        output: SrcK8S
        type: add_kubernetes
        labelMap:
          fromNode:
          - match: "topology\.kubernetes\.io/.*"
            output: NodeLabels
          fromPod:
          - match: ".*"
            output: PodLabels
```

and for instance when applied to a pod that is running on a node that has the following labels:
```
kubernetes.io/arch=amd64
kubernetes.io/hostname=ip-10-0-139-105.eu-west-3.compute.internal
kubernetes.io/os=linux
node-role.kubernetes.io/worker=
node.kubernetes.io/instance-type=m6i.xlarge
node.openshift.io/os_id=rhcos
topology.ebs.csi.aws.com/zone=eu-west-3a
topology.kubernetes.io/region=eu-west-3
topology.kubernetes.io/zone=eu-west-3a
```

... it would result in json:

```
{
SrcK8S_Name: "my-app",
...
SrcK8S_NodeLabels: "topology.kubernetes.io/region=eu-west-3,topology.kubernetes.io/zone=eu-west-3a",
SrcK8S_PodLabels: "app=my-app,controller-revision-hash=5ccbdbc54,pod-template-generation=1,version=main",
}

```


But tbh, while I think it would be nice to have that, I also see that it would come at a higher cost than labels statically added from the operator when they are supposed to be constant in a given cluster, such as the `topology.kubernetes.io/region` / `zone` listed above. So maybe for these ones, which are the ones interesting for our multi-cluster work, simply adding static labels as text in the same was as for the cluster ID (without "If") would not only be good enough, but also more performant

jotak:
ie using the same "add_field_if_doesnt_exist" like here:  https://github.com/netobserv/network-observability-operator/pull/386/files#diff-c761638f65b982f9ecf3717616920a5a3c0bb1670fd8ee7e9ab51afc4f166581R578-R580

... but containing a string expression of some chosen labels instead of the cluster id

KalmanMeth:
I moved the add_labels to transform_filter, now that we have also moved the add_if etc to transform_filter.
I also added a generic add_label (in addition to add_label_if) and a generic add_field which add the label/field directly without performing tests on existing fields.

pkg/pipeline/transform/transform_network.go
KalmanMeth: done

pkg/api/transform_filter.go
```diff
@@ -27,9 +27,12 @@ type TransformFilterOperationEnum struct {
...
+	AddField                 string `yaml:"add_field" json:"add_field" doc:"adds (input) field to the entry; overrides previous value if present (key=input, value=value)"`
 	AddFieldIfDoesntExist    string `yaml:"add_field_if_doesnt_exist" json:"add_field_if_doesnt_exist" doc:"adds a field to the entry if the field does not exist"`
 	AddFieldIf               string `yaml:"add_field_if" json:"add_field_if" doc:"add output field set to assignee if input field satisfies criteria from parameters field"`
 	AddRegExIf               string `yaml:"add_regex_if" json:"add_regex_if" doc:"add output field if input field satisfies regex pattern from parameters field"`
+	AddLabel                 string `yaml:"add_label" json:"add_label" doc:"add (input) field to list of labels with value taken from Value field (key=input, value=value)"`
+	AddLabelIf               string `yaml:"add_label_if" json:"add_label_if" doc:"add output field to list of labels with value taken from assignee field if input field satisfies criteria from parameters field"`
```
ronensc, line 35: nitpick: some doc strings are written in imperative mood `"add (input) field..."`, and some in present simple `"adds (input) field..."`. Which is preferred?

ronensc (approved):

jotak (approved):

PR #448, opened 2023-06-29
NETOBSERV-1131: need to allow setting multiple filters

jotak:
Improve the promEncode API to allow setting multiple filters. To avoid breaking change, the existing filter API is kept and marked as deprecated.

Operator PR: https://github.com/netobserv/network-observability-operator/pull/387

### DIFF
--- a/pkg/api/encode_prom.go
+++ b/pkg/api/encode_prom.go
@@ -41,12 +41,20 @@ func PromEncodeOperationName(operation string) string {
 }
 
 type PromMetricsItem struct {
-	Name     string            `yaml:"name" json:"name" doc:"the metric name"`
-	Type     string            `yaml:"type" json:"type" enum:"PromEncodeOperationEnum" doc:"one of the following:"`
-	Filter   PromMetricsFilter `yaml:"filter" json:"filter" doc:"an optional criterion to filter entries by"`
-	ValueKey string            `yaml:"valueKey" json:"valueKey" doc:"entry key from which to resolve metric value"`
-	Labels   []string          `yaml:"labels" json:"labels" doc:"labels to be associated with the metric"`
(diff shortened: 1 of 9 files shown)


pkg/api/encode_prom.go
```diff
@@ -41,12 +41,20 @@ func PromEncodeOperationName(operation string) string {
...
+	ValueKey string              `yaml:"valueKey" json:"valueKey" doc:"entry key from which to resolve metric value"`
+	Labels   []string            `yaml:"labels" json:"labels" doc:"labels to be associated with the metric"`
+	Buckets  []float64           `yaml:"buckets" json:"buckets" doc:"histogram buckets"`
+}
+
+func (i *PromMetricsItem) GetFilters() []PromMetricsFilter {
+	if len(i.Filters) == 0 && i.Filter.Key != "" {
+		return []PromMetricsFilter{i.Filter}
+	}
+	return i.Filters
```
jpinsonneau, line 57: Can't we just append i.Filter to i.Filters if specified ? So any config would work

```suggestion
func (i *PromMetricsItem) GetFilters() []PromMetricsFilter {
	if i.Filter.Key != "" {
		return append(i.Filters, i.Filter)
	}
	return i.Filters
```

jotak: that would allow weird configuration where people can use both the new and the deprecated fields ... the code looks simpler indeed, but in terms of  config/ux I think I prefer not to allow weird things like that ("principle of least surprise")

jotak:
I'll merge this one as it isn't user facing without https://github.com/netobserv/network-observability-operator/pull/387
- pre-merge tests can be done on https://github.com/netobserv/network-observability-operator/pull/387

KalmanMeth (approved):

PR #450, opened 2023-07-06
NETOBSERV-1017 extract timebased indexKeys

jpinsonneau:
In order to get [top Pod / Service](https://issues.redhat.com/browse/NETOBSERV-1017) talkers this PR implements `IndexKeys` as an array for `timebased` stage.

It also add the capability to index on non string fields using `ConvertToString` utils


pkg/pipeline/extract/timebased/filters.go
```diff
@@ -149,10 +150,21 @@ func (fs *FilterStruct) CreateGenericMap() []config.GenericMap {
...
 			"operation":        fs.Rule.OperationType,
 			"operation_key":    fs.Rule.OperationKey,
-			"key":              result.key,
+			"key":              result.values,
 			"operation_result": result.operationResult,
 		}
-		t[fs.Rule.IndexKey] = result.key
+
+		// append operation key and result as key / value
+		t[fs.Rule.OperationKey] = result.operationResult
```
jpinsonneau, line 158: I feel more user friendly to output the `<operation_key>: <operation_result>` as output like `"Bytes":500` especially when [reusing these in prometheus encode](https://github.com/netobserv/network-observability-operator/pull/391/files#diff-2ec52703296accf2f2ab72dea73684fe5433df463b13cd15765e856a9edbe80a)

Full output example:
```
{
   "key":"10.0.0.2,11.0.0.1,0",
   "name":"BotK_SrcDst_Bytes",
   "operation":"avg",
   "operation_key":"Bytes",
   "operation_result":500,
   "Bytes":500,
   "index_key":"SrcAddr,DstAddr,Direction",
   "SrcAddr":"10.0.0.2",
   "DstAddr":"11.0.0.1",
   "Direction":"0"
}
```

Following encode example:
```
encode:
  type: prom
  prom:
    metrics:
      - name: top_pods_egress_bytes_total
        type: counter
        valuekey: Bytes
        filters:
          - key: FlowDirection
            value: "1"
        labels:
          - SrcK8S_Name
          - SrcK8S_Namespace
          - SrcK8S_Type
          - SrcK8S_HostName
```

Maybe also we can remove `operation_key` & `operation_result` (unless we want to keep compatibility with previous versions) ?

jpinsonneau: Similarly `key` could be removed from the JSON

pkg/utils/convert.go
```diff
@@ -242,3 +242,35 @@ func ConvertToBool(unk interface{}) (bool, error) {
...
+	case uint64:
+		return strconv.FormatUint(i, 10), nil
+	case uint32:
+		return strconv.FormatUint(uint64(i), 10), nil
+	case uint:
+		return strconv.FormatUint(uint64(i), 10), nil
+	case string:
+		return i, nil
+	default:
+		v := reflect.ValueOf(unk)
```
KalmanMeth, line 267: Could't we use fmt.Sprintf(%v) to convert anything to a string?

pkg/pipeline/extract/timebased/filters.go
KalmanMeth: I am not aware of who (if anybody) is using the timebased TopK feature. The specified fields were included because it was thought that they might be useful for the user, but only user feedback can give us a real indication. If you find your proposed fields useful for what you are doing, then this is the first concrete user feedback we have, and you should add whatever fields you find useful.

KalmanMeth: I think the `operation_key` should be kept to help the user remember what action was performed to produce the metric. By the time the user sees the metric, he may no longer be aware of the aggregation that was performed to produce it, so some indication in the flow log might be helpful.

pkg/pipeline/extract/timebased/filters.go
```diff
@@ -149,10 +150,21 @@ func (fs *FilterStruct) CreateGenericMap() []config.GenericMap {
...
+
+		// append index key / value pairs
+		values := strings.Split(result.values, ",")
+		if len(fs.Rule.IndexKeys) == len(values) {
+			for i, k := range fs.Rule.IndexKeys {
+				t[k] = values[i]
+			}
+		}
+
```
KalmanMeth, line 167: Once you've added in the individual fields, I don't see the need to keep the `key`, `operation_key`, and `operation_result` fields.

pkg/utils/convert.go
jpinsonneau: sounds good, https://github.com/netobserv/flowlogs-pipeline/pull/450/commits/b8cc56454d4bdcede79bdc8da1ba86531174ddd2

pkg/pipeline/extract/timebased/filters.go
jpinsonneau: Yes exactly; done in https://github.com/netobserv/flowlogs-pipeline/pull/450/commits/25842d7f598b04d13e02897c5fba9a7d7d0ddb4f

KalmanMeth (approved):

jpinsonneau:
Thanks @KalmanMeth :partying_face:

jotak:
@KalmanMeth @jpinsonneau should we mark that as a breaking change? (the API is preserved however the output fields are different if I understand correctly (operation_key and operation_result being removed)

I'm not concerned for netobserv, but other FLP users

jpinsonneau:
> @KalmanMeth @jpinsonneau should we mark that as a breaking change? (the API is preserved however the output fields are different if I understand correctly (operation_key and operation_result being removed)
> 
> I'm not concerned for netobserv, but other FLP users

Added the label on that PR. Check https://github.com/netobserv/flowlogs-pipeline/pull/450#discussion_r1264665272 discussion

KalmanMeth:
@jpinsonneau For completeness we should mark it as a breaking change (although you did maintain back compatibility for the current implementation, indicating deprecation).

PR #451, opened 2023-07-11
NETOBSERV-1186: extract DNS flags rcode field and stringify it

msherif1234:
- try compress out unused fields to reduce flp generated map size
- for dns tracker extract and decode response code so it can be used as UI filter


pkg/pipeline/decode/decode_protobuf.go
```diff
@@ -66,6 +66,7 @@ func PBFlowToMap(flow *pbflow.Record) config.GenericMap {
 		"DnsResponseTimeMs":      flow.TimeDnsRsp.AsTime().UnixMilli(),
 		"DnsId":                  flow.GetDnsId(),
 		"DnsFlags":               flow.GetDnsFlags(),
+		"DnsFlagsResponseCode":   dnsRcodeToStr(flow.GetDnsFlags() & 0xF),
```
jpinsonneau, line 69: Would it be possible to append the field only when a value is provided ?

It would save storage

msherif1234: created `Append` method and conditionally add dns or tcpdrop when they happens

jpinsonneau:
> New image: quay.io/netobserv/flowlogs-pipeline:86f203e. It will expire after two weeks.

Thanks @msherif1234 ! That works perfectly with front end.

A small change will be needed for `Drops filter` -> `Without drops` case as it was checking for `"TcpDropPackets":0` but the other cases works fine.
I'm preparing a PR for this and the new `ResponseCode` field

pkg/config/generic_map.go
```diff
@@ -42,3 +42,16 @@ func (m GenericMap) IsDuplicate() bool {
...
+	result := make(GenericMap, len(m)+len(n))
+
+	for k, v := range m {
+		result[k] = v
+	}
+	for k, v := range n {
+		result[k] = v
+	}
+	return result
+}
```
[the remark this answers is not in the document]
msherif1234: currently the use of this api for non overlapping ConfigMaps I can add a check if the k is already in the result from the 1st round then just skip it ?

pkg/pipeline/decode/decode_protobuf.go
```diff
@@ -37,36 +39,76 @@ func PBFlowToMap(flow *pbflow.Record) config.GenericMap {
...
+		"AgentIP":         ipToStr(flow.AgentIp),
+	}
+
+	proto := flow.Transport.GetProtocol()
+	if proto == syscall.IPPROTO_ICMP || proto == syscall.IPPROTO_ICMPV6 {
+		icmpOut := config.GenericMap{
+			"IcmpType": flow.GetIcmpType(),
+			"IcmpCode": flow.GetIcmpCode(),
+		}
+		out = out.Append(icmpOut)
```
jotak, line 65: out of curiosity: why here and everywhere, you don't add the fields to `out` directly?

msherif1234: no real reason I switched to direct assign to save the memory used with map allocation and removed Append method

PR #453, opened 2023-07-13
move add_if and add_regex_if to transform_filter from transform_network

KalmanMeth:


jotak:
cc @ronensc @eranra wdyt?
It would be a breaking change of course, but it sounds like some of the filters were wrongly defined as "network filters" whereas they were completely unrelated to network.
We'll need to document this change (and how to "migrate") in the release note

jotak:
@KalmanMeth could you rebase please? Some CI checks aren't triggered because you were several weeks behind `main`

jotak:
I just verified that there is no regression with netobserv (these moved rules aren't used, so that's fine)
Code looks good to me

/lgtm

KalmanMeth:
@jotak @ronensc I think this PR just needs a formal approval from a reviewer before we can merge.

PR #460, opened 2023-07-23
fixed handling of empty list of labels

KalmanMeth:


pkg/pipeline/encode/encode_prom.go
```diff
@@ -234,9 +238,11 @@ func (e *EncodeProm) extractLabelsAndKey(flow config.GenericMap, info *api.PromM
 	key.WriteRune('|')
 	for _, t := range info.Labels {
 		entryLabels[t] = ""
-		if v, ok := flow[t]; ok {
-			entryLabels[t] = fmt.Sprintf("%v", v)
+		v, ok := flow[t]
+		if !ok {
+			return nil, ""
```
jotak, line 243: I don't think we should do that: when a label is missing, it's currently set as empty, which is fine, or why would we want something different?

To take an example, we often have a bunch of metrics labelled by namespace; they contain data for both namespaced flows (e.g. pod to pod) and non-namespace flows (e.g. node to node: here the namespace labels are empty) and even hybrid pod-to-node flows. With this change we would loose many records from those metrics.

Issue #461, opened 2023-07-24
FLP go validation fails for me on MAC.


dushyantbehl:
![image](https://github.com/netobserv/flowlogs-pipeline/assets/1579780/baed45c5-79fc-44d8-a014-963615ad9384)

PR #462, opened 2023-07-24
Fix go validation check

dushyantbehl:
Fixes https://github.com/netobserv/flowlogs-pipeline/issues/461


jotak:
like I said on slack, I wonder if we should keep maintaining this MIN_GO_VERSION, as it sounds the same as the built-in `go.mod` go version

dushyantbehl:
More discussion here..will see what to do next. 
https://github.com/netobserv/netobserv-ebpf-agent/pull/164

PR #463, opened 2023-07-25
Add test for missing labels

jotak:
@KalmanMeth here's a test that clarifies the behaviour that IMO we should keep, wrt to missing labels and https://github.com/netobserv/flowlogs-pipeline/pull/460


jpinsonneau (approved):
LGTM :+1: 

Is it related to a specific story ?

jotak:
@jpinsonneau that was related to this ticket https://github.com/netobserv/flowlogs-pipeline/issues/464 where the proposed fix was to ignore counting records for metrics when there are missing labels, but that would have broken a bunch of things, hence this added test to write in stone the expected behaviour here.
Thanks for the review!
/approve

PR #468, opened 2023-08-01
NETOBSERV-1111 Change RTT output to ns

dushyantbehl:


jpinsonneau (approved):
LGTM thanks @dushyantbehl

PR #469, opened 2023-08-01
NETOBSERV-1101: support [Mutual] TLS for grpc server

jotak:
Related PRs:
- eBPF Agent: https://github.com/netobserv/netobserv-ebpf-agent/pull/166
- Operator: WIP

**Breaking change:**
The Client TLS configuration API changes (previously used only for Kafka):
- In Go, struct is renamed from ` ClientTLS` to `TLSConfig`
- Metrics server TLS now uses this same `TLSConfig` API
- New field `type` can be `none`, `simple` or `mutual`. If you were previously using TLS, set it to `simple` for 1-1 migration. Empty defaults to `none`.
- JSON  (and Go) fields are renamed:
  - `userCertPath` => `certPath`
  - `userKeyPath` => `keyPath`


jotak:
Since mtls is quite symetric, I wonder if it still makes sense to have dedicated structs for ClientTLS and ServerTLS, or if we should merge them ...

PR #470, opened 2023-08-04
NETOBSERV-1208 & NETOBSERV-1233 Aggregators skip missing fields

jpinsonneau:
Following https://github.com/netobserv/flowlogs-pipeline/pull/451 we removed unused fields to the genericMap by default.

Aggregators on connection tracking needs to be updated to:
- avoid reporting unecessary errors
- be able to keep last provided field

This PR introduce a new `ReportMissing` field optional to allow error reporting on aggregations.

Please let me know if you want to have a different default behavior or a more flexible solution.

### DIFF
--- a/pkg/pipeline/extract/conntrack/aggregator.go
+++ b/pkg/pipeline/extract/conntrack/aggregator.go
@@ -37,11 +37,12 @@ type aggregator interface {
 }
 
 type aggregateBase struct {
-	inputField  string
-	outputField string
-	splitAB     bool
-	initVal     interface{}
-	metrics     *metricsType
+	inputField    string
+	outputField   string
+	splitAB       bool
+	initVal       interface{}
+	metrics       *metricsType
+	reportMissing bool
 }
 
 type aSum struct{ aggregateBase }
@@ -64,7 +65,7 @@ func newAggregator(of api.OutputField, metrics *metricsType) (aggregator, error)
 	} else {
 		inputField = of.Name
 	}
-	aggBase := aggregateBase{inputField: inputField, outputField: of.Name, splitAB: of.SplitAB, metrics: metrics}
+	aggBase := aggregateBase{inputField: inputField, outputField: of.Name, splitAB: of.SplitAB, metrics: metrics, reportMissing: of.ReportMissing}
 	var agg aggregator
 	switch of.Operation {
 	case api.ConnTrackOperationName("Sum"):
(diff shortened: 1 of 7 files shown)


pkg/pipeline/extract/conntrack/aggregator.go
```diff
@@ -109,10 +110,15 @@ func (agg *aggregateBase) getOutputField(d direction) string {
 func (agg *aggregateBase) getInputFieldValue(flowLog config.GenericMap) (float64, error) {
 	rawValue, ok := flowLog[agg.inputField]
 	if !ok {
-		if agg.metrics != nil {
-			agg.metrics.aggregatorErrors.WithLabelValues("MissingFieldError", agg.inputField).Inc()
+		// error only if explicitly specified as FLP skip empty fields by default to reduce storage size
+		if agg.reportMissing {
```
msherif1234, line 114: I assume netobserv app always expect this bool to be `false` ? when can this bool be `true` I am wondering if we even need to bool and we can omit the error always ?

pkg/pipeline/extract/conntrack/aggregator.go
```diff
@@ -179,11 +185,13 @@ func (agg *aMax) update(conn connection, flowLog config.GenericMap, d direction,
 }
 
 func (cp *aFirst) update(conn connection, flowLog config.GenericMap, d direction, isNew bool) {
-	if isNew {
+	if isNew && flowLog[cp.inputField] != nil {
```
msherif1234, line 188: I assume if we don't allow by default empty fields then we won't need this check as well as the one in the update function ?

pkg/pipeline/extract/conntrack/aggregator.go
jpinsonneau: I feel this could be usefull even for `netobserv`. 
`FlowDirection`, `IfDirection`, `TimeFlowStartMs`, `Bytes`, `Packets` fields for example.

pkg/pipeline/extract/conntrack/aggregator.go
jpinsonneau: The goal here is to avoid reporting empty fields for `first` and `last` operations (just below this one). 
Without it, these will create field with nil: 
https://github.com/netobserv/flowlogs-pipeline/blob/381890dd4cf93fadebc491e7beb5c4165a9ceff7/pkg/pipeline/extract/conntrack/conn.go#L66

I'm not sure to get your point. Can you please clarify ?

jpinsonneau: After testing it seems better to work directly in the `toGenericMap` function
https://github.com/netobserv/flowlogs-pipeline/pull/470/commits/74611d21c7b2a45686c4f754124f30c6fa05c19b

That avoid to manage per aggregate cases

jpinsonneau: @msherif1234 any reason not skipping zeros for Bytes and Packets fields ? 
https://github.com/netobserv/flowlogs-pipeline/pull/470/commits/de53e15aa87a4cc7a398a91e9c7c6e0adce0a213

For consistency it would be useful

msherif1234: no @jpinsonneau it makes sense to skip empty fields

pkg/pipeline/extract/conntrack/conn.go
```diff
@@ -100,7 +101,9 @@ func (c *connType) getNextHeartbeatTime() time.Time {
 func (c *connType) toGenericMap() config.GenericMap {
 	gm := config.GenericMap{}
 	for k, v := range c.aggFields {
-		gm[k] = v
+		if v != nil && (reflect.TypeOf(v).Kind() != reflect.Float64 || v.(float64) != 0) {
```
msherif1234, line 104: should this be && i.e
```golang
if v != nil && reflect.TypeOf(v).Kind() == reflect.Float64 && v.(float64) != 0 {
```

jpinsonneau: `aggFields` can be other types than `float64` here
Example: `string` for `PktDropLatestState` or `PktDropLatestDropCause` on `last` aggregate

msherif1234: /lgtm

jpinsonneau:
merging this to unlock [operator PR](https://github.com/netobserv/network-observability-operator/pull/401)

PR #471, opened 2023-08-10
NETOBSERV-1233: Remove L3 fields for ARP ethtype packets and update unit-test

msherif1234:
- remove l3 fields fom FLP storgae for layer2 EthTypes protocol like ARP
- extend unit-test to test various protocols and L2 protocols 
- update conntrack to skip l2 flows


pkg/pipeline/extract/conntrack/conntrack.go
```diff
@@ -61,6 +61,11 @@ func (ct *conntrackImpl) Extract(flowLogs []config.GenericMap) []config.GenericM
 
 	var outputRecords []config.GenericMap
 	for _, fl := range flowLogs {
+		if !fl.IsValidProtocol() {
+			log.Debugf("skipping layer2 protocols flow log %v", fl)
+			ct.metrics.inputRecords.WithLabelValues("discarded").Inc()
```
jpinsonneau, line 66:
```suggestion
			ct.metrics.inputRecords.WithLabelValues("rejected").Inc()
```

Any reason not using the same metric as below ? Else maybe we should make the difference clear for the user

jpinsonneau (dismissed):
Small comment, otherwise looks good ! thanks @msherif1234 !

pkg/pipeline/extract/conntrack/conntrack.go
msherif1234: I felt rejected because of any error but in this case there is no error just l2 protocol that we aren't supporting but I am up for better suggestion ?

jpinsonneau: What about [a note in the README](https://github.com/netobserv/flowlogs-pipeline/blob/ba6bf34ad04a4ded723579b7a1027d3727f3450c/README.md?plain=1#L650) explaining the diffs then ?

msherif1234: Hi @jpinsonneau I checked readme it doesn't really explain any of the metrics ? is there another place where this is already doc ?

jpinsonneau: nope @msherif1234 it's something we should add around the line I mentionned I guess

jpinsonneau (dismissed):
LGTM thanks @msherif1234

msherif1234:
I am still seeing a couple of instances of the following in flp logs
```sh
time=2023-08-11T12:51:47Z level=warning msg=Missing field SrcPort
time=2023-08-11T12:51:47Z level=warning msg=Missing field DstPort
time=2023-08-11T12:51:57Z level=warning msg=Missing field SrcPort
time=2023-08-11T12:51:57Z level=warning msg=Missing field DstPort
```

pkg/pipeline/extract/conntrack/conntrack.go
```diff
@@ -61,6 +61,16 @@ func (ct *conntrackImpl) Extract(flowLogs []config.GenericMap) []config.GenericM
...
+			continue
+		}
+		if fl.IsNoneTransportProtocol() {
+			log.Debugf("skipping none transport protocols (ICMPv4/6) flow log %v", fl)
```
ronensc, line 70: Perhaps we should state this condition the other way around?
```golang
if !IsTransportProtocol() {
```
and say that conntrack supports only TCP and UDP

pkg/pipeline/extract/conntrack/conntrack.go
```diff
@@ -61,6 +61,16 @@ func (ct *conntrackImpl) Extract(flowLogs []config.GenericMap) []config.GenericM
...
+			ct.metrics.inputRecords.WithLabelValues("discarded").Inc()
+			continue
+		}
```
ronensc, line 73: I think I would extract the invocation of `IsValidProtocol()` and `IsNoneTransportProtocol()` to a method like `filterFlowLog()` to make this loop look more general.

pkg/config/generic_map.go
```diff
@@ -42,3 +49,26 @@ func (m GenericMap) IsDuplicate() bool {
...
+	if proto, ok := m[protoFieldName]; ok {
+		if proto, err := utils.ConvertToFloat64(proto); err == nil {
+			if proto == float64(syscall.IPPROTO_IGMP) || proto == float64(syscall.IPPROTO_ICMPV6) {
+				return true
+			}
+		}
+		if proto, err := utils.ConvertToUint32(proto); err == nil {
+			if proto == syscall.IPPROTO_ICMP || proto == syscall.IPPROTO_ICMPV6 {
+				return true
+			}
```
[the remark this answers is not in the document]
msherif1234: typo should be ICMP for both sorry

pkg/config/generic_map.go
```diff
@@ -17,11 +17,18 @@
...
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/utils"
+)
 
 type GenericMap map[string]interface{}
 
-const duplicateFieldName = "Duplicate"
+const (
+	duplicateFieldName = "Duplicate"
+	protoFieldName     = "Proto"
```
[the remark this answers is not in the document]
msherif1234: no following this is the key of the generic CM ?

pkg/config/generic_map.go
```diff
@@ -42,3 +49,28 @@ func (m GenericMap) IsDuplicate() bool {
...
+		if proto, err := utils.ConvertToFloat64(v); err == nil {
+			if proto == float64(syscall.IPPROTO_TCP) || proto == float64(syscall.IPPROTO_UDP) || proto == float64(syscall.IPPROTO_SCTP) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func (m GenericMap) FilterFLowLog() bool {
```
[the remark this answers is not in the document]
ronensc: I think `FilterFlowLog()` should be a method `conntrackImpl` and get a flowLog as an argument. Sorry for not making it clear in my previous comment.

pkg/config/generic_map.go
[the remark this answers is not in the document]
ronensc: My point is:
What if we get flow logs from a source that gives the protocol field a name different from `Proto`, for example `Protocol`?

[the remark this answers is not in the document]
jpinsonneau: so we need also to make `Duplicate` configurable ?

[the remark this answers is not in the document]
ronensc: Yes. Actually, the same claim can be applied to `Duplicate` as well.
However, I'm fine with leaving `Duplicate` and `Proto` hardcoded and make them configurable when there is a real use case.

jpinsonneau (dismissed):

msherif1234:
@Amoghrd can u pls check out this fix and add the required tags if changes are ok ?

pkg/pipeline/extract/conntrack/conntrack.go
```diff
@@ -56,11 +56,22 @@ type conntrackImpl struct {
 	metrics                          *metricsType
 }
 
+func (ct *conntrackImpl) filterFLowLog(fl config.GenericMap) bool {
```
ronensc, line 59: nitpick: Could you please rename `filterFLowLog` to `filterFlowLog`?

ronensc (approved):

ronensc (approved):

PR #474, opened 2023-08-23
NETOBSERV-1247: using Meta informer for replicaSets

msherif1234:
This is follow up to the work @jpinsonneau started in https://github.com/jpinsonneau/flowlogs-pipeline/tree/update_metainformer

- switch to metadata informer for RS

### DIFF
--- a/pkg/pipeline/transform/kubernetes/kubernetes.go
+++ b/pkg/pipeline/transform/kubernetes/kubernetes.go
@@ -19,18 +19,22 @@ package kubernetes
 
 import (
 	"fmt"
+	appsv1 "k8s.io/api/apps/v1"
 	"net"
 	"os"
 	"path"
 	"time"
 
 	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/transform/kubernetes/cni"
+
 	log "github.com/sirupsen/logrus"
-	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/metadata"
+	"k8s.io/client-go/metadata/metadatainformer"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
@@ -60,6 +64,7 @@ type KubeData struct {
 	// replicaSets caches the ReplicaSets as partially-filled *ObjectMeta pointers
 	replicaSets cache.SharedIndexInformer
 	stopChan    chan struct{}
+	mdStopChan  chan struct{}
 }
 
 type Owner struct {
@@ -145,7 +150,7 @@ func (k *KubeData) getOwner(info *Info) Owner {
 			log.WithError(err).WithField("key", info.Namespace+"/"+ownerReference.Name).
 				Debug("can't get ReplicaSet info from informer. Ignoring")
 		} else if ok {
-			rsInfo := item.(*metav1.ObjectMeta)
+			rsInfo := item.(*metav1.PartialObjectMetadata)
 			if len(rsInfo.OwnerReferences) > 0 {
 				return Owner{
 					Name: rsInfo.OwnerReferences[0].Name,
@@ -267,7 +272,7 @@ func (k *KubeData) initServiceInformer(informerFactory informers.SharedInformerF
 		ips := make([]string, 0, len(svc.Spec.ClusterIPs))
 		for _, ip := range svc.Spec.ClusterIPs {
 			// ignoring None IPs
-			if ip != v1.ClusterIPNone {
+			if isServiceIPSet(ip) {
 				ips = append(ips, ip)
 			}
 		}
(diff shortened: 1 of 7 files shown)


msherif1234:
image

jpinsonneau:
> image

The initial issue was on a cluster having `15k+ replicasets`. We should try to reproduce something around this number :smile_cat:

jpinsonneau (dismissed):
Code looks good, not tested yet
Do you see differences in the enriched flows ?

msherif1234:
- using hey-ho too `/hey-ho.sh -n 50`, which generate the following RS load
```sh
$ oc get rs -A | wc -l
343
```
metadata_informer

if we ignore few flp pods with PR showing much higher usage then I would say this is mostly the same , I am not able to increase RS to lagre number w/o increasing current default of 250 pods/node which prevented me from swapping FLP images to test before and after

jpinsonneau:
> if we ignore few flp pods with PR showing much higher usage then I would say this is mostly the same , I am not able to increase RS to lagre number w/o increasing current default of 250 pods/node which prevented me from swapping FLP images to test before and after

So do you think it's not worth it or should it be verified by QEs ?

msherif1234:
> > if we ignore few flp pods with PR showing much higher usage then I would say this is mostly the same , I am not able to increase RS to lagre number w/o increasing current default of 250 pods/node which prevented me from swapping FLP images to test before and after
> 
> So do you think it's not worth it or should it be verified by QEs ?

I think QE verified on scale run we might have better numbers there @memodi WDYT ?

memodi:
I guess the idea here to rollout the new revisions to existing deployments, we don't have any tool that does that today. Simply, increasing the cluster size won't help since FLP pods would scale equally.

memodi:
@msherif1234 @jpinsonneau - does all FLP tries to build similar cache? I am trying to think if we had say 120 node cluster with 100 deployments on each worker would be appropriate scenario? cc @nathan-weinberg

jpinsonneau:
> @msherif1234 @jpinsonneau - does all FLP tries to build similar cache? 

Yes every FLP build the exact same cache. That's why the kube API is taking a burst on Netobserv startup
https://issues.redhat.com/browse/NETOBSERV-1248

> I am trying to think if we had say 120 node cluster with 100 deployments on each worker would be appropriate scenario? cc @nathan-weinberg

So you would be around `12k` pods ? More if you plan multiple replicas ? 
That sounds good ! Compare FLPs overall CPU / Memory without that PR and with that PR.
A plus would also be checking kube api perfs.

nathan-weinberg:
I ran a 120 node cluster-density-v2 test on two separate but identical clusters - the "control" cluster performed as expected but the cluster with the patched image saw large amounts of flows being dropped due to the following FLP error:
```bash
time=2023-09-05T16:35:37Z level=info component=client error=server returned HTTP status 429 Too Many Requests (429): Maximum active stream limit exceeded, reduce the number of active streams (reduce labels or reduce label values), or contact your Loki administrator to see if the limit can be increased fields.level=error fields.msg=final error sending batch host=lokistack-gateway-http.netobserv.svc.cluster.local:8080 module=export/loki status=429
```

You can see this in the Dittybopper dashboard screenshot pasted below (the first test on the left and the second test on the right): 
image

The LokiStack we are using is a `1x.medium` size with the following limits set:
```yaml
spec:
  limits:
    global:
      ingestion:
        ingestionBurstSize: 100
        ingestionRate: 500
        maxGlobalStreamsPerTenant: 100000
        perStreamRateLimit: 6
        perStreamRateLimitBurst: 30
```
This is the same configuration that was used in the control cluster as well

jpinsonneau:
@msherif1234 where are we on that one ? Did you reproduced the drops observed by @nathan-weinberg ?

msherif1234:
> @msherif1234 where are we on that one ? Did you reproduced the drops observed by @nathan-weinberg ?

I am not able to repro the same error using

pkg/pipeline/transform/kubernetes/kubernetes.go
```diff
@@ -291,23 +294,13 @@ func (k *KubeData) initServiceInformer(informerFactory informers.SharedInformerF
...
-	}); err != nil {
-		return fmt.Errorf("can't set ReplicaSets transform: %w", err)
-	}
+func (k *KubeData) initReplicaSetInformer(informerFactory metadatainformer.SharedInformerFactory) error {
+	k.replicaSets = informerFactory.ForResource(
+		schema.GroupVersionResource{
+			Group:    "extensions",
+			Version:  "v1beta1",
+			Resource: "deployments",
+		}).Informer()
```
jotak, line 303: I don't get this: we need to get a Replicaset informer, not a Deployment?

jotak:
Unlike what the JIRA description says I don't think we can expect any improvement in terms of memory usage here ?   Because in the end we don't change what we're storing in cache (our cache is already filtered to get only the needed information), but it should improve the kube-api burst traffic a little I guess ? Just "a little" because it is only affecting replicasets while the bulk of the traffic is likely coming from pods informers ? Unfortunately, I wish we could tweak informers to get from kube-api only the few info that we need (IPs from status + a few other things), but I haven't seen this in informer options ... :-(

pkg/pipeline/transform/kubernetes/kubernetes.go
msherif1234: changed to rs

pkg/pipeline/transform/kubernetes/kubernetes.go
```diff
@@ -360,8 +358,9 @@ func LoadConfig(kubeConfigPath string) (*rest.Config, error) {
 	return config, nil
 }
 
-func (k *KubeData) initInformers(client kubernetes.Interface) error {
+func (k *KubeData) initInformers(client kubernetes.Interface, metaClient metadata.Interface) error {
 	informerFactory := informers.NewSharedInformerFactory(client, syncTime)
+	metadataInformerFactory := metadatainformer.NewSharedInformerFactory(metaClient, syncTime)
```
jotak, line 363: I guess you also need to `Start` and `WaitForCacheSync` this new informer below (around line 382)

pkg/pipeline/transform/kubernetes/kubernetes.go
```diff
@@ -291,23 +294,13 @@ func (k *KubeData) initServiceInformer(informerFactory informers.SharedInformerF
 	return nil
 }
 
-func (k *KubeData) initReplicaSetInformer(informerFactory informers.SharedInformerFactory) error {
-	k.replicaSets = informerFactory.Apps().V1().ReplicaSets().Informer()
-	// To save space, instead of storing a complete *appvs1.Replicaset instance, the
-	// informer's cache will store a *metav1.ObjectMeta with the minimal required fields
-	if err := k.replicaSets.SetTransform(func(i interface{}) (interface{}, error) {
```
jotak, line 298: I think the SetTransform should be kept, because it reduces memory usage even more, by keeping only Name/Namespace/OwnerRefs instead of the full metadata

jotak:
/lgtm
For test failing: this should fix it: https://github.com/netobserv/flowlogs-pipeline/pull/534

jotak (approved):

Issue #473, opened 2023-08-23
Failing make ocp-deploy


aslom:
I was testing deploying the latest versionto OpenShift with @eranra and ran into two problem

I am using the latest version:

```
git rev-parse HEAD
914d470447c1ff8d9bbf3bb371a74039d08bbf17
```

problems:

1) it is not obvious that $USER is used during deployment as default org and need to be changed by IMAGE_ORG=netobserv

2) even after changing IMAGE_ORG the depoyments still fails

```
IMAGE_ORG=netobserv make ocp-deploy
sed 's|%IMAGE_TAG_BASE%|quay.io/netobserv/flowlogs-pipeline|g;s|%VERSION%|latest|g' contrib/kubernetes/deployment.yaml > /tmp/deployment.yaml
kubectl --ignore-not-found=true  delete configmap flowlogs-pipeline-configuration || true
configmap "flowlogs-pipeline-configuration" deleted
kubectl --ignore-not-found=true delete -f /tmp/deployment.yaml || true
deployment.apps "flowlogs-pipeline" deleted
service "flowlogs-pipeline" deleted
service "flowlogs-pipeline-metrics" deleted
serviceaccount "flowlogs-pipeline" deleted
clusterrole.rbac.authorization.k8s.io "flowlogs-pipeline" deleted
clusterrolebinding.rbac.authorization.k8s.io "flowlogs-pipeline" deleted
kubectl --ignore-not-found=true delete -f contrib/kubernetes/deployment-loki.yaml || true
configmap "loki-config" deleted
deployment.apps "loki" deleted
service "loki" deleted
kubectl --ignore-not-found=true delete -f contrib/kubernetes/deployment-loki-storage.yaml || true
persistentvolumeclaim "loki-store" deleted
pkill --oldest --full "3100:3100"
pkill: illegal option -- -
usage: pkill [-signal] [-ILfilnovx] [-F pidfile] [-G gid]
             [-P ppid] [-U uid] [-g pgrp] [-t tty] [-u euid]
             pattern ...
make: [undeploy-loki] Error 2 (ignored)
kubectl --ignore-not-found=true delete -f contrib/kubernetes/deployment-prometheus.yaml || true
deployment.apps "prometheus" deleted
service "prometheus" deleted
configmap "prometheus-server-conf" deleted
pkill --oldest --full "9090:9090"
pkill: illegal option -- -
usage: pkill [-signal] [-ILfilnovx] [-F pidfile] [-G gid]
[...]

PR #475, opened 2023-08-23
Update README and config

ronensc:
Following the feedback of @aslom in #473:

1. The README was updated to run `make ocp-deploy` with `IMAGE_ORG=netobserv`
2. The configuration of FLP was updated. #418 introduced a new API field in `extract_aggregate` which required this update.
3. On the way, updated the operational metric prefix to `flp_op_` (added underscore) to avoid metric names like `flp_opmetrics_processed`

### DIFF
--- a/contrib/kubernetes/flowlogs-pipeline.conf.yaml
+++ b/contrib/kubernetes/flowlogs-pipeline.conf.yaml
@@ -106,8 +106,8 @@ parameters:
         input: timeReceived
       scheduling:
       - endConnectionTimeout: 10s
-        heartbeatInterval: 30s
         terminatingTimeout: 5s
+        heartbeatInterval: 30s
       tcpFlags:
         fieldName: TCPFlags
         detectEndConnection: true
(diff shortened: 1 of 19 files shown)


contrib/kubernetes/flowlogs-pipeline.conf.yaml
```diff
@@ -437,6 +454,7 @@ parameters:
         filter:
           key: name
           value: dest_service_count
+        filters: []
```
jpinsonneau, line 457: `filter` is deprecated in favor of `filters` so we should convert these to arrays

contrib/kubernetes/flowlogs-pipeline.conf.yaml
```diff
@@ -255,7 +256,10 @@ parameters:
       - name: bandwidth_per_network_service
         type: counter
         filter:
-          key: name
+          key: ""
+          value: ""
```
KalmanMeth, line 260: Since `filter` is being deprecated, shouldn't we just delete this section?

ronensc: done

KalmanMeth (approved):

PR #478, opened 2023-08-29
NETOBSERV-1352 enhance prom filters for RTT metrics

jpinsonneau:
This PR improve prom metrics filtering introducing:
- [x] `!nil` to match presenceof the field with any value
- [X] `nil` to match absence of the field
- [x] `|` splitter to match at least one value

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -13,10 +13,10 @@ Following is the supported API format for prometheus encode:
                      agg_histogram: counts samples in configurable buckets, pre-aggregated via an Aggregate stage
                  filter: an optional criterion to filter entries by. Deprecated: use filters instead.
                      key: the key to match and filter by
-                     value: the value to match and filter by. Use !nil / nil to match presence / absence
+                     value: the value to match and filter by. Use !nil / nil to match presence / absence. Add multiple matching values using '|' rune such as 'a|b' to match either 'a' or 'b'.
                  filters: a list of criteria to filter entries by
                          key: the key to match and filter by
-                         value: the value to match and filter by. Use !nil / nil to match presence / absence
+                         value: the value to match and filter by. Use !nil / nil to match presence / absence. Add multiple matching values using '|' rune such as 'a|b' to match either 'a' or 'b'.
                  valueKey: entry key from which to resolve metric value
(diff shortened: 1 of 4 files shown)


jotak (approved):
LGTM

jotak:
@jpinsonneau I flagged LGTM, but taken in the light of bug https://issues.redhat.com/browse/NETOBSERV-1344, we will need to have filters such as "FlowDirection==1 OR FlowDirection==2" and/or "FlowDirection!=0" and/or "FlowDirection~=1|2" ... we should probably think about that before merging

jpinsonneau:
> @jpinsonneau I flagged LGTM, but taken in the light of bug https://issues.redhat.com/browse/NETOBSERV-1344, we will need to have filters such as "FlowDirection==1 OR FlowDirection==2" and/or "FlowDirection!=0" and/or "FlowDirection~=1|2" ... we should probably think about that before merging

Let's just merge all metrics PRs pending before doing the fix everywhere so we don't miss any ?

jotak:
I'm ok to go with it but just to mention a small concern, it creates some reserved words or characters that are not possible to have in filters as raw text .. even if today it sounds unlikely to have "nil" or "|" actually written in a field and to be willing to filter on that (as a text value), still we are creating reserved keyword that we could avoid by just changing the API.

E.g: as an alternative, we could implement a `type` param on filters, as an enum: `exact` (= current behaviour), `presence`, `absence`, `regex`.
wdyt? we could do that as a follow-up

jpinsonneau:
> I'm ok to go with it but just to mention a small concern, it creates some reserved words or characters that are not possible to have in filters as raw text .. even if today it sounds unlikely to have "nil" or "|" actually written in a field and to be willing to filter on that (as a text value), still we are creating reserved keyword that we could avoid by just changing the API.
> 
> E.g: as an alternative, we could implement a `type` param on filters, as an enum: `exact` (= current behaviour), `presence`, `absence`, `regex`. wdyt? we could do that as a follow-up

As you prefer. I don't have strong opinion on that. Feel free to merge if it's blocking other PRs else I can refactor here; just let me know.

jotak:
/approve
(this is blocking other PRs)

PR #483, opened 2023-09-01
NETOBSERV-1275: Introduce new "INNER" direction for inner-node traffic

jotak:
Related PR: https://github.com/netobserv/network-observability-console-plugin/pull/378

The flows (and duplicates) generated for inner-node traffic differs compared to node-to-node traffic, and reinterpret direction isn't able to decide between ingress or egress. This is causing discrepancies with the dedup mechanism that filters out flows where Duplicate=true and also favors ingress over egress, since the Duplicate flag can be set randomly on ingress or on egress.

To fix that, the proposed solution is to create this new INNER direction specifically for this kind of traffic. Deduping this INNER traffic can then rely solely on the Duplicate flag, since that flag was set from a single Agent (single node) there will always be only one Duplicate=false.

### DIFF
--- a/pkg/pipeline/transform/transform_network.go
+++ b/pkg/pipeline/transform/transform_network.go
@@ -124,10 +124,7 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
 				}
 			}
 		case api.OpReinterpretDirection:
-			// only reinterpret direction on flowlogs
-			if rt, ok := outputEntry["_RecordType"]; !ok || rt == "flowLog" {
-				reinterpretDirection(outputEntry, &n.DirectionInfo)
-			}
+			reinterpretDirection(outputEntry, &n.DirectionInfo)
 		case api.OpAddIPCategory:
 			if strIP, ok := outputEntry[rule.Input].(string); ok {
 				cat, ok := n.ipCatCache.GetCacheEntry(strIP)


jotak:
TODO: doc update

jpinsonneau (dismissed):
Looks good, let's just wait for plugin PR before merge.

Edit: what about managing [NETOBSERV-1109](https://issues.redhat.com/browse/NETOBSERV-1109) in this PR at the same time ?
We would just need to revert https://github.com/netobserv/flowlogs-pipeline/commit/a05a86d93c7ba900099d862b53979dc0516e66a4

jpinsonneau:
Another case where reinterpret doesn't work as expected is when `service` is involved.
In that situation the related `node` is not set and reinterpret is done incorrectly.

```
{
  "AgentIP": "10.0.1.79",
  "Bytes": 151,
  "DstAddr": "10.129.0.33",
  "DstK8S_HostIP": "10.0.1.79",
  "DstK8S_HostName": "ip-10-0-1-79.ec2.internal",
  "DstK8S_Name": "flowlogs-pipeline-tfbtz",
  "DstK8S_Namespace": "netobserv",
  "DstK8S_OwnerName": "flowlogs-pipeline",
  "DstK8S_OwnerType": "DaemonSet",
  "DstK8S_Type": "Pod",
  "DstMac": "0A:58:0A:81:00:21",
  "DstPort": 52682,
  "Duplicate": false,
  "Etype": 2048,
  "FlowDirection": "0",
  "IfDirection": 1,
  "Interface": "947d95f7ea5991c",
  "K8S_ClusterName": "8c12585d-2e05-47b7-b41d-a44f7e4b93f0",
  "Packets": 1,
  "Proto": 17,
  "SrcAddr": "172.30.0.10",
  "SrcK8S_Name": "dns-default",
  "SrcK8S_Namespace": "openshift-dns",
  "SrcK8S_OwnerName": "dns-default",
  "SrcK8S_OwnerType": "Service",
  "SrcK8S_Type": "Service",
  "SrcMac": "0A:58:0A:81:00:06",
  "SrcPort": 53,
  "TimeFlowEndMs": 1693833862664,
  "TimeFlowStartMs": 1693833862664,
  "TimeReceived": 1693833866,
  "_HashId": "b3de5075fb27159f",
  "_RecordType": "flowLog",
  "app": "netobserv-flowcollector"
}
```

jotak:
> Another case where reinterpret doesn't work as expected is when `service` is involved. In that situation the related `node` is not set and reinterpret is done incorrectly.

In this example, AgentIP == DstK8S_HostIP so reinterpret says it's an Ingress (seems like this is a DNS response). Why is it incorrect? It seems an ingress indeed, despite not knowing the other end's host IP ?

jpinsonneau:
> In this example, AgentIP == DstK8S_HostIP so reinterpret says it's an Ingress (seems like this is a DNS response observed from the sender). Why is it incorrect? It seems an ingress indeed, despite not knowing the other end's host IP ?

DNS `requests` and `responses` are using the same flow, one egress and one ingress.
https://github.com/netobserv/netobserv-ebpf-agent/blob/main/bpf/dns_tracker.h#L25-L28
https://issues.redhat.com/browse/NETOBSERV-1240 

Would you expect `Src` and `Dst` to be flipped for the response ?

> @jpinsonneau for conversation I can remove that `if`, you think it's good enough? (no change to do on query side?)

Yes please :+1:  No changes on query side as far as I seen.

jpinsonneau:
> > Would you expect `Src` and `Dst` to be flipped for the response ?
> 
> Well this is how it has always worked with netflows, so yes, for the response, Src must be the DNS server, and Dst must be the querier
> 
> Edit: but it seems to be already the case, no?

Yes my bad, I retested and see 1 egress (Pod -> Service) and 2 ingress (Service -> Pod) with one duplicate.
But why don't we consider these as `Inner` since services are running on every nodes ?

jotak:
Services aren't really running on nodes, I'm not sure we could consider them as "Inner traffic" ... 
I think the root cause here for DNS is different, it's that the DNS tracker sets its data on a flow that potentially has been flagged as duplicate. Ideally, the DNS tracker should not do that ; it should pick the non-duplicate flow to enrich it , but I'm not sure we can do that in the bpf program .. cc @msherif1234 do you know?

jotak:
Regarding my last comment: DNS duplicates are currently being considered via [this ticket](https://issues.redhat.com/browse/NETOBSERV-1240) and dealt with from the Agent

jotak:
/approve
will be tested by QE post-merge

PR #513, opened 2023-10-09
NETOBSERV-1231: Improve promencode API: scaling & filters

jotak:
- New ValueScale config to allow scale conversion, e.g. RTT ns to seconds
- To avoid having reserved characters or words in filters, use a dedicated API instead: new setting Filter.Type (exact,presence,absence or regex)
- Pre-compute regex by setting up a "predicate" function per filter

**Breaking-change** : metrics currently defining filters such as "FlowDirection: 0|2" must be adapted with the new Filter Type API

## Dependencies

n/a

## Checklist

If you are not familiar with our processes or don't know what to answer in the list below, let us know in a comment: the maintainers will take care of that.

* [ ] Will this change affect NetObserv / Network Observability operator? If not, you can ignore the rest of this checklist.
* [ ] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._
* [ ] Does this PR require product documentation?
  * [ ] If so, make sure the JIRA epic is labelled with "documentation" and provides a description relevant for doc writers, such as use cases or scenarios. Any required step to activate or configure the feature should be documented there, such as new CRD knobs.
* [ ] Does this PR require a product release notes entry?
  * [ ] If so, fill in "Release Note Text" in the JIRA.
* [ ] Is there anything else the QE team should know before testing? E.g: configuration changes, environment setup, etc.
  * [ ] If so, make sure it is described in the JIRA ticket.
* QE requirements (check 1 from the list):
  * [ ] Standard QE validation, with pre-merge tests unless stated otherwise.
  * [ ] Regression tests only (e.g. refactoring with no user-facing change).
  * [x] No QE (e.g. trivial change with high reviewer's confidence, or per agreement with the QE team).

PR #531, opened 2023-10-24
Opentelemetry exporter

KalmanMeth:
## Description

Export logs, traces, and metrics to an Opentelemetry collector.
Configuration enables to specify using either grpc or http protocol.
TLS details may be specified. If not specified, then the tls option is set to insecure.
Sample configuration:
```

  - name: otlp_collector
    encode:
      type: otlptraces
      otlptraces:
        address: localhost
        port: 4317
        connectionType: grpc
  - name: otlp_instana
    encode:
      type: otlptrace
      otlptrace:
        httpaddress: "1.2.3.4:443"
        tls:
          insecureSkipVerify: true
        headers:
          x-instana-key: *****
          x-instana-host: host
          x-instana-time: "0"

```

Headers to be sent to collector may be specified. 
Field names should be transformed to Opentelemetry-conformant format using a transform_generic stage.
```

  - name: otlp_transform
    transform:
      type: generic
      generic:
        policy: replace_keys
        rules:
          - input: Bytes
            output: bytes
            multiplier: 1
          - input: Packets
            output: packets
          - input: DstAddr
            output: destination.address
          - input: DstMac
            output: destination.mac
          - input: DstHostIP
            output: destination.host.address
          - input: DstK8S_HostName
            output: destination.k8s.node.name
          - input: DstPort
            output: destination.port
          - input: DstPod
            output: destination.k8s.pod.name
          - input: DstK8S_Name
            output: destination.k8s.name
          - input: DstK8S_Type
            output: destination.k8s.type
          - input: DstK8S_OwnerName
            output: destination.k8s.owner.name
          - input: DstK8S_Type
            output: destination.k8s.type
          - input: DstK8S_HostIP
            output: destination.k8s.host.address
          - input: DstK8S_HostName
            output: destination.k8s.host.name
          - input: DstK8S_Namespace
            output: destination.k8s.namespace.name
          - input: SrcAddr
            output: source.address
          - input: SrcMac
            output: source.mac
          - input: SrcHostIP
            output: source.host.address
          - input: SrcK8S_HostName
            output: source.k8s.node.name
          - input: SrcPort
            output: source.port
          - input: SrcPod
            output: source.k8s.pod.name
          - input: SrcK8S_Name
            output: source.k8s.name
          - input: SrcK8S_Type
            output: source.k8s.type
          - input: SrcK8S_OwnerName
            output: source.k8s.owner.name
          - input: SrcK8S_OwnerType
            output: source.k8s.owner.type
          - input: SrcK8S_Namespace
            output: source.k8s.namespace.name
          - input: SrcK8S_HostIP
            output: source.k8s.host.address
          - input: SrcK8S_HostName
            output: source.k8s.host.name
          - input: TimeReceived
            output: timereceived
          - input: Proto
            output: protocol
          - input: FlowDirection
            output: flowdirection

```

## Dependencies

n/a

## Checklist


jotak:
Thanks @KalmanMeth !
On semantic convention, I wonder if we should have something more "baked-in" rather than asking the user to configure the translations. Especially, some stuff might be hard to do only by config, e.g: our `SrcK8S_Name` should actually be translated into `source.k8s.pod.name` or `source.k8s.service.name` or `source.k8s.node.name` depending on what is `SrcK8S_Type`.

So perhaps, since the knowledge currently lies in the `add_kubernetes` transform, we could have a flag `otel` on/off in this block that, if turned on, would automatically use the otel semantics instead of our current names ?

KalmanMeth:
> Thanks @KalmanMeth ! On semantic convention, I wonder if we should have something more "baked-in" rather than asking the user to configure the translations. Especially, some stuff might be hard to do only by config, e.g: our `SrcK8S_Name` should actually be translated into `source.k8s.pod.name` or `source.k8s.service.name` or `source.k8s.node.name` depending on what is `SrcK8S_Type`.
> 
> So perhaps, since the knowledge currently lies in the `add_kubernetes` transform, we could have a flag `otel` on/off in this block that, if turned on, would automatically use the otel semantics instead of our current names ?

I added the otel option to add_kubernetes by setting the `assignee` field to `otel`. I defined the fields like `k8s.node.name`, `k8s.pod.name`, etc. We should have a discussion on which other fields we think we should provide in this manner as opposed to using the transform-generic capabilities.

KalmanMeth:
I have working counters and gauges for the opentelemetry metrics. I removed the code for opentelemetry historgrams. This may have to wait for a future PR.

pkg/pipeline/encode/opentelemetry/encode_otlptrace.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Encode encodes a metric to be exported
+func (e *EncodeOtlpTrace) Encode(metricRecord config.GenericMap) {
+	log.Tracef("entering EncodeOtlpTrace. entry = %v", metricRecord)
+	_, span := e.tracer.Start(e.ctx, flpEncodeSpanName)
+	defer span.End()
+	attributes := obtainAttributesFromEntry(metricRecord)
+	span.SetAttributes(*attributes...)
+}
```
jotak, line 54: I am not sure what is the goal here with traces, but I don't think we're doing something useful here.
Normally, traces are provided by the workloads that are being traced, for instance when a workload A sends a request to service S, A would generate a span saying "I am sending a request to S".

I believe what we could possibly do, as an external observer, is to generate spans on behalf of the workloads & services as long as we see their IPs flowing. I'm not 100% sure how useful it is, but that's the best I can see. But it's not really what this is doing here: it creates spans for FLP without providing any real information about the network connections being observed in otel format.

pkg/pipeline/encode/opentelemetry/encode_otlpmetrics.go
```diff
@@ -0,0 +1,300 @@
...
+		switch filter.Value {
+		case "nil":
+			if found {
+				return nil
+			}
+		case "!nil":
+			if !found {
+				return nil
+			}
+		default:
```
jotak, line 146: The code that came from encode_prom is obsolete, see https://github.com/netobserv/flowlogs-pipeline/pull/513/files#diff-30d11303a4ad1ec82e4f39e3ae6dc7d044c8dd370c536a98ba90febbfe2d425aR209

jotak: basically: these special keywords "nil" / "!nil" have been replaced with a more explicit API where you can have more advanced filters, e.g. using regexp

jotak:
Focusing on code reviewing strictly speaking rather than if it meets expectations because I must say I am not quite sure about the use cases but you @KalmanMeth have apparently one using instana so you're you own judge on that :-) I struggle a bit to see where netobserv can bring value on tracing (unless we captured http headers with our agents, which we don't)

So apart from [this comment](https://github.com/netobserv/flowlogs-pipeline/pull/531#discussion_r1400738683) this looks good to me.

I guess there can be follow-ups to improve more the semantic conventions compliance.

KalmanMeth:
@jotak The trace capability now breaks down a single log entry into sub-entries - one for source and one for destination. I added instructions to test the setup using the opentelemetry collector and jaeger tracing UI that is fed with information from flp. I would like to progress with this PR to merge.

jotak (approved):
lgtm
thanks @KalmanMeth !

OlivierCazade:
Adding breaking change label due to change in name type in the API

PR #554, opened 2023-12-12
NETOBSERV-1244: Add Kubernetes Infra transform rule

OlivierCazade:
## Description

Add Kubernetes Infra transform rule

## Dependencies

Related operator PR:
https://github.com/netobserv/network-observability-operator/pull/517

## Checklist

* [X] Will this change affect NetObserv / Network Observability operator? If not, you can ignore the rest of this checklist.
* [X] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._
* [X] Does this PR require a product release notes entry?
  * [X] If so, fill in "Release Note Text" in the JIRA.

### DIFF
--- a/pkg/pipeline/transform/transform_network.go
+++ b/pkg/pipeline/transform/transform_network.go
@@ -99,10 +99,8 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
 			outputEntry[rule.Output] = serviceName
 		case api.OpAddKubernetes:
 			fillInK8s(outputEntry, rule)
-
 		case api.OpAddKubernetesInfra:
 			fillInK8sInfra(outputEntry, rule)
-
 		case api.OpReinterpretDirection:
 			reinterpretDirection(outputEntry, &n.DirectionInfo)
 		case api.OpAddIPCategory:
--- a/pkg/pipeline/transform/transform_network.go
+++ b/pkg/pipeline/transform/transform_network.go
@@ -99,6 +99,10 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
 			outputEntry[rule.Output] = serviceName
 		case api.OpAddKubernetes:
 			fillInK8s(outputEntry, rule)
+
+		case api.OpAddKubernetesInfra:
+			fillInK8sInfra(outputEntry, rule)
+
 		case api.OpReinterpretDirection:
 			reinterpretDirection(outputEntry, &n.DirectionInfo)
 		case api.OpAddIPCategory:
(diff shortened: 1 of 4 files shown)


pkg/pipeline/transform/transform_network.go
```diff
@@ -143,6 +161,28 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
...
+	if nsLen >= additionalPrefixLen && obj.Namespace[:additionalPrefixLen] == additionalInfraPrefix {
+		return false
+	}
+	//Special case with openshift and kubernetes service in default namespace
+	if obj.Namespace == "default" && (obj.Name == "kubernetes" || obj.Name == "openshift") {
+		return false
+	}
+	return true
+}
+
```
jpinsonneau, line 185:
```suggestion
```

pkg/pipeline/transform/transform_network.go
```diff
@@ -123,6 +123,24 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
...
+				logrus.WithError(err).Tracef("can't find kubernetes info for IP %v", outputEntry["SrcAddr"])
+				continue
+			}
+			dstKubeInfo, err := kubernetes.Data.GetInfo(fmt.Sprintf("%s", outputEntry["DstAddr"]))
+			if err != nil {
+				logrus.WithError(err).Tracef("can't find kubernetes info for IP %v", outputEntry["DstAddr"])
+				continue
+			}
+
+			if objectIsApp(*srcKubeInfo, rule.Parameters) || objectIsApp(*dstKubeInfo, rule.Parameters) {
```
jpinsonneau, line 138: We should move `GetInfo` calls into `objectIsApp` function. 

Let's say `srcKubeInfo` is app and `dstKubeInfo` is not available; we will miss `K8S_FlowType` whereas it could be identified as app since we know the source.

Also, when both `src` and `dst` are not identified, should we consider these as `app` ?

pkg/pipeline/transform/transform_network.go
```diff
@@ -123,6 +123,24 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
...
 			}
+		case api.OpAddKubernetesInfra:
+			srcKubeInfo, err := kubernetes.Data.GetInfo(fmt.Sprintf("%s", outputEntry["SrcAddr"]))
+			if err != nil {
+				logrus.WithError(err).Tracef("can't find kubernetes info for IP %v", outputEntry["SrcAddr"])
+				continue
+			}
+			dstKubeInfo, err := kubernetes.Data.GetInfo(fmt.Sprintf("%s", outputEntry["DstAddr"]))
+			if err != nil {
+				logrus.WithError(err).Tracef("can't find kubernetes info for IP %v", outputEntry["DstAddr"])
```
KalmanMeth, line 134: Is it OK that `SrcAddr` and `DstAddr` are hard-coded? Should these field names be configurable?

pkg/pipeline/transform/transform_network.go
OlivierCazade: Moved the GetInfo in obectIsApp thanks.

For now, every flow that is not tagged as app is tagged as infra.

pkg/pipeline/transform/transform_network.go
OlivierCazade: I added dedicated structure to configure rules.
I did not change other rules because I don't want to introduce API breaking change in this PR but I will create a following PR with the changes in the other rules.

pkg/pipeline/transform/transform_network.go
```diff
@@ -99,6 +99,10 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
 			outputEntry[rule.Output] = serviceName
 		case api.OpAddKubernetes:
 			fillInK8s(outputEntry, rule)
+
+		case api.OpAddKubernetesInfra:
+			fillInK8sInfra(outputEntry, rule)
+
 		case api.OpReinterpretDirection:
```
jpinsonneau, line 106:
```suggestion
		case api.OpAddKubernetesInfra:
			fillInK8sInfra(outputEntry, rule)
		case api.OpReinterpretDirection:
```

pkg/pipeline/transform/transform_network.go
```diff
@@ -182,6 +186,47 @@ func fillInK8s(outputEntry config.GenericMap, rule api.NetworkTransformRule) {
 	}
 }
 
+func fillInK8sInfra(outputEntry config.GenericMap, rule api.NetworkTransformRule) {
+	if rule.KubernetesInfra == nil {
+		logrus.Error("transformation rule: Missing Kubernetes Infra configuration ")
+		return
+	}
```
jpinsonneau, line 193: Should we provide a default behavior here ?

jpinsonneau (dismissed):
Looks good ! just small nit

pkg/pipeline/transform/transform_network.go
OlivierCazade: The case here happens if a user added a K8sinfra rule but did not configure it.

In this case, for now we ignore the rule and log an error.

PR #575, opened 2024-01-24
NETOBSERV-1203: Added option to add zones in k8s transform rule

OlivierCazade:
## Description

Zones are added to the flows if the rule is configured with the new addZone bool.

## Dependencies

n/a

## Checklist

* [x] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._

### DIFF
--- a/pkg/pipeline/transform/transform_network.go
+++ b/pkg/pipeline/transform/transform_network.go
@@ -148,6 +148,7 @@ func fillInK8s(outputEntry config.GenericMap, rule api.NetworkTransformRule) {
 				outputEntry[rule.Output+"_HostName"] = kubeInfo.HostName
 			}
 		}
+		fillInK8sZone(outputEntry, rule, *kubeInfo)
 	} else {
 		// NOTE: Some of these fields are taken from opentelemetry specs.
 		// See https://opentelemetry.io/docs/specs/semconv/resource/k8s/
@@ -181,6 +182,40 @@ func fillInK8s(outputEntry config.GenericMap, rule api.NetworkTransformRule) {
 				outputEntry[rule.Output+"k8s.host.name"] = kubeInfo.HostName
 			}
 		}
+	}
+}
+
+const nodeZoneLabelName = "topology.kubernetes.io/zone"
+
+func fillInK8sZone(outputEntry config.GenericMap, rule api.NetworkTransformRule, kubeInfo kubernetes.Info) {
+	if rule.Kubernetes == nil || !rule.Kubernetes.AddZone {
+		//Nothing to do
+		return
+	}
+	switch kubeInfo.Type {
+	case kubernetes.TypeNode:
+		zone, ok := kubeInfo.Labels[nodeZoneLabelName]
+		if ok {
+			outputEntry[rule.Output+"_Zone"] = zone
+		}
+		return
+	case kubernetes.TypePod:
+		nodeInfo, err := kubernetes.Data.GetNodeInfo(kubeInfo.HostName)
+		if err != nil {
+			logrus.WithError(err).Tracef("can't find nodes info for node %v", kubeInfo.HostName)
+			return
+		}
+		if nodeInfo != nil {
+			zone, ok := nodeInfo.Labels[nodeZoneLabelName]
+			if ok {
+				outputEntry[rule.Output+"_Zone"] = zone
+			}
+		}
+		return
+
+	case kubernetes.TypeService:
+		//A service is not assigned to a dedicated zone, skipping
+		return
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/pipeline/transform/transform_network.go
```diff
@@ -148,6 +148,7 @@ func fillInK8s(outputEntry config.GenericMap, rule api.NetworkTransformRule) {
 				outputEntry[rule.Output+"_HostName"] = kubeInfo.HostName
 			}
 		}
+		fillInK8sZone(outputEntry, rule, *kubeInfo)
```
jpinsonneau, line 151: We should also manage the open telemetry case in the else statement below this line.

- In case of a node we can add `k8s.node.zone`
- Other cases `k8s.host.zone`

Maybe filling the zones directly inside `fetchInformers` function in `kubernetes.go` file would actually be better ?
You can pass the rule to it and query the info accordingly.

Then in `transform_network.go` you conditionally fill the output as same as what's done for `HostIP` and `HostName`

WDYT ?

pkg/pipeline/transform/transform_network.go
```diff
@@ -184,6 +185,40 @@ func fillInK8s(outputEntry config.GenericMap, rule api.NetworkTransformRule) {
 	}
 }
 
+const nodeZoneLabelName = "topology.kubernetes.io/zone"
```
jpinsonneau, line 188: Would this be configurable ? I see `topology.kubernetes.io/region` & `topology.kubernetes.io/zone` and I wonder if some clusters have only region for example :thinking: 

https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion

pkg/pipeline/transform/transform_network.go
```diff
@@ -184,6 +185,40 @@ func fillInK8s(outputEntry config.GenericMap, rule api.NetworkTransformRule) {
...
+func fillInK8sZone(outputEntry config.GenericMap, rule api.NetworkTransformRule, kubeInfo kubernetes.Info) {
+	if rule.Kubernetes == nil || !rule.Kubernetes.AddZone {
+		//Nothing to do
+		return
+	}
+	switch kubeInfo.Type {
+	case kubernetes.TypeNode:
+		zone, ok := kubeInfo.Labels[nodeZoneLabelName]
+		if ok {
+			outputEntry[rule.Output+"_Zone"] = zone
```
jpinsonneau, line 199: Since this is node related, I would suggest `HostZone` for consistency with `HostIP` & `HostName`

```suggestion
			outputEntry[rule.Output+"_HostZone"] = zone
```

pkg/pipeline/transform/transform_network.go
```diff
@@ -184,6 +185,40 @@ func fillInK8s(outputEntry config.GenericMap, rule api.NetworkTransformRule) {
...
+	case kubernetes.TypePod:
+		nodeInfo, err := kubernetes.Data.GetNodeInfo(kubeInfo.HostName)
+		if err != nil {
+			logrus.WithError(err).Tracef("can't find nodes info for node %v", kubeInfo.HostName)
+			return
+		}
+		if nodeInfo != nil {
+			zone, ok := nodeInfo.Labels[nodeZoneLabelName]
+			if ok {
+				outputEntry[rule.Output+"_Zone"] = zone
```
jpinsonneau, line 211: Same as previous comm here

pkg/pipeline/transform/transform_network.go
OlivierCazade: IMO this is not host related the same way. HostIP and HostName belongs to the host, the pod doest not have this IP or name.

But the container is in the zone, we juste get the information through the node.

pkg/pipeline/transform/transform_network.go
OlivierCazade: I thought about adding this in fetchInformers, but this mean querying the informer for nodes for each flow even when the feature is disabled.

I prefered keeping it separated for performances reasons.

pkg/pipeline/transform/transform_network.go
OlivierCazade: Zone and region are not exactly the same thing, generally a region has multiple zones.

In your example `us-east-1` vs `us-east-1c`

jpinsonneau: so are we sure we always have zones when we have region ?

jpinsonneau: Ok sounds legit. Thanks !

pkg/pipeline/transform/transform_network.go
OlivierCazade: About telemetry this is now supported with a uniq tag: `k8s.zone`

jpinsonneau (approved):
Looks good, thanks @OlivierCazade !

OlivierCazade:
As discussed in scrum today, this will be tested post merge.

PR #579, opened 2024-01-31
Layer enrichment: remove hard-coded pieces, add tests

jotak:
No user-facing change.

(this PR requires to update the operator for configuring FLP)

### DIFF
--- a/pkg/api/transform_network.go
+++ b/pkg/api/transform_network.go
@@ -73,9 +73,15 @@ type NetworkTransformRule struct {
 }
 
 type K8sInfraRule struct {
-	Inputs      []string `yaml:"inputs,omitempty" json:"inputs,omitempty" doc:"entry inputs fields"`
-	Output      string   `yaml:"output,omitempty" json:"output,omitempty" doc:"entry output field"`
-	InfraPrefix string   `yaml:"infra_prefixes,omitempty" json:"infra_prefixes,omitempty" doc:"Namespace prefixes that will be tagged as infra"`
+	Inputs        []string       `yaml:"inputs,omitempty" json:"inputs,omitempty" doc:"entry inputs fields"`
+	Output        string         `yaml:"output,omitempty" json:"output,omitempty" doc:"entry output field"`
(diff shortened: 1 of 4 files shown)


pkg/api/transform_network.go
```diff
@@ -73,9 +73,15 @@ type NetworkTransformRule struct {
...
-	InfraPrefix string   `yaml:"infra_prefixes,omitempty" json:"infra_prefixes,omitempty" doc:"Namespace prefixes that will be tagged as infra"`
+	Inputs        []string       `yaml:"inputs,omitempty" json:"inputs,omitempty" doc:"entry inputs fields"`
+	Output        string         `yaml:"output,omitempty" json:"output,omitempty" doc:"entry output field"`
+	InfraPrefixes []string       `yaml:"infra_prefixes,omitempty" json:"infra_prefixes,omitempty" doc:"Namespace prefixes that will be tagged as infra"`
+	InfraRefs     []K8sReference `yaml:"infra_refs,omitempty" json:"infra_refs,omitempty" doc:"Additional object references to be tagged as infra"`
+}
+
+type K8sReference struct {
+	Name      string `yaml:"name,omitempty" json:"name,omitempty" doc:"name of the object"`
+	Namespace string `yaml:"namespace,omitempty" json:"namespace,omitempty" doc:"namespace of the object"`
```
jpinsonneau, line 84: I feel a bit confusing having these two arrays separated.

Why not having a single `filters` array similar to what we have in prometheus ?
https://github.com/netobserv/flowlogs-pipeline/blob/31e41804369087a22b45c2a6e017e5af3b388ebe/pkg/api/encode_prom.go#L70-L74

jotak: hmm that would force us to have in a common struct some values (fields) that are quite different, a single prefix in a case, or a name+namespace couple in the other .. we could of course force a generic structure for having both, but I'm not sure this is cleaner than the current solution (?)

jotak:
/hold
as @OlivierCazade started a similar work on his side

jotak:
superseded by https://github.com/netobserv/flowlogs-pipeline/pull/580

PR #580, opened 2024-02-01
Refactoring of transform network API

OlivierCazade:
## Description

Created different type for the different transform rules

## Dependencies

n/a

## Checklist

  * [x] Regression tests only (e.g. refactoring with no user-facing change).


OlivierCazade:
/hold
This is breaking change in the configuration API, we need to synchronize

jotak (commented):
Not yet reviewed in deep, but the approach looks good to me
specializing each rule params via their own struct rather than forcing a generic approach looks much better!
I'll have some work that depends on this, for NETOBSERV-1426

pkg/api/transform_network.go
```diff
@@ -63,23 +63,50 @@ func TransformNetworkOperationName(operation string) string {
...
-	Kubernetes      *K8sRule      `yaml:"kubernetes,omitempty" json:"kubernetes,omitempty" doc:"Kubernetes rule specific configuration"`
+	Type            string                 `yaml:"type,omitempty" json:"type,omitempty" enum:"TransformNetworkOperationEnum" doc:"one of the following:"`
+	KubernetesInfra *K8sInfraRule          `yaml:"kubernetes_infra,omitempty" json:"kubernetes_infra,omitempty" doc:"Kubernetes infra rule configuration"`
+	Kubernetes      *K8sRule               `yaml:"kubernetes,omitempty" json:"kubernetes,omitempty" doc:"Kubernetes rule configuration"`
+	AddSubnet       *NetworkAddSubnetRule  `yaml:"add_subnet,omitempty" json:"add_subnet,omitempty" doc:"Add subnet rule configuration"`
+	AddLocation     *NetworkGenericRule    `yaml:"add_location,omitempty" json:"add_location,omitempty" doc:"Add location rule configuration"`
+	AddIPCategory   *NetworkGenericRule    `yaml:"add_ip_category,omitempty" json:"add_ip_category,omitempty" doc:"Add ip category rule configuration"`
+	AddService      *NetworkAddServiceRule `yaml:"add_service,omitempty" json:"add_service,omitempty" doc:"Add service rule configuration"`
```
jpinsonneau, line 72: It seems we often have same type of rules applied in a row; on both source & destination for example.

Would it make sense to allow arrays on the objects above ?

Then you will be able to call something like:
```
api.NetworkTransformRules{{
    Type: api.AddKubernetesRuleType,
    Kubernetes: []api.K8sRule{
      api.K8sRule{
          Input:  "SrcAddr",
          Output: "SrcK8S",
      }, 
      api.K8sRule{
          Input:  "DstAddr",
          Output: "DstK8S",
      },
    },
}}})
```

We could even get rid of the type by just providing objects :thinking:

jotak (commented):
lgtm
I don't have a preference regarding Julien's suggestion (having an inner slice in k8s rules) .. it avoids some repetition but makes having an additional nested slice, hard to say which is simpler.
before merging we need to have the operator PR ready for the breaking change

pkg/api/transform_network.go
OlivierCazade: We could remove the type object but this would change consistency with the rest of the API that is using it.

About allowing an array, this mean we would have two nested level of rules instead of having one flat array of rules. This add complexity and I am not sure to see a value in doing it.

jotak:
@OlivierCazade can you edit the description to briefly explain what are the breaking changes, and give an example of how to convert an old config to a new one? This would help users who are hit by the breaking change to quickly figure out how to handle it

jotak:
I'm adding no-qe / no-doc as there's no user facing changes - I tested myself along with https://github.com/netobserv/network-observability-operator/pull/562

PR #596, opened 2024-02-16
Fix yaml unmarshaling of inlined config

jotak:
The `PromConnectionInfo` data wasn't unmarshaled when using yaml (it worked with json)
This is needed for agent's direct-flp feature


jpinsonneau (approved):
:+1: good catch !

PR #604, opened 2024-02-21
WIP NETOBSERV-1471 add tcp write stage

jpinsonneau:
## Description

This PR adds TCP write stage to FLP.
We need to decide if we move to [gRPC](https://github.com/netobserv/flowlogs-pipeline/pull/621) instead before merging.

## Dependencies

n/a

## Checklist

### DIFF
--- a/pkg/pipeline/write/write_tcp.go
+++ b/pkg/pipeline/write/write_tcp.go
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2021 IBM, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package write
+
+import (
+	"encoding/json"
+	"fmt"
+	"net"
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+	"github.com/sirupsen/logrus"
+)
+
+type writeTCP struct {
+	address string
+	conn    net.Conn
+}
+
+// Write writes a flow to tcp connection
+func (t *writeTCP) Write(v config.GenericMap) {
+	logrus.Tracef("entering writeTCP Write")
+	b, _ := json.Marshal(v)
+	// append new line between each record to split on client side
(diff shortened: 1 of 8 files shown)


pkg/pipeline/write/write_tcp.go
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (C) 2021 IBM, Inc.
```
KalmanMeth, line 2: Fix (or remove) Copyright statement

pkg/pipeline/write/write_tcp.go
```diff
@@ -0,0 +1,69 @@
...
+		return nil, fmt.Errorf("Write.TCP.Port must be specified")
+	}
+
+	logrus.Debugf("NewWriteTCP Listen %s", writeTCP.address)
+	l, err := net.Listen("tcp", writeTCP.address)
+	if err != nil {
+		return nil, err
+	}
+	defer l.Close()
+	clientConn, err := l.Accept()
```
KalmanMeth, line 62: It looks like you are expecting some other entity to initiate a TCP connection to FLP to which FLP will write its flow logs.
The code will block here until a single connection is completed. The rest of FLP will wait on this connection. Is this the desired behavior? 

What is the use case you have in mind? 

What if I have multiple targets to which I want to send my data via TCP? Can I set up multiple writeTCP stages? How will it work? or do we want to put the listen() operation into a separate thread and accumulate the connections that were attempted and send flow logs to all current connections?

pkg/pipeline/write/write_tcp.go
jpinsonneau: This is coming from both [eBPF packets exporter](https://github.com/netobserv/netobserv-ebpf-agent/blob/9507ebf9fbc5ba65f40befa4a55011e92ef32068/pkg/exporter/packets.go) and [flp stdout write stage](https://github.com/netobserv/flowlogs-pipeline/blob/main/pkg/pipeline/write/write_stdout.go) :smile: 

I just copied the code from there but I can remove the copyright if you want.

pkg/pipeline/write/write_tcp.go
jpinsonneau: This is a temporary PR used in netobserv CLI: https://github.com/netobserv/network-observability-cli/pull/2#discussion_r1497191885

The CLI deploys eBPF agents using `direct-flp` mode which embedd flowlogs-pipeline into the same pods.
Currently the connection between these pods and cli is made [using TCP](https://github.com/netobserv/network-observability-cli/blob/main/cmd/flow-capture.go#L68) after calling [oc port-forward](https://github.com/netobserv/network-observability-cli/blob/main/res/functions.sh#L37-L57) commands.
Using gRPC instead is still in discussion.

If we keep that TCP implementation, we can indeed improve and makes the wait optionnal + allow multiple targets :+1:

jpinsonneau:
Closing this in favor of https://github.com/netobserv/flowlogs-pipeline/pull/621

PR #606, opened 2024-02-22
Add test on IPFIX writer

jotak:
- Adding a test for the IPFIX exporter
- Small refactoring to remove some boilerplate code and have all fields more easily accessible

cc @bhale , I believe this test can help you when adding more ipfix fields


OlivierCazade (approved):
LGTM

PR #622, opened 2024-03-05
Use yaml unmarshaller instead of json's

jotak:
## Description

- This works without changing the input as YAML is a superset of JSON
- This allows to get rid of our prometheus model override, which was due to a bug when using JSON unmarshaller
- Refactor / simplify tests and confgen, doing less of the decode/reencode magic

## Dependencies

n/a

## Checklist


jotak:
Attempt fails, cf https://github.com/netobserv/flowlogs-pipeline/pull/623
We'd need to get rid of viper if we want to use yaml rather than json. I don't think viper brings anything super useful here, anyway.

KalmanMeth:
@eranra I think viper was originally used to allow passing in command line configuration parameters. We need to decide whether we still need this.

jotak:
@KalmanMeth yes it allows using various sources for parameters, but comes with this limitation of not being case sensitive which is annoying for yaml. Do you have any idea if the other parameter sources (such as ENV or command line) are actually used out in the wild?

eranra:
> @KalmanMeth yes it allows using various sources for parameters, but comes with this limitation of not being case sensitive which is annoying for yaml. Do you have any idea if the other parameter sources (such as ENV or command line) are actually used out in the wild?

@jotak @KalmanMeth In past projects we used cobra https://github.com/spf13/cobra - I do not know if this is better in compared with viper. In any event, I think that for the completeness and simplicity of a "stand-alone" tool such as FLP it does make sense to keep the option to provide command-line parameters

PR #629, opened 2024-03-12
NETOBSERV-1514: drop go1.20 support

jotak:
I had to upgrade the linter, and took that opportunity to enable more linters (same config as what we have on the agent)
But that led to quite a lot of fixes, this is why the PR is pretty big; There's no (intentional) functional changes


OlivierCazade (approved):
LGTM

PR #633, opened 2024-03-14
Provide k8s templates to just deploy an ipfix collector

jotak:


memodi (approved):
thanks @jotak
/lgtm
@Amoghrd - can you please try it out with FLP as collector? I think it's worth adding it in our test case while we figure what we can we do about `go-ipfix` collector

Amoghrd:
Will do

PR #639, opened 2024-03-19
NETOBSERV-1248 deduper using infinispan cache

jpinsonneau:
## Description

Implements infinispan shared cache:
- [X] Add cache functions for deduper case
Keeps `Egress` flows between two nodes in cache
Then merge `interfaces` and `ifdirections` when Ingress is captured
- [X] Make url configurable
- [ ] Auth / tls
- [ ] Implements extra cases

Merged flows contains a list of interfaces / ifdirections with a `*` next to names coming from Egress flow:
Screenshot from 2024-03-18 20-19-43
Screenshot from 2024-03-18 20-19-55

Steps to use this code:
- `go mod tidy && go mod vendor`
- Install `Infinispan Operator` OR `Red Hat Data Grid Operator`
- Create `cache` cluster in netobserv namespace with disabled TLS / Auth
```yaml
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: cache
  namespace: netobserv
spec:
  security:
    endpointAuthentication: false
    endpointEncryption:
      clientCert: None
      type: None
    endpointSecretName: cache-generated-secret
  container:
    memory: 1Gi
  service:
    replicationFactor: 2
    type: Cache
  jmx: {}
  configListener:
    enabled: true
    logging:
      level: info
  upgrades:
    type: Shutdown
  version: 8.4.6-2
  replicas: 1
```
- Create `deduper` cache
```yaml
apiVersion: infinispan.org/v2alpha1
kind: Cache
metadata:
  name: deduper
  namespace: netobserv
spec:
  clusterName: cache
  name: deduper
  updates:
    strategy: retain
```

These will provide a service `cache.netobserv.svc.cluster.local` on port `11222`.

## Dependencies

n/a

## Checklist


pkg/pipeline/transform/infinispan/cache.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return &Cache{
+		HttpClient: client,
+		ispn:       ispn,
+	}, nil
+}
+
+func (c *Cache) getDedupKey(v config.GenericMap) string {
+	return regex.ReplaceAllString(
+		fmt.Sprintf("%d-%d-%s-%d-%s-%d", v["TimeFlowEndMs"], v["Proto"], v["SrcAddr"], v["SrcPort"], v["DstAddr"], v["DstPort"]),
```
jpinsonneau, line 54: I put `TimeFlowEndMs` on top of the 5 tuple to avoid conflicts when 2 Egress comes before first Ingress.
As is, I miss some flows as you can see in the screens. We need to find a better way to unique identify each flows.

Any thoughts ?

pkg/pipeline/transform/infinispan/cache.go
```diff
@@ -0,0 +1,90 @@
...
+
+func NewCache(kubeConfigPath string) (*Cache, error) {
+	log.Debug("Entering NewCache ...")
+
+	client := New(Config{
+		Endpoint: "http://cache.netobserv.svc.cluster.local:11222",
+		/*Credentials: &Credentials{
+			Username: "netobserv",
+			Password: "netobserv",
+		},*/
```
jpinsonneau, line 42: Need to make these configurable and add TLS from secret

pkg/pipeline/transform/transform_network.go
```diff
@@ -119,7 +121,33 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
...
+					// append interfaces and direction to flow
+					srcFlow := *found
+					log.Debugf("merging cached flow %v with %v ...", srcFlow, outputEntry)
+					if len(srcFlow["Interfaces"].([]interface{})) == len(srcFlow["IfDirections"].([]interface{})) {
+						for i := 0; i < len(srcFlow["Interfaces"].([]interface{})); i++ {
+							outputEntry["Interfaces"] = append(outputEntry["Interfaces"].([]string), fmt.Sprintf("%s*", srcFlow["Interfaces"].([]interface{})[i].(string)))
+							outputEntry["IfDirections"] = append(outputEntry["IfDirections"].([]int), int(srcFlow["IfDirections"].([]interface{})[i].(float64)))
+						}
+					}
+					log.Debugf("merged flow: %v", outputEntry)
```
jpinsonneau, line 148: :point_up_2: merging `Egress` flow content in `Ingress` one.
For now just merging `Interfaces` and `IfDirections` and adding a `*` next to the names.
We could imagine merging more feature related contents, TCP flags and also check Bytes / Packets amounts

pkg/pipeline/transform/infinispan/cache.go
jotak: Perhaps we can do an assumption here, that the interfaces are not going to change over time across flows in the same 5-tuples. In fact I'm already doing this same assumption [here](https://github.com/netobserv/network-observability-console-plugin/pull/489/files#diff-b428443ff8eb670d5b01d47a599d4d67f24acddd9857e1dd327378499a2e3409R42) 
On top of that, it should be much better for the memory footprint.

The downside is that it works for merging interfaces, but might not work if we want to reuse the cache for other purposes.

pkg/pipeline/transform/transform_network.go
jotak: so if Egress comes after Ingress, it will be ignored? Perhaps it's another reason to not have the timestamp in the key as discussed [here](https://github.com/netobserv/flowlogs-pipeline/pull/639/files#r1530003170)

pkg/pipeline/transform/infinispan/cache.go
jpinsonneau: this cache is only for deduper so I'm fine with that; 

However are we sure that the interfaces are not changing when the pod is recreated for example ?
The ip addresses can be the same but ovn generated interfaces could change :thinking:

jpinsonneau: on top of that, I could push only fields we want to merge in the cache instead of the whole flow

pkg/pipeline/transform/transform_network.go
jpinsonneau: yeah currently if `Egress` comes after `Ingress`, it's pushed in the cache and never retreived.
Only `Ingress` will appear in storage and will not have merged fields.

An improvment could be to send first flow to the cache whatever direction is found and merge on the "second" one; but that will double the cache queries

pkg/pipeline/transform/infinispan/cache.go
jotak: > However are we sure that the interfaces are not changing when the pod is recreated for example ?
> The ip addresses can be the same but ovn generated interfaces could change

A pod restart would very very very likely use a different IP as far as I know; and even if that's not the case, the connection would need to be reinitiated so the source port would change

jotak: but perhaps in any case there should be an expiry mechanism

pkg/pipeline/transform/infinispan/cache.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/infinispan/infinispan-operator/pkg/mime"
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+	"github.com/sirupsen/logrus"
+
+	"github.com/infinispan/infinispan-operator/pkg/infinispan/client/api"
+	v14 "github.com/infinispan/infinispan-operator/pkg/infinispan/client/v14"
+)
+
+var log = logrus.WithField("component", "transform.Network.Cache")
+var regex = regexp.MustCompile(`[^a-zA-Z0-9 ]+`)
```
jotak, line 19: is there forbidden characters in key that require this replaceAll ?

jpinsonneau: I had issues with my initial attempt either with `.` or `:` when I had standart ips + ports key
I can give another try to get rid of this regex and improve perfs indeed

pkg/pipeline/transform/infinispan/cache.go
jpinsonneau: the cache already expire when full :smile_cat: you can also set a lifespan
https://infinispan.org/docs/stable/titles/configuring/configuring.html#expiration_configuring-memory-usage

jpinsonneau: I made changes in https://github.com/netobserv/flowlogs-pipeline/pull/639/commits/335d7370beb104917ab9806b4663b6f40a1f42f6

pkg/pipeline/transform/infinispan/cache.go
jpinsonneau: also addressed in https://github.com/netobserv/flowlogs-pipeline/pull/639/commits/335d7370beb104917ab9806b4663b6f40a1f42f6

pkg/pipeline/transform/transform_network.go
```diff
@@ -119,7 +121,10 @@ func (n *Network) Transform(inputEntry config.GenericMap) (config.GenericMap, bo
 			}
 			kubernetes.EnrichLayer(outputEntry, rule.KubernetesInfra)
 		case api.OpReinterpretDirection:
-			reinterpretDirection(outputEntry, &n.DirectionInfo)
+			isInner, isSrcReporter := reinterpretDirection(outputEntry, &n.DirectionInfo)
+			if n.cache != nil {
+				return n.cache.DedupFlows(isInner, isSrcReporter, outputEntry)
```
jotak, line 126: I was weirdly seeing no flow under layer=app filter and now I understand : this `return` breaks the rules loop processing :-)
It should be `continue`

jotak:
The new version is better but still increases the overall resource usage in my tests, although it's decrease for Loki & storage:
https://docs.google.com/spreadsheets/d/1qakBaK1dk_rERO30k1cSR4W-Nn0SXW4A3lqQ1sZC4rE/edit#gid=705512397

(By overall I mean: everything deployed in netobserv* namespaces - so that includes the infinispan deployment and loki)

CPU overall +146% (1.68)
Memory overall +30% (3GB)

Loki CPU -44% (0.035)
Loki memory -46% (652MB)
Loki storage -53%

So the benefit here is really on storage, and comes at a price in CPU and memory. We could run tests at a bigger scale to check if it's different.

Also one thing that seems wrong from my tests is that the captured traffic has dropped quite a lot, as if we were removing more flows than expected. That's to investigate.

pkg/pipeline/transform/transform_network.go
jpinsonneau: This case push flow to cache so you don't need to manage extra transform rules on it right ?
The goal is actually to break the pipeline here.

When the second flow will hit this stage, the first flow is retreived from cache and other transform rules are applied such as `layer`

jpinsonneau:
@jotak is this something we should still consider ?

I can replace the json `Marshal` / `Unmarshal` by `fmt.Sprintf` to improve FLP CPU usage but we will still have the infinispan resource impact :thinking: 
If you think it's not worth I will just close that PR.

PR #660, opened 2024-05-02
NETOBSERV-1597: skip RecordKeyMissing error

jpinsonneau:
## Description

Skip all `RecordKeyMissing` and consider these as `0` since we removed empty fields from eBPF agent.

https://github.com/netobserv/netobserv-ebpf-agent/blob/release-1.5/pkg/decode/decode_protobuf.go#L61-L67

If this needs to be configurable for specific cases please let me know. 

## Dependencies

n/a

## Checklist

* [X] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._

### DIFF
--- a/pkg/pipeline/encode/metrics_common.go
+++ b/pkg/pipeline/encode/metrics_common.go
@@ -228,8 +228,8 @@ func (m *MetricsCommonStruct) extractGenericValue(flow config.GenericMap, info *
 	}
 	val, found := flow[info.ValueKey]
 	if !found {
-		// No value means 0 to keep storage lightweight
-		return 0
+		// No value might mean 0 for counters, to keep storage lightweight - it can safely be ignored
+		return nil
 	}
 	return val
 }
--- a/pkg/pipeline/encode/metrics_common.go
+++ b/pkg/pipeline/encode/metrics_common.go
@@ -228,8 +228,8 @@ func (m *MetricsCommonStruct) extractGenericValue(flow config.GenericMap, info *
 	}
 	val, found := flow[info.ValueKey]
 	if !found {
-		m.errorsCounter.WithLabelValues("RecordKeyMissing", info.Name, info.ValueKey).Inc()
-		return nil
+		// No value means 0 to keep storage lightweight
+		return 0
 	}
 	return val
 }


pkg/pipeline/encode/metrics_common.go
```diff
@@ -228,8 +228,8 @@ func (m *MetricsCommonStruct) extractGenericValue(flow config.GenericMap, info *
 	}
 	val, found := flow[info.ValueKey]
 	if !found {
-		m.errorsCounter.WithLabelValues("RecordKeyMissing", info.Name, info.ValueKey).Inc()
-		return nil
+		// No value means 0 to keep storage lightweight
+		return 0
```
jotak, line 232: we should still return nil, I think: for counters, returning 0 or nil will have the same effect (ie. adding nothing to the counter) but for histograms, return 0 would actually corrupt the stats computation

jotak (approved):

jotak:
thanks @memodi 
/approve

pkg/pipeline/encode/metrics_common.go
jpinsonneau: It depends of the histograms I guess... Do you have some concrete examples in mind ?

For data consistency we should probably add filters before doing such assumptions. For example, DNS metrics must have DNSId, DNSLatencyMs fields etc. That's what we do on plugin query side.

jotak: All the histograms that we use are latency stuff, so I don't see a case where we would want to say: no latency == 0ms. If a flow is missing RTT, or Dns latency, then we should just ignore it for histogram stats. Expecting an explicit filter would add extra processing and could easily be error-prone IMO. I'd return your question: do you have an example in mind where we need to say nil == 0 for an histogram ?

jpinsonneau: bytes, packets and their respective drops fields must be considered as 0 when keys are not present

jotak: yes but that's fine these are counters not histograms, adding 0 to the counter is equivalent to ignoring it

Issue #641, opened 2024-03-24
flowlogs pipeline memory consumtion large clusters.


hihellobolke:
While following https://docs.openshift.com/container-platform/4.12/network_observability/configuring-operator.html 
 
And running flow logs pipeline with k8s enrichment on large clusters ~20k pods, the memory consumtion is huuuge... Now as this was running as daemonset, this kind of ddos' api server.

Would it be easier to allow use some sort of shared k8s enrich cache for all flp?

![image](https://github.com/netobserv/flowlogs-pipeline/assets/2939715/b413d62d-46fd-4fa6-a54d-5d2615871450)

Perhaps cache can be smarter like https://redis.io/docs/manual/client-side-caching/

In the end we had to recreate a custom grpc server that used shared cache to achieve network tracing on larger clusters.

jotak:
Hello @hihellobolke , thanks for reaching out
We are indeed exploring options for a shared cache, however before coming there, we don't recommend the daemonset approach for deploying netobserv on large clusters, as you see this doesn't scale very well. It's recommended to use the Kafka deployment model instead (in FlowCollector, setting `spec.deploymentMode` to `Kafka` [as documented here](https://docs.openshift.com/container-platform/4.15/network_observability/configuring-operator.html#network-observability-flowcollector-kafka-config_network_observability)). This way, FLP is deployed as a Deployment that you can scale up & down.

Would that work for you?

jotak:
Speaking of the proposed solution, I think this could work indeed. We have [a PoC that introduces Infinispan as a distributed cache](https://github.com/netobserv/flowlogs-pipeline/pull/639) (though in that context it was for another purpose, not for Kube API caching)

Another approach could be to not use k8s informers in FLP, and use k8s watches instead.  Because a problem with k8s informers is they cache whole gvks and not just the queried resources, leading to higher memory consumption & traffic bandwidth with kube API. We did this already in the operator, with something we called "narrowcache": https://github.com/netobserv/network-observability-operator/pull/476 , to cut down memory usage.

A downside that would affect both options would be slower processing time on cache misses, as everything would be lazy-loaded in FLP.

PR #681, opened 2024-06-21
NETOBSERV-1248: shared k8s cache

jotak:
The goal is to allow deploying a dedicated FLP instance as just a k8s cache component, alongside with the "traditional" FLP deployments.
The k8s cache is a Kafka producer, and traditional instances are
consumers.
    
- New FLP binary, k8s-cache, that just starts the informers and writes update events to
  Kafka
- New optional config for kubernetes enrichment, allowing to read from
  kafka instead of using informers

cf https://github.com/netobserv/flowlogs-pipeline/issues/641

Operator PR: https://github.com/netobserv/network-observability-operator/pull/684

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,15 +39,17 @@ IMAGE_TAG_BASE ?= quay.io/$(IMAGE_ORG)/flowlogs-pipeline
 
 # Image URL to use all building/pushing image targets
 IMAGE ?= $(IMAGE_TAG_BASE):$(VERSION)
+IMAGE_CACHE ?= $(IMAGE_TAG_BASE)-cache:$(VERSION)
 OCI_BUILD_OPTS ?=
 
 # Image building tool (docker / podman) - docker is preferred in CI
 OCI_BIN_PATH = $(shell which docker 2>/dev/null || which podman)
-OCI_BIN ?= $(shell basename ${OCI_BIN_PATH})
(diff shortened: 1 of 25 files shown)


Makefile
```diff
@@ -61,18 +63,21 @@ FORCE: ;
...
 	DOCKER_BUILDKIT=1 $(OCI_BIN) push ${IMAGE}-$(1);
+	DOCKER_BUILDKIT=1 $(OCI_BIN) push ${IMAGE_CACHE}-$(1);
 endef
 
 # manifest create a single arch target provided as argument
 define manifest_add_target
 	echo 'manifest add target $(1)'; \
 	DOCKER_BUILDKIT=1 $(OCI_BIN) manifest add ${IMAGE} ${IMAGE}-$(1);
+	DOCKER_BUILDKIT=1 $(OCI_BIN) manifest add ${IMAGE_CACHE} ${IMAGE_CACHE}-$(1);
 endef
```
jpinsonneau, line 81: Would it be better to have dedicated functions for cache case ?
It would avoid building both when you need only one :thinking:

jotak:
Superseded by #848

Makefile
jotak: actually I rewrote that differently in #848
I don't build a new image anymore, instead I include both binaries in the same image. It avoids having to refactor all the build process

PR #723, opened 2024-09-20
remove dependency gopacket

jotak:


msherif1234:
don't you need to wait for https://github.com/netobserv/netobserv-ebpf-agent/pull/416 then update the agent instead ?

jotak:
that's only a temporary PR (that's why it's a draft) .. it is not meant to be merged but I wanted to verify with perf tests if it helps or not, given that there seem to be several regressions

PR #740, opened 2024-10-16
NETOBSERV-1692: allow KEEP filtering logic

jotak:
## Description

- Allow to define filters for keeping flows. These rules are processed separately as the logic is reverse (first match means it's kept)
- Allow sampling within these new rules
- Allow regex checks (pre-compiled; need to check error on preprocess)

## Dependencies

n/a

## Checklist

* [x] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -162,6 +162,7 @@ Following is the supported API format for filter transformations:
                     remove_entry_if_equal: removes the entry if the field value equals specified value
                     remove_entry_if_not_equal: removes the entry if the field value does not equal specified value
                     remove_entry_all_satisfied: removes the entry if all of the defined rules are satisfied
+                    keep_entry: keeps the entry if the set of rules are all satisfied
                     add_field: adds (input) field to the entry; overrides previous value if present (key=input, value=value)
                     add_field_if_doesnt_exist: adds a field to the entry if the field does not exist
                     add_field_if: add output field set to assignee if input field satisfies criteria from parameters field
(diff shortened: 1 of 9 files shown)


docs/api.md
```diff
@@ -161,6 +161,7 @@ Following is the supported API format for filter transformations:
                     remove_entry_if_equal: removes the entry if the field value equals specified value
                     remove_entry_if_not_equal: removes the entry if the field value does not equal specified value
                     remove_entry_all_satisfied: removes the entry if all of the defined rules are satisfied
+                    keep_entry: keeps the entry if the set of rules are all satisfied
```
jpinsonneau, line 164:
```suggestion
                    keep_entry_all_satisfied: keeps the entry if the set of rules are all satisfied
```

For consistency

pkg/api/transform_filter.go
```diff
@@ -37,6 +46,7 @@ const (
 	RemoveEntryIfEqual       TransformFilterEnum = "remove_entry_if_equal"        // removes the entry if the field value equals specified value
 	RemoveEntryIfNotEqual    TransformFilterEnum = "remove_entry_if_not_equal"    // removes the entry if the field value does not equal specified value
 	RemoveEntryAllSatisfied  TransformFilterEnum = "remove_entry_all_satisfied"   // removes the entry if all of the defined rules are satisfied
+	KeepEntry                TransformFilterEnum = "keep_entry"                   // keeps the entry if the set of rules are all satisfied
```
jpinsonneau, line 49:
```suggestion
	KeepEntryAllSatisfied                TransformFilterEnum = "keep_entry_all_satisfied"                   // keeps the entry if the set of rules are all satisfied
```

pkg/pipeline/transform/transform_filter.go
```diff
@@ -159,25 +175,91 @@ func isRemoveEntrySatisfied(entry config.GenericMap, rules []*api.RemoveEntryRul
...
+		case api.KeepEntryIfNotRegexMatch:
+			if ok {
+				match, ok := checkRegex(r.KeepEntry.Value, val)
+				if !ok || match {
+					return false
+				}
+			} else {
+				return false
+			}
+		}
```
jpinsonneau, line 216: Could we reuse the code from encode_prom_metric.go here ?

+ the implementations are differents since the prom one just negate the result of Regex function for the not match case which means the NotRegex function will return true if the field is missing.

jotak: one other difference is that in prom filters api everything is string whereas here it can be anything, such as ints
using the native type of the field in the genericmap is more performant .. so maybe it's the prometheus api that I should align with this one for that matter..
I'll think more about that tomorrow!

docs/api.md
jotak: I used a simpler name because there's only 1 "keep entry" rule, as opposed to "remove entry" where there are a bunch of different ones which need to be distinguished from each other.

pkg/pipeline/transform/transform_filter.go
jotak: @jpinsonneau I finally did it: https://github.com/netobserv/flowlogs-pipeline/pull/740/commits/76131e0d8b973c4d3f0968e513536c13ef13f297

jpinsonneau (approved):

PR #748, opened 2024-11-06
Bumped golangci-lint from v1.56.2 to v1.61.0

PMaynard:
## Description

Bumped golangci-lint from v1.56.2 to v1.61.0

## Dependencies

n/a

## Checklist

Sorry, just fixing local build errors.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,7 +5,7 @@ linters:
     - cyclop
     - errname
     - exhaustive
-    - exportloopref
+    - copyloopvar
     - gocritic
     - gofmt
     - gosimple
@@ -16,15 +16,17 @@ linters:
     - stylecheck
     - typecheck
     - unused
+run:
+  go: "1.22"
 linters-settings:
-  stylecheck:
-    go: "1.22"
   gocritic:
     enabled-checks:
       - hugeParam
       - rangeExprCopy
       - rangeValCopy
       - indexAlloc
-      - deprecatedComment
+    settings:
+      ifElseChain:
+        minThreshold: 3
   cyclop:
     max-complexity: 20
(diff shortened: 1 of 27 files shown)


.bingo/Variables.mk
```diff
@@ -29,11 +29,11 @@ $(BINGO): $(BINGO_DIR)/bingo.mod
 	@echo "(re)installing $(GOBIN)/bingo-v0.5.1"
 	@cd $(BINGO_DIR) && $(GO) build -mod=mod -modfile=bingo.mod -o=$(GOBIN)/bingo-v0.5.1 "github.com/bwplotka/bingo"
 
-GOLANGCI_LINT := $(GOBIN)/golangci-lint-v1.56.2
+GOLANGCI_LINT := $(GOBIN)/golangci-lint-v1.61.0
```
jotak, line 32: this is an auto-generated file (see header), it's not how it should be updated.
TBH I'm not a fan of "bingo", the tool used for bin dependencies here (I find it a bit opaque for the simple job we're asking .. and I always forget how it works :-/ ) .. let me see if I can quickly remove that and align this repo we our other ones

jotak (changes requested):
Thanks @PMaynard 
.. but this probably won't do what you expect, this is a generated file that you changed

jotak:
now there's a bunch of lint errors to fix 😊

jotak:
I'm working on fixing the issues, will update soon

jotak (approved):

.bingo/Variables.mk
PMaynard: ops, my bad. Good choice removing bingo.

PR #797, opened 2025-01-23
NETOBSERV-1904: Enrich using UDN info

jotak:
## Description

Use the OVN annotation "k8s.ovn.org/pod-networks" to index pods  based on their IP + UDN label.
This is only enabled when SecondaryNetworks is configured with the "udn" index.
It leverages the "Udns" field that the ebpf agent populates, to find udn labels per flow. This should allow indexing with overlapping IPs.

It also populates the Src/Dst Network Name with the matched UDN label.

## Dependencies

n/a

### DIFF
--- a/pkg/pipeline/transform/kubernetes/cni/udn.go
+++ b/pkg/pipeline/transform/kubernetes/cni/udn.go
@@ -1,0 +1,88 @@
+package cni
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/api"
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	ovnAnnotation = "k8s.ovn.org/pod-networks"
+)
+
+type UDNHandler struct {
+}
+
+func UDNKey(label, ip string) SecondaryNetKey {
+	key := label + "~" + ip
+	return SecondaryNetKey{NetworkName: label, Key: key}
+}
+
+func (m *UDNHandler) Manages(indexKey string) bool {
+	return indexKey == "udn"
+}
+
+func (m *UDNHandler) BuildKeys(flow config.GenericMap, rule *api.K8sRule) []SecondaryNetKey {
+	var keys []SecondaryNetKey
+
+	var ip string
+	var udns []string
+	var ok bool
+	if len(rule.IPField) > 0 {
+		ip, ok = flow.LookupString(rule.IPField)
+		if !ok {
+			return nil
+		}
+	}
+	if len(rule.UDNsField) > 0 {
+		v, ok := flow[rule.UDNsField]
+		if !ok {
+			return nil
+		}
+		udns, ok = v.([]string)
+		if !ok || len(udns) == 0 {
+			return nil
+		}
+	}
+
+	for _, udn := range udns {
+		if udn != "" && udn != "default" {
+			keys = append(keys, UDNKey(udn, ip))
+		}
+	}
+
+	return keys
+}
+
+func (m *UDNHandler) GetPodUniqueKeys(pod *v1.Pod) ([]string, error) {
+	// Example:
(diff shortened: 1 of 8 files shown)


pkg/pipeline/transform/kubernetes/cni/udn.go
```diff
@@ -0,0 +1,88 @@
...
+		if err := json.Unmarshal([]byte(statusAnnotationJSON), &annot); err != nil {
+			return nil, fmt.Errorf("failed to index from OVN annotation, cannot read annotation %s: %w", ovnAnnotation, err)
+		}
+		var keys []string
+		for label, info := range annot {
+			if label != "default" {
+				if rawip, ok := info["ip_address"]; ok {
+					if ip, ok := rawip.(string); ok {
+						// IP has a CIDR prefix (bug??)
+						parts := strings.SplitN(ip, "/", 2)
```
msherif1234, line 75: why not using `net.ParseCIDR()` ?

pkg/pipeline/transform/kubernetes/cni/udn_test.go
```diff
@@ -0,0 +1,31 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	udnHandler = UDNHandler{}
+	udnPod     = v1.Pod{ObjectMeta: metav1.ObjectMeta{Annotations: map[string]string{}}}
+	udnConfig  = `{"default":{"ip_addresses":["10.128.2.20/23"],"mac_address":"0a:58:0a:80:02:14","routes":[{"dest":"10.128.0.0/14","nextHop":"10.128.2.1"},{"dest":"100.64.0.0/16","nextHop":"10.128.2.1"}],"ip_address":"10.128.2.20/23","role":"infrastructure-locked"},
+	"mesh-arena/primary-udn":{"ip_addresses":["10.200.200.12/24"],"mac_address":"0a:58:0a:c8:c8:0c","gateway_ips":["10.200.200.1"],"routes":[{"dest":"172.30.0.0/16","nextHop":"10.200.200.1"},{"dest":"100.65.0.0/16","nextHop":"10.200.200.1"}],"ip_address":"10.200.200.12/24","gateway_ip":"10.200.200.1","tunnel_id":16,"role":"primary"}}`
+)
```
msherif1234, line 16: probably good to add v6 test case too

pkg/pipeline/transform/kubernetes/cni/udn.go
jotak: Because it looks weird and I initially thought it _should_ be a plain IP ... doing this "split" would work even if it's later changed to be a plain IP, so I feel it's safer. And I don't need the IP / CIDR object, just a string, so no need for parsing.

jotak: (btw, I'll add the test to verify it also works with a plain IP, and IPv6)

pkg/pipeline/transform/kubernetes/informers/informers.go
```diff
@@ -328,10 +342,22 @@ func (k *Informers) initPodInformer(informerFactory inf.SharedInformerFactory) e
...
+				log.WithError(err).Infof("Secondary network cannot be identified")
+			}
+		}
+		if k.hasUDN {
+			if udnKeys, err := udn.GetPodUniqueKeys(pod); err != nil {
+				// Log the error as Info, do not block other ips indexing
+				log.WithError(err).Infof("UDNs cannot be identified")
+			} else {
+				keys = append(keys, udnKeys...)
+			}
```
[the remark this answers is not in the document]
jotak: I haven't tested live, but I can add a unit test for that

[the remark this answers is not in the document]
jotak: done

pkg/pipeline/transform/kubernetes/cni/udn_test.go
jotak: done

jpinsonneau (approved):

Issue #850, opened 2025-03-08
Enhancement: Provide precompiled binaries for Linux e.g. RHEL(8)


Nachtfalkeaw:
In our Environment it is difficult or Impossible to compile These applications. To make it useable for us a precomoiled binary for Linux RHEL(8) would be benefical.

I assume it would Help other to quickstart and Test this tool.

jotak:
Hi @Nachtfalkeaw ,

Thanks for opening this discussion. Currently, this component is used mostly as a container running in Kubernetes - it is published as such on quay.io: https://quay.io/repository/netobserv/flowlogs-pipeline?tab=tags

We can definitely explore the idea to _also_ provide binaries.
In the meantime, you may try extracting the binary from the image, if you wish to try it.

Nachtfalkeaw:
Hi,

Thanks for Feedback. I am Not familar with this docker/kubernetes stuff. Will try to load the Container and try to extract the File and try it in RHEL8.

I tried other goflow2 forks and some Had issues with glibc Missing for my RHEL8 Environment.

Nachtfalkeaw:
@jotak 
I tried what you suggested however it has problems with glibc:

```
root@ubiquiti:/opt/flow-log-pipeline# ./flowlogs-pipeline --help
./flowlogs-pipeline: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./flowlogs-pipeline)
./flowlogs-pipeline: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./flowlogs-pipeline)
root@ubiquiti:
```

maybe you can provide a build with these flags/options?
https://github.com/netsampler/goflow2/issues/381

jotak:
Hi @Nachtfalkeaw 
Sorry for the long delay. I'm running a build with CGO_ENABLE=0 in #932 to see if that helps

jotak:
I've also added the binary extraction so they'll be automatically included in the next releases

Nachtfalkeaw:
Looks like the app would be able to start now. I do not get these errors I got before - however I do not have any working config available.

jotak:
@Nachtfalkeaw what are you trying to do? Collect netflows or IPFIX data, from an existing source?

Here's a config example that listens netflows/ipfix and write them to stdout:

```yaml
log-level: info
pipeline:
  - name: ingest
  - name: out
    follows: ingest
parameters:
[...]

PR #932, opened 2025-04-24
Disable cgo on upstream builds and attach binaries to releases

jotak:
See #850


jotak:
Releasing seems to work fine, see https://github.com/netobserv/flowlogs-pipeline/releases/tag/untagged-c055728b3262cdd48862

OlivierCazade (approved):
LGTM

Issue #940, opened 2025-05-02
SamplerAddress with unexpected format - SamplerAddress:[10 30 102 1] - Transform/Format values?


Nachtfalkeaw:
Hello,

I collect IPFIX from enterprise firewalls. I used "goflow2" earlier I the sampler_address was like
`SamplerAddress:10.30.102.1`

With flowlogs-pipeline the result is this:
`SamplerAddress:[10 30 102 1]`

I would like to have the sampler address as a real IP address like in goflow2.
and/or I would like to see a mapping possibility to map a received sampler address to a value in a list, e.g.

`SamplerAddress:[10 30 102 1]  --->  SamplerAddress:MyFirewallName`

Or maybe a possibility to transform a value via regex or other syntax?

This is my pipeline:
```
pipeline:
  - name: ipfix_netflow_ingest
  - name: transform_add_subnet
    follows: ipfix_netflow_ingest
  - name: loki_write
    follows: transform_add_subnet
  - name: local_write
    follows: transform_add_subnet

parameters:
  - name: ipfix_netflow_ingest
    ingest:
      type: collector
      collector:
        hostName: u999fmlab001l
        port: 2055

  - name: transform_add_subnet
    transform:
      type: network
      network:
        rules:
          - type: add_subnet
            add_subnet:
              input: SrcAddr
              output: SrcSubnet24
              subnet_mask: /24
          - type: add_subnet
            add_subnet:
              input: SrcAddr
              output: SrcSubnet16
              subnet_mask: /16
          - type: add_subnet
            add_subnet:
              input: DstAddr
              output: DstSubnet24
              subnet_mask: /24
          - type: add_subnet
            add_subnet:
              input: DstAddr
              output: DstSubnet16
              subnet_mask: /16

  - name: local_write
    write:
      type: stdout

  - name: loki_write
    write:
      type: loki
      loki:
```

jotak:
We actually use goflow2 under the hood, I think it's just a matter of rendering as string

PR #950, opened 2025-05-05
NetFlow/IPFIX: render sampler address as string

jotak:
Fixes https://github.com/netobserv/flowlogs-pipeline/issues/940

cc @Nachtfalkeaw


OlivierCazade (approved):
LGTM

PR #967, opened 2025-05-22
Fix suppressGoMetrics, rename it to suppressDefaultMetrics

### LINKED ISSUE #944: metricsSettings - suppressGoMetrics results in not metrics at all
Hello,

I use this config:
```
metricsSettings:
  suppressGoMetrics: true
  prefix: flp_operational_
  port: 9102
```

If I set `suppressGoMetrics: true` I do not get any metrics at all. If I set it to "false" I receive Go Metrics and the other expected metrics.

jotak:
Fixes #944

The suppress flag was creating a dedicated prometheus registry, but without changing the default registry, which was still in use.

**Breaking change**
If you configured metrics settings with `suppressGoMetrics: true`, you must update it to `suppressDefaultMetrics: true`


OlivierCazade (approved):
LGTM

PR #969, opened 2025-05-22
Expose goflow2 workers and sockets config

### LINKED ISSUE #951: Goflow2 Performance Parameters Like worker, buffer
Hello,
Goflow2 allows to Limit worker and Buffer size. This helps to Limit the CPU usage and the memory usage in high flow Environments. And of course it allowes to process more flows with more Ressources.

This is from the Performance Page:

```
$ goflow2 -listen flow://:6343/?count=4&workers=16&blocking=false&queue_size=1000000
                                ^        ^          ^              ^
                                ┃        ┃          ┃              ┗ In buffered mode, the amount of packets stored in memory
                                ┃        ┃          ┗ Real-time mode
                                ┃        ┗ Decoding workers
                                ┗ Open sockets listening
```

What are the Default Parameters in flowlogs-pipeline and is it possible to adjust These?

I am using loki.write to Forward flows and i am Not Sure what Happens If the Loki Server ist mit reachable or overloaded (http Status 429) and flp is in backoff time. It May lead to huge memory usage If Buffer value is high and Socket Count, too.

In the ingesti Stage it could be a good possibility to add These Parameters.

jotak:
Fixes #951 (partly)

Note that this PR doesn't affect netobserv operator, which doesn't use ipfix/netflow as an input

Example:

```yaml
    ingest:
      type: collector
      collector:
        hostName: localhost
        port: 2055
        workers: 8
        sockets: 4
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -90,6 +90,8 @@ Following is the supported API format for the NetFlow / IPFIX collector:
          port: the port number to listen on, for IPFIX/NetFlow v9. Omit or set to 0 to disable IPFIX/NetFlow v9 ingestion
          portLegacy: the port number to listen on, for legacy NetFlow v5. Omit or set to 0 to disable NetFlow v5 ingestion
          batchMaxLen: the number of accumulated flows before being forwarded for processing
+         workers: the number of netflow/ipfix decoding workers
+         sockets: the number of listening sockets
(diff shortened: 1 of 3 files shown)


OlivierCazade (approved):
LGTM

PR #992, opened 2025-06-11
NETOBSERV-2285: opt: use uint64 hash instead of string for cache key

jotak:
In timed_cache, use uint64 hash (fnv 64a) instead of string builder
Uses the same implementation as in prom go client

Also:
- [Bump agent, with improved pb decode](https://github.com/netobserv/flowlogs-pipeline/pull/992/commits/a8e13410001541b7ab48dcab3dac4a0e8e313b20)

image

- [reduce allocations in metrics encode](https://github.com/netobserv/flowlogs-pipeline/pull/992/commits/da2f2679d60fab7c2cca388004928305e4a3f18f)

image


OlivierCazade (approved):
LGTM

PR #1090, opened 2025-09-15
Fail on wrong pipeline config with duplicated stage names

### LINKED ISSUE #333: Fix usage of multiple stages with same name
We saw an example with the following configuration:
```
pipeline:
- name: ingest_collector
- follows: ingest_collector
  name: transform_generic
- follows: transform_generic
  name: transform_network
- follows: transform_network
  name: extract_aggregate
- follows: transform_network
  name: extract_conntrack
- follows: extract_conntrack
  name: extract_aggregate
- follows: extract_aggregate
  name: encode_prom
```
The same stage (extract_aggregate) is specified twice as following 2 different stages. This was done to achieve the desired output, one with connection tracking and without.
Apparently this works because we now use channels as input, so a stage may receive from 2 different sources.
Not sure this is desired to be legal. We should probably have detected this and issued an error.
Ensure that things work as desgined, and provide the correct sequence of stages to achieve the desired outcome.

jotak:
Fixes #333

### DIFF
--- a/pkg/pipeline/pipeline_builder.go
+++ b/pkg/pipeline/pipeline_builder.go
@@ -161,6 +161,9 @@ func (b *builder) presetIngester(ing ingest.Ingester) {
 func (b *builder) readStages() error {
 	for _, param := range b.configParams {
 		log.Debugf("stage = %v", param.Name)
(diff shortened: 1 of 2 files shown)


leandroberetta (approved):
lgtm

PR #1227, opened 2026-04-02
Enhance release workflow with SBOM and signing

Helion55:
Updated the release workflow to include SBOM generation and image signing.

## Description
Added new jobs in ```release.yml``` workflow,
1. cosign-installer
2. Sign and verify the image with GitHub OIDC Token
3. Attest Provenance attestations
4. Generate SBOM
5. Attest SBOM

Solves: https://github.com/netobserv/netobserv-operator/issues/2490#issuecomment-4167949613

## Summary by CodeRabbit

* **Chores / Release Hardening**
  * Hardened the release pipeline with automated container image signing and verification using OIDC-based authentication.
  * Automatically generate and publish build provenance attestations and an SPDX SBOM for each release, and publish these security artifacts to the registry.
  * Improved release reliability by capturing the multi-architecture image digest during the build and using it for signing and attestation.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,22 +5,28 @@ on:
 
 permissions:
   contents: write
+  id-token: write
+  attestations: write
+  artifact-metadata: write
 
 env:
   WF_REGISTRY_USER: netobserv+github_ci
   WF_ORG: netobserv
   WF_MULTIARCH_TARGETS: amd64 arm64 ppc64le s390x
   WF_REGISTRY_NAME: quay.io
+  WF_IMAGE_NAME: quay.io/netobserv/flowlogs-pipeline
 
 jobs:
   push-image:
     name: build and push release
     runs-on: ubuntu-latest
+
     steps:
       - name: checkout
         uses: actions/checkout@v6
         with:
           persist-credentials: false
+
       - name: validate tag
         run: |
           tag=`git describe --exact-match --tags 2> /dev/null`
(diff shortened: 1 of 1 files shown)


jotak:
FYI I've run a test here: https://github.com/netobserv/flowlogs-pipeline/actions/runs/25151488067/job/73722871869
 and it's failing.
It's not super easy to test for you as you don't have permissions to run the workflows (or you would need to set up all that on your own fork, and a quay account with your own quay secret in github) .. sound like an overkill for you, I think the best would be to move some of the logic into the makefile, which can be run from a dev machine, to not depend on actions for testing.

.github/workflows/release.yml
```diff
@@ -67,3 +68,47 @@ jobs:
                 data: await fs.readFile(`./release-assets/${file}`)
               }); 
             }
+      - name: cosign-installer
+        # You may pin to the exact commit or the version.
+        # uses: sigstore/cosign-installer@cad07c2e89fa2edd6e2d7bab4c1aa38e53f76003
```
jotak: we should actually do exactly this, pin all tagged versions used here and in other actions to their digest sha

jotak:
I've started another try after fixing an issue in the image name & multi-arch: https://github.com/netobserv/flowlogs-pipeline/actions/runs/25152829470/job/73727257890
There's two issues next:
- we don't have the images digests at the signing stage, signing refers to a tag, which isn't secure and will stop working at some point according to the warning message. So we need to adapt our build steps to provide the digests as an output.
- OIDC isn't configured; need to find how to do that.

.github/workflows/release.yml
Helion55: I have made the changes here.

jotak:
Thanks @Helion55 
I've rebased and fixed an issue with cosign `--certificate-identity` argument which was incorrect ; I just tested the workflow, it works well! Signing + SBOM

jotak (approved):

PR #1297, opened 2026-07-16
NETOBSERV-2515: Use TLS environment variables for TLS config if present

leandroberetta:
## Description

Use TLS environment variables for TLS config if present

## Dependencies

https://github.com/netobserv/netobserv-operator/pull/2823

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added environment-driven TLS profile overrides for minimum TLS version, cipher suites, and curve preferences.
  * Added optional TLS configuration for gRPC write connections.
* **Bug Fixes**
  * Invalid TLS settings now fail safely without partially applying changes.
  * Ineffective cipher-suite overrides are ignored under TLS 1.3.
* **Tests**
  * Expanded coverage for TLS profiles, gRPC connections, and secure server defaults.
* **Refactor**
  * Simplified Prometheus TLS handling.

### DIFF
--- a/pkg/pipeline/write/write_grpc.go
+++ b/pkg/pipeline/write/write_grpc.go
@@ -2,6 +2,7 @@ package write
 
 import (
 	"context"
+	cryptotls "crypto/tls"
 	"encoding/json"
 	"fmt"
 
@@ -46,7 +47,15 @@ func NewWriteGRPC(params config.StageParam) (Writer, error) {
 		return nil, fmt.Errorf("write.grpc param is mandatory: %v", params.Write)
 	}
 	logrus.Debugf("NewWriteGRPC ConnectClient %s:%d...", writeGRPC.hostIP, writeGRPC.hostPort)
-	clientConn, err := grpc.ConnectClient(writeGRPC.hostIP, writeGRPC.hostPort, nil)
+	var tlsCfg *cryptotls.Config
+	if params.Write.GRPC.TLS != nil {
+		var err error
+		tlsCfg, err = params.Write.GRPC.TLS.Build()
+		if err != nil {
+			return nil, fmt.Errorf("failed to build TLS config: %w", err)
+		}
+	}
+	clientConn, err := grpc.ConnectClient(writeGRPC.hostIP, writeGRPC.hostPort, tlsCfg)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/pipeline/write/write_grpc.go
+++ b/pkg/pipeline/write/write_grpc.go
@@ -65,19 +65,28 @@ func NewWriteGRPC(params config.StageParam) (Writer, error) {
 // If cfg is set, it is used to build the explicit TLS configuration, which is
 // then extended with any TLS-profile environment overrides (TLS_MIN_VERSION,
 // TLS_CIPHER_SUITES, TLS_CURVE_PREFERENCES).
-// If cfg is nil, nil is returned and the connection remains insecure: profile
-// env vars only constrain TLS when TLS is already configured, they must not
-// enable it on a previously plaintext writer.
+// If cfg is nil, a TLS configuration is only created when the environment
+// yields at least one valid override; otherwise nil is returned and the
+// connection remains insecure, preserving prior behavior.
 func resolveTLSConfig(cfg *api.ClientTLS) (*cryptotls.Config, error) {
-	if cfg == nil {
-		return nil, nil
+	if cfg != nil {
+		tlsCfg, err := cfg.Build()
+		if err != nil {
+			return nil, fmt.Errorf("failed to build TLS config: %w", err)
+		}
+		if _, err := tlsprofile.Apply(tlsCfg); err != nil {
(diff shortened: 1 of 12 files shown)


pkg/pipeline/write/write_grpc.go
```diff
@@ -46,7 +47,15 @@ func NewWriteGRPC(params config.StageParam) (Writer, error) {
...
-	clientConn, err := grpc.ConnectClient(writeGRPC.hostIP, writeGRPC.hostPort, nil)
+	var tlsCfg *cryptotls.Config
+	if params.Write.GRPC.TLS != nil {
+		var err error
+		tlsCfg, err = params.Write.GRPC.TLS.Build()
+		if err != nil {
+			return nil, fmt.Errorf("failed to build TLS config: %w", err)
+		}
+	}
+	clientConn, err := grpc.ConnectClient(writeGRPC.hostIP, writeGRPC.hostPort, tlsCfg)
```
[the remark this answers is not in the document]
jpinsonneau: 👆 that seems valid point

pkg/tlsprofile/tlsprofile.go
```diff
@@ -0,0 +1,89 @@
...
+func Apply(c *tls.Config) {
+	if c == nil {
+		return
+	}
+	if v := minVersionFromEnv(); v != 0 {
+		c.MinVersion = v
+	}
+	if suites := cipherSuitesFromEnv(); len(suites) > 0 {
+		c.CipherSuites = suites
+	}
```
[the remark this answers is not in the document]
jpinsonneau: +1

pkg/tlsprofile/tlsprofile.go
```diff
@@ -0,0 +1,89 @@
...
+	if curves := curvePrefsFromEnv(); len(curves) > 0 {
+		c.CurvePreferences = curves
```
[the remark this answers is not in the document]
jpinsonneau: +1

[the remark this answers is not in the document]
leandroberetta: I'm not sure about this one, OpenShift profiles (Old, Intermediate, Modern, Custom), allow TLS version < 1.3.

pkg/pipeline/write/write_grpc.go
```diff
@@ -46,10 +49,44 @@ func NewWriteGRPC(params config.StageParam) (Writer, error) {
...
+	envCfg := &cryptotls.Config{MinVersion: cryptotls.VersionTLS13}
+	applied, err := tlsprofile.Apply(envCfg)
+	if err != nil {
+		return nil, fmt.Errorf("invalid TLS profile override: %w", err)
+	}
+	if !applied {
+		return nil, nil
+	}
+	return envCfg, nil
+}
```
jpinsonneau, line 92: We should ensure to have tls in the `write.grpc` config:

```suggestion
// If cfg is nil, nil is returned and the connection remains insecure: profile
// env vars only constrain TLS when TLS is already configured, they must not
// enable it on a previously plaintext writer.
func resolveTLSConfig(cfg *api.ClientTLS) (*cryptotls.Config, error) {
	if cfg == nil {
		return nil, nil
	}
	tlsCfg, err := cfg.Build()
	if err != nil {
		return nil, fmt.Errorf("failed to build TLS config: %w", err)
	}
	if _, err := tlsprofile.Apply(tlsCfg); err != nil {
		return nil, fmt.Errorf("invalid TLS profile override: %w", err)
	}
	return tlsCfg, nil
}
```

leandroberetta: Hi @jpinsonneau at first I applied your suggestion but then I realized the behaviour was as I intended, when the TLS env vars are present, TLS should be configured. Am I misunderstanding something?

jotak: I think on one hand @jpinsonneau is right, when no TLS is used, TLS profile should be ignored (if I remember correctly, this behavior is explicitly stated in the TLS profile guide).

But the problem is that TLS was not implemented here previously. IMO that's a mistake, it should have been implemented and enabled by default, like we're doing everywhere else? I'm pretty sure we discussed that in the past but I don't remember the details. If I'm right, this is used by the CLI.

IMO we should enabled TLS by default here (so if config is nil, still apply the profile like @leandroberetta is doing) and in parallel check how it's done on the CLI and verify it works with TLS enabled.

jpinsonneau: oh ok if the goal is to enable by default then skip my suggestion :wink:

leandroberetta: According to the tls guide:

> All TLS servers must honor the cluster profile, including servers listening only on localhost or 
  inside a pod.

It's true that previously we don't have tls enabled but I agree that it was a mistake.

Took a look into the CLI and yes, it will require be updated to use TLS too in the grpc server (collector). I'm working on it.

The behaviour is like the comment in the code, if the env vars for TLS are present, they will be used, otherwise it won't configure tls.

Also, I've just seen ther is an skill to verify the tls adoption, I will use it to validate.

PR #1308, opened 2026-08-10
NETOBSERV-2783 Enrich ASN labels

jpinsonneau:
## Summary
- Add `add_asn_label` network transform: LPM IP → local ASN from `FRRConfiguration` (`frrk8s.metallb.io/v1beta1`)
- Index `routers[].prefixes` and `toAdvertise` prefixes → `routers[].asn` (not `toReceive` / peer ASN)
- Soft-fail when FRR CRD is absent (cloud / no BGP); enrichment no-ops
- Separate stage name `asns` (ready for operator Dynamic wiring later)
- Unit tests for extract, LPM, enrich, and network transform wiring

## Usage
```yaml
name: asns
transform:
  type: network
  network:
    rules:
      - type: add_asn_label
        add_asn_label: { input: SrcAddr, output: SrcASN }
      - type: add_asn_label
        add_asn_label: { input: DstAddr, output: DstASN }
```

## Out of scope (follow-ups)
- FlowCollector admin `asnLabels` maps
- Operator pipeline / Dynamic CM wiring
- Remote/peer ASN from RIB/BMP

## Test plan
- [x] Unit tests (`pkg/pipeline/transform/frr`, `Test_AddASNLabel`)
- [x] `make lint` / build clean
- [x] Manual: BM cluster with FRR + `FRRConfiguration`; assert `SrcASN`/`DstASN` on flows matching advertised CIDRs
- [ ] Manual: cluster without FRR CRD; pipeline starts, no ASN fields

@luisjira PTAL

## Summary by CodeRabbit

- **New Features**
  - Added network transform support for enriching flow records with source and destination ASN labels.
  - ASN resolution uses the most specific matching network prefix and updates automatically from FRR configuration data.
  - Added an example configuration and sample flow records demonstrating ASN enrichment.

- **Bug Fixes**
  - ASN resolution now handles overlapping or duplicate prefixes deterministically.

- **Documentation**
  - Reformatted CLI help text in the README for improved alignment and readability.

### DIFF
--- a/pkg/pipeline/transform/frr/enrich.go
+++ b/pkg/pipeline/transform/frr/enrich.go
@@ -1,0 +1,91 @@
+/*
+ * Copyright (C) 2026 IBM, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package frr
+
+import (
+	"strconv"
+	"sync"
+
+	"github.com/netobserv/flowlogs-pipeline/pkg/api"
+	"github.com/netobserv/flowlogs-pipeline/pkg/config"
+)
+
+var (
+	storeMu sync.RWMutex
+	store   Store
+)
+
+// InitStore starts the FRRConfiguration informer. Safe to call once; subsequent
+// calls are no-ops while a store is already set.
+func InitStore(kubeConfigPath string) error {
+	storeMu.Lock()
+	defer storeMu.Unlock()
+	if store != nil {
+		return nil
+	}
+	s := NewInformerStore()
+	if err := s.Start(kubeConfigPath); err != nil {
+		return err
+	}
+	store = s
+	return nil
+}
+
+// SetStore injects a Store implementation (for tests).
+func SetStore(s Store) {
+	storeMu.Lock()
+	defer storeMu.Unlock()
+	store = s
+}
+
+// ResetStore clears the package-level store (for tests).
+func ResetStore() {
+	storeMu.Lock()
+	defer storeMu.Unlock()
+	if is, ok := store.(*InformerStore); ok && is != nil {
+		is.Stop()
+	}
+	store = nil
+}
+
+func getStore() Store {
+	storeMu.RLock()
+	defer storeMu.RUnlock()
+	return store
+}
+
+// Enrich looks up the IP in inputField via LPM against FRR advertised prefixes
+// and writes the matching local ASN (stringified) to outputField.
+func Enrich(outputEntry config.GenericMap, rule *api.NetworkAddASNLabelRule) {
+	if rule == nil || rule.Input == "" || rule.Output == "" {
+		log.Error("add_asn_label rule: missing input or output configuration")
+		return
+	}
+	ip, ok := outputEntry.LookupString(rule.Input)
+	if !ok || ip == "" {
+		return
+	}
+	s := getStore()
+	if s == nil {
+		return
+	}
+	asn, ok := s.Lookup(ip)
+	if !ok {
+		return
+	}
+	outputEntry[rule.Output] = strconv.FormatUint(uint64(asn), 10)
+}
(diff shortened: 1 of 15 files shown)


pkg/pipeline/transform/frr/store.go
```diff
@@ -0,0 +1,226 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/cache"
+)
+
+const (
+	frrGroup    = "frrk8s.metallb.io"
+	frrVersion  = "v1beta1"
+	frrResource = "frrconfigurations"
```
jotak, line 40: Is it really the metallb API that we need to check here? I'm not well aware of the BGP setup that we want to support, but I thought that would rather use some API from ovn-k

See for reference: https://ovn-kubernetes.io/master/okeps/okep-5088-evpn/#frr-integration-with-ovn-kubernetes (I haven't read/understood everything)

jpinsonneau: The group name is historical. On OCP, CNO runs frr-k8s in `openshift-frr-k8s`; MetalLB and OVN-K share the same FRR instance and the same CRD. OVN-K docs explicitly use:

```
apiVersion: frrk8s.metallb.io/v1beta1
kind: FRRConfiguration
```

See [OKEP-5296](https://ovn-kubernetes.io/master/okeps/okep-5296-bgp/). OVN-K does not replace that API; it generates `FRRConfigurations` from `RouteAdvertisements`.

jpinsonneau: Should also work with BGP cloud:
https://github.com/openshift/bgp-cloud-connector#frrconfiguration-from-cudnbgpconfig--one-per-az

pkg/pipeline/transform/frr/enrich.go
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright (C) 2026 IBM, Inc.
```
jotak, line 2: did you move to IBM already? 😁

pkg/pipeline/transform/frr/enrich.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+var (
+	storeMu sync.RWMutex
+	store   Store
+)
+
+// InitStore starts the FRRConfiguration informer. Safe to call once; subsequent
+// calls are no-ops while a store is already set.
+func InitStore(kubeConfigPath string) error {
```
jotak: I think we could avoid using global+mutex here? If I understand correctly the mutex is only for protection in tests, because it's a global shared state. But if we make `InitStore` returns `(Store, error)`, we simplify all of that. `Network` would store the frrStore within its struct, and tests would create and if needed inject their own store in `tr.(*Network).frrStore` ? That would remove all the global getter / setter / reset as well.

pkg/pipeline/transform/frr/enrich.go
jotak: (all new files have the same)

jpinsonneau: don't we keep their copyright on the new files ?

pkg/pipeline/transform/frr/enrich.go
jpinsonneau: Sure, let's try that approach

jpinsonneau: https://github.com/netobserv/flowlogs-pipeline/pull/1308/commits/4c52b937c3ad6ffbe7290188c422184bb651e33f

pkg/pipeline/transform/transform_network.go
```diff
@@ -239,56 +243,77 @@ func initNetworkServices(cfg *api.TransformNetwork) (*netdb.ServiceNames, error)
...
+	if flags.needToInitKubeData {
+		if err := kubernetes.InitInformerDatasource(&jsonNetworkTransform.KubeConfig, opMetrics); err != nil {
 			return nil, err
 		}
 	}
 
+	var frrStore frr.Store
+	if flags.needToInitFRR {
+		// Soft-fail: BGP/FRR is optional; enrichment simply no-ops without a store.
+		s, err := frr.InitStore(jsonNetworkTransform.KubeConfig.ConfigPath)
```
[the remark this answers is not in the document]
jpinsonneau: If the FLPInProcess stop the pod will stop. It's not worth implementing that here

pkg/pipeline/transform/frr/enrich.go
jotak: no, the copyright tells "this code/file was created by IBM" which is factually wrong, would be a misattribution.

If you look at FLP file headers, it's a mix of no-headers (the redhat way) and IBM copyright headers - depending on who created the file.

jpinsonneau: Done in https://github.com/netobserv/flowlogs-pipeline/pull/1308/commits/e6c2a8f28261b584b00f7f313592cff1c93b8b51

Thanks !

jotak (approved):
thanks!
/lgtm

PR #1310, opened 2026-08-14
NETOBSERV-2916: Fix secondary-networks enrichment with informers

Amoghrd:
## Description

Secondary networks enrichment is broken with FLP informers

## Dependencies
https://github.com/netobserv/netobserv-operator/pull/2880

## Checklist

  * [x] No QE (e.g. trivial change with high reviewer's confidence, or per agreement with the QE team).

## Summary by CodeRabbit

- **New Features**
  - Added a `--secondary-networks` option for configuring multiple secondary networks.
  - Supports semicolon-separated network groups with comma-separated index types.
  - Automatically trims entries, ignores empty groups and values, and applies valid configurations during initialization.
  - Secondary-network lookups now identify the associated network name alongside pod metadata.
  - Improves flexibility and accuracy when defining and resolving network-related settings for informer-based workflows.

### DIFF
--- a/pkg/pipeline/transform/kubernetes/datasource/datasource.go
+++ b/pkg/pipeline/transform/kubernetes/datasource/datasource.go
@@ -43,6 +43,7 @@ func (d *Datasource) IndexLookup(potentialKeys []string, ip string) *model.Resou
 	if d.kubernetesStore != nil {
 		return d.kubernetesStore.IndexLookup(potentialKeys, ip)
 	}
+	// Fallback to local informers if available (nil when k8scache is enabled)
 	if d.Informers != nil {
 		return d.Informers.IndexLookup(potentialKeys, ip)
 	}
--- a/pkg/pipeline/transform/kubernetes/datasource/datasource.go
+++ b/pkg/pipeline/transform/kubernetes/datasource/datasource.go
@@ -40,21 +40,13 @@ func NewDatasourceK8sCache() *Datasource {
 }
 
 func (d *Datasource) IndexLookup(potentialKeys []string, ip string) *model.ResourceMetaData {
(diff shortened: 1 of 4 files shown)


flp-informers
jotak, line 1: please remove this file

pkg/pipeline/transform/kubernetes/datasource/datasource.go
```diff
@@ -40,14 +40,21 @@ func NewDatasourceK8sCache() *Datasource {
...
-	if d.Informers != nil {
-		return d.Informers.IndexLookup(potentialKeys, ip)
+	if result != nil {
+		for _, key := range potentialKeys {
+			if name, ok := result.SecondaryNetNames[key]; ok {
+				result.NetworkName = name
+				break
+			}
+		}
 	}
```
jotak: I think this shouldn't be needed, NetworkName should be set alongside any other informer data during lookup. If that's not the case, then something else must be wrong ...
I'll try to dig more on that

jotak: After checking, there's indeed a missing NetworkName update but this should be set in KubernetesStore's IndexLookup instead
And since it means mutating the `ResourceMetaData` object, it must be write-lock protected (currently `KubernetesStore.IndexLookup` is only read-lock protected)

jotak (approved):
thanks @Amoghrd 
/lgtm

Amoghrd:
Test passing post-changes

 [sig-netobserv] Network_Observability Author:aramesha-High-81677-Validate UDN with NetObserv [Serial] [484.977 seconds]

   SUCCESS! 9m7.818438742s --- PASS: TestBackend (548.13s)
  PASS

PR #1335, opened 2026-09-17
NETOBSERV-3001: prefer non-terminated pods for IP attribution

jotak:
## Description

We still index all pods, included the terminated ones, in order to account for the latency in flow processing (agent evicts flows in batches every N seconds, default 15, which leaves plenty of room for informers to be sync'ed too early)

So the approach is to introduce a preference algorithm at query time, when IPs collide.

## Dependencies

n/a

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Kubernetes lookups now prefer active Pods when multiple Pods share the same IP or custom key.
  * Terminated Pods are correctly identified when they have succeeded or failed.
  * Terminated Pods remain available when they are the only matching result.
  * Multiple-match reporting is now limited to cases where no active Pod can be selected.


pkg/pipeline/transform/kubernetes/informers/informers.go
```diff
@@ -116,11 +133,11 @@ func (k *Informers) infoForCustomKeys(idx cache.Indexer, kind string, potentialK
 			return nil, false
 		}
 		if len(objs) > 0 {
-			info := objs[0].(*model.ResourceMetaData)
+			info, ambiguous := preferLive(objs)
```
leandroberetta, line 136: Previously, multiple matches were always reported as ambiguous. I understand that a terminated Pod and a live Pod sharing an IP is no longer ambiguous because the live Pod is preferred. However, if multiple live Pods match the same IP or custom index key, the first one is returned and the lookup is reported as non-ambiguous. Although this should not normally happen for primary IPs, shouldn’t we preserve the ambiguity signal for that case?

leandroberetta (commented):
lgtm mostly but I have a minor question. Thanks @jotak.

pkg/pipeline/transform/kubernetes/informers/informers.go
jotak: yeah that's a good point, I started doing like you said but that forces us to scan the whole list of candidates only for ambiguity reporting (which really is just a debugging/stats thing, not impacting correctness)
So I preferred to return early for keeping the same performance as before

leandroberetta (approved):
lgtm

pkg/pipeline/transform/kubernetes/informers/informers.go
```diff
@@ -400,6 +417,7 @@ func (k *Informers) initPodInformer(informerFactory inf.SharedInformerFactory, c
 			SecondaryNetKeys:  flatKeys,
 			SecondaryNetNames: namedKeys,
 			IPs:               ips,
+			Terminated:        pod.Status.Phase == v1.PodSucceeded || pod.Status.Phase == v1.PodFailed,
```
stleerh, line 420: Even though k8scache mode isn't exposed to the user, since it's adding a new field to `ResourceMetaData`, ideally it should:

1. Add a new `terminated` field to `ResourceEntry` in proto/k8scache.proto
2. Map this field in both directions in pkg/pipeline/transform/kubernetes/k8scache/converter.go

jotak: good point, I didn't think about cache informers ; not a problem for 1.12.x but I'll fix for 2.0

jotak: => https://github.com/netobserv/flowlogs-pipeline/pull/1336

